PR #49, opened 2023-01-25
feat(spec-renderer-mini): fit & finish [khcp-5769]

kaiarrowood:
# Summary

Fit & finish for `spec-renderer-mini`.
Addresses [KHCP-5769](https://konghq.atlassian.net/browse/KHCP-5769).

Includes:
- style updates
- new prop `tags`
- documentation updates

image

## PR Checklist

[KHCP-5769]: https://konghq.atlassian.net/browse/KHCP-5769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/portal/spec-renderer-mini/src/components/SpecRendererMini.vue
+++ b/packages/portal/spec-renderer-mini/src/components/SpecRendererMini.vue
@@ -63,6 +63,7 @@
           <SpecItem
             v-for="item in untaggedItems"
             :key="item.key"
+            class="mini-spec-renderer-untagged"
             :is-summary="isSummary"
             :item="item"
             @click="handleSelection"
@@ -230,6 +231,10 @@ const hasRequiredProps = (items: object[], requiredProps: string[]): boolean => 
     }
   }
 
+  .mini-spec-renderer-untagged {
+    margin-left: -12px;
+  }
+
   .is-summary {
     .mini-spec-renderer-section {
       border: 1px solid var(--grey-200);
@@ -249,6 +254,11 @@ const hasRequiredProps = (items: object[], requiredProps: string[]): boolean => 
         font-family: monospace;
         max-width: 65%;
       }
+    }
+
+    .mini-spec-renderer-untagged {
+      margin-top: 12px;
+      margin-left: unset;
     }
   }
 }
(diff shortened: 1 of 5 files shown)


packages/portal/spec-renderer-mini/src/types/specs.ts
```diff
@@ -10,3 +10,14 @@ export interface SpecItemType {
   selected?: boolean
   key?: string
 }
+
+interface SpecExternalDoc {
+  description: string
+  url: string
```
mptap, line 16: Should there be a `name` attribute to identify the Doc?

packages/portal/spec-renderer-mini/src/components/SpecRendererMini.vue
```diff
@@ -182,21 +231,54 @@ const hasRequiredProps = (items: SpecItemType[]): boolean => {
...
 <style lang="scss">
-.kong-portal-spec-renderer-mini {
+.kong-spec-renderer-mini {
   // all but the first have top margin
   .k-collapse + .k-collapse {
     margin-top: var(--spacing-md);
   }
+
+  .k-collapse .k-collapse-heading {
+    margin: 0 !important;
```
mptap, line 275: It would be nice to avoid using !important` if possible, can we use any other css specifier?

mptap (approved):
Lgtm, apart from a couple of non-blocking questions

packages/portal/spec-renderer-mini/src/types/specs.ts
kaiarrowood: I'm not sure. I was using the swagger-ui default .yaml as my guide when building the object so it may be only a partial definition and pretty sure it doesn't display a name in the UI because it hides it under some generic "click to see more" text. 

Either way I don't think it matters because we aren't doing anything with that part of the object currently anyway. Perhaps down the line if there is some requirement we can update the definition as needed.

packages/portal/spec-renderer-mini/src/components/SpecRendererMini.vue
kaiarrowood: Unfortunately we don't have an alternative currently because the style I am overriding with the !important has itself a !important on it since it's one of our utility classes.. the alternative would be to update KCollapse to not use the utility class for this style but I wasn't sure the need in here warranted the change there.

packages/portal/spec-renderer-mini/src/types/specs.ts
tkajtoch: OpenAPI [doesn't support](https://swagger.io/specification/#external-documentation-object) external doc `name` by default. It could be extended for our use case, but I don't think it's needed.

FYI _technically_ the `description` field can have CommonMark formatting.

tkajtoch (approved):
LGTM

PR #73, opened 2023-01-30
feat(spec-renderer-mini): add filtering by tag and prepare to integrate into spec-renderer package [TDX-2777]

tkajtoch:
# Summary

* Add filtering feature
  * Add `isFilterable` (boolean, defaults to `true`) and `filterFunc` (function, defaults to built-in filtering function) props
* Remove support for the obsolete `isSummary` prop
* Refactor the code to prepare for renaming and moving to the spec-renderer package
* Update types and docs
* Make all styles scoped

## PR Checklist


packages/portal/spec-renderer-mini/src/components/OperationsListItem.vue
```diff
@@ -0,0 +1,171 @@
...
+    :data-testid="`mini-spec-item-${item.path}`"
+    type="button"
+    @click="emit('click', item)"
+  >
+    <h2
+      class="summary truncate"
+      :title="item.summary ?? ''"
+    >
+      {{ item.summary }}
+    </h2>
```
adamdehaven, line 14: question: should there be a `v-if` here to hide the `h2` tag if `summary` is empty?

packages/portal/spec-renderer-mini/src/components/OperationsListItem.vue
```diff
@@ -0,0 +1,171 @@
+<template>
+  <button
+    class="item"
+    :class="{ 'item--selected': isSelected, [`item--method-${item.method}`]: true }"
+    :data-testid="`mini-spec-item-${item.path}`"
+    type="button"
+    @click="emit('click', item)"
+  >
```
adamdehaven, line 8: question: any reason you're using a regular button here instead of `KButton`?

packages/portal/spec-renderer-mini/src/components/OperationsListSectionHeader.vue
```diff
@@ -0,0 +1,104 @@
+<template>
+  <button
```
adamdehaven, line 2: question: any reason you're using a button here instead of `KButton`?

packages/portal/spec-renderer-mini/src/components/OperationsListSectionHeader.vue
```diff
@@ -0,0 +1,104 @@
...
+    :aria-controls="contentElementId"
+    :aria-expanded="isCollapsed ? 'false' : 'true'"
+    :aria-label="isCollapsed ? `Expand section &quot;${name}&quot;` : `Collapse section &quot;${name}&quot;`"
+    class="header"
+    type="button"
+    @click="emits('toggle')"
```
adamdehaven, line 8: Where you have this defined as `emits` can you change to `emit` to match the standard?

packages/portal/spec-renderer-mini/src/components/OperationsListSectionHeader.vue
```diff
@@ -0,0 +1,104 @@
...
+    type: String,
+    default: undefined,
+  },
+  contentElementId: {
+    type: String,
+    required: true,
+  },
+})
+
+const emits = defineEmits(['toggle'])
```
adamdehaven, line 64:
```suggestion
const emit = defineEmits(['toggle'])
```

packages/portal/spec-renderer-mini/README.md
```diff
@@ -120,8 +153,44 @@ If enabled, will display list summary view with a few more details instead of ta
 
 Controls the width of the rendered spec. Currently supports numbers (will be converted to px), auto, `vw`, `vh`, and percentages for width.
 
+### `filterFunc`
```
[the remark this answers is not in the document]
tkajtoch: I was following https://kongponents.konghq.com/components/select.html#filterfunc, but sure, I can adjust it!

packages/portal/spec-renderer-mini/src/components/OperationsListItem.vue
tkajtoch: The button here is used specifically for accessibility semantics. Using KButton would not have any added value, in my opinion. The same applies for other native `<button>` usages in this PR

packages/portal/spec-renderer-mini/src/components/OperationsListSectionHeader.vue
tkajtoch: Fixed!

packages/portal/spec-renderer-mini/src/components/OperationsList.vue
```diff
@@ -0,0 +1,298 @@
...
+      return items.filter((item) => item.tags?.some((tag) => tag.toLowerCase().includes(query)))
+    }) as OperationListFilterFunction,
+    validator: (maybeFunc) => !!maybeFunc && typeof maybeFunc === 'function',
+  },
+})
+
+const emit = defineEmits(['selected'])
+
+// Generate unique identifier of this instance for safe HTML element id generation
+const uid = ref([...Array(8)].map(() => Math.random().toString(36)[2]).join(''))
```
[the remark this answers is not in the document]
tkajtoch: We don't need anything cryptographically random here. `uid` provides a pseudo-unique fragment for DOM element identifiers when the same operations and tags are rendered in multiple `OperationsList` on the same page which is very unlikely. The reason it's here is to be a safety layer and help assistive technology link section headings and contents.

`uuid` would work great here, but it takes a bit too much space, and `Crypto.randomUUID()` isn't as widely available yet.

packages/portal/spec-renderer-mini/README.md
```diff
@@ -53,7 +55,7 @@ app.use(SpecRendererMini)
 
 ```
 
-```html
+```vue
```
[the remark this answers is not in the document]
tkajtoch: https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L7165 it is valid for GFM, but I'll change it to be safe npmjs renders it correctly

packages/portal/spec-renderer-mini/src/components/OperationsList.vue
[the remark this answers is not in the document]
adamdehaven: My only other suggestion then would be to create this as a `computed` instead of `ref` since the value doesn't change.

packages/portal/spec-renderer-mini/README.md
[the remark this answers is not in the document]
adamdehaven: So really the reason I suggest changing is that when we use VuePress/VitePress and they integrate with PrismJS, `vue` isn't a valid language; so this keeps us "the same" across implementations

packages/portal/spec-renderer-mini/src/components/OperationsList.vue
```diff
@@ -0,0 +1,298 @@
...
+        <!-- Items without any tags -->
+        <div class="section">
+          <template
+            v-for="item in untaggedItems"
+            :key="`${item.path}-${item.method}`"
+          >
+            <slot
+              :item="item"
+              name="item"
+              @click="handleSelection"
```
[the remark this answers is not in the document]
tkajtoch: Good catch! Switched to passing a `select` prop to the slot

packages/portal/spec-renderer-mini/README.md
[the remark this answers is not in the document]
tkajtoch: that is a strong argument, even if we're not using VuePress for now :) I switched back to `html`!

packages/portal/spec-renderer-mini/sandbox/App.vue
```diff
@@ -65,11 +61,11 @@
 
 <script setup lang="ts">
 import { onMounted, ref, watch } from 'vue'
-import { SpecItemType, SpecTag } from '../src/types'
+import { Operation, Tag } from '../src/types'
 import { KButton, KLabel } from '@kong/kongponents'
 import SpecRendererMini from '../src'
```
davidma415, line 66: Don't we need to update the name of this folder (and file) if we aren't calling it spec renderer mini anymore?

davidma415:
One small nit - we should rename the entire folder as well as update all usages. Should this be called `SpecOperationsList`?

tkajtoch:
@davidma415 I wanted to leave the full package renaming to another PR. This one got a bit too big... You're right, it should be exported as `SpecOperationsList` when we move it to `@kong-ui-public/spec-renderer`, but I think internally we can just leave it as is.

davidma415 (approved):
LGTM

PR #113, opened 2023-02-06
test(spec-renderer): component tests [khcp-5839]

kaiarrowood:
# Summary

Add component tests for spec-renderer components.
Address [KHCP-5839](https://konghq.atlassian.net/browse/KHCP-5839) and [KHCP-5786](https://konghq.atlassian.net/browse/KHCP-5786).

Non-test-related changes:
- Add new slot `ops-list-empty-state` in `SpecRenderer` linked to `SpecOperationsList` `empty-state` slot
- Default `hasSidebar` in `SpecDetails` to `false`
- `SpecDetails` emits new event `active-operation` (not documented, since only used for testing purposes)
- Hide filter input in `SpecOperationsList` if in error state
- Fix untagged items to adhere to `disableSelection` prop in `SpecOperationsList`

## PR Checklist

[KHCP-5839]: https://konghq.atlassian.net/browse/KHCP-5839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KHCP-5786]: https://konghq.atlassian.net/browse/KHCP-5786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/portal/spec-renderer/src/components/SpecOperationsList.vue
```diff
@@ -119,6 +130,7 @@ import type { OperationListFilterFunction, Operation, OperationListItem, Tag } f
 import { v1 as uuidv1 } from 'uuid'
 import clonedeep from 'lodash.clonedeep'
 import composables from '../composables'
+import { KInput, KIcon, KCollapse } from '@kong/kongponents'
```
adamdehaven, line 133: issue: this needs to be removed; Kongponents are available within the host app. If they aren't working in the tests, then the test setup needs to be updated instead.

packages/portal/spec-renderer/src/components/SpecDetails.cy.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import { OperationListItem } from '../types'
+import { jsonSpec, specOpItem } from '../../fixtures/spec-data'
+import SpecDetails from './SpecDetails.vue'
+
+describe('<SpecDetails />', () => {
+  it('renders correctly with props', () => {
+    cy.mount(SpecDetails, {
+      props: {
+        document: jsonSpec as any,
```
adamdehaven, line 11: question: if you're going to cast the type, can't you cast it the same as the PropType in the component?

packages/portal/spec-renderer/src/components/SpecDetails.cy.ts
```diff
@@ -0,0 +1,88 @@
...
+      },
+    })
+
+    cy.getTestId('kong-public-ui-spec-details-swagger').should('be.visible')
+    cy.getTestId('default-styles').should('exist')
+
+    cy.getTestId('kong-public-ui-spec-details-swagger').then(() => {
+      // Check for emitted event
+      // @ts-ignore - vueWrapper DOES exist
+      cy.wrap(Cypress.vueWrapper.emitted()).should('have.property', 'active-operation')
```
adamdehaven, line 78: @evanlindsey heres an example of what I was talking about

packages/portal/spec-renderer/src/components/SpecOperationsList.vue
```diff
@@ -154,7 +165,9 @@ const props = defineProps({
     type: String,
     default: '210',
   },
-
+  /**
+   * Used for component testing
+   */
```
adamdehaven, line 170: I don't see the `testMode` prop being utilized in the component anywhere, but it's passed as `true` in the component test. 

What is this prop doing? Sorry if I'm missing something

packages/portal/spec-renderer/src/components/SpecRenderer.cy.ts
```diff
@@ -1,12 +1,127 @@
...
+    cy.getTestId('kong-public-ui-spec-details-swagger').should('be.visible')
+    cy.getTestId('default-styles').should('exist')
+  })
+
+  it('renders operationsList with correct px width via navWidth prop', () => {
+    const width = 350
+
+    cy.mount(SpecRenderer, {
+      props: {
+        navWidth: width + '',
```
adamdehaven, line 32:
```suggestion
        navWidth: String(width),
```

packages/portal/spec-renderer/src/components/operations-list/OperationsListItem.vue
```diff
@@ -1,12 +1,12 @@
...
   <button
-    class="item"
+    class="spec-operations-list-item"
     :class="{
       'item--selected': isSelected,
       [`item--method-${item.method}`]: true,
       'disabled': disableSelection
     }"
-    :data-testid="`spec-operations-list-item-${item.path}`"
+    :data-testid="`spec-operations-list-item-${item.method.toLowerCase()}${item.path.replaceAll('/', '-')}-${item.tag || '-'}`"
```
adamdehaven, line 9: suggestion: this attribute has a lot of logic; worth moving to a computed variable?

packages/portal/spec-renderer/src/components/SpecDetails.vue
```diff
@@ -77,6 +80,8 @@ function showAndScrollToOperation() {
 
   swaggerRef.value.showOperation(props.activeOperation)
   swaggerRef.value.scrollToOperation(props.activeOperation)
+  // required for component testing
+  emit('active-operation')
```
adamdehaven, line 84: question: I see that this was added for the component tests, but I'm not sure I understand what the test is actually achieving by looking for this emit? 

Essentially the test is just verifying that the prop value is set to the correct value and that the component mounts (since this emit is triggered in the onMounted function), so I'm not sure this is actually providing any sort of test coverage beyond checking that the prop mounted and could be removed.

packages/portal/spec-renderer/src/components/SpecDetails.cy.ts
evanlindsey: @adamdehaven we need to add `"cypress/vue"` to this types array: https://github.com/Kong/public-ui-components/blob/main/packages/portal/spec-renderer/tsconfig.json#L7

It works in `khcp-ui` because of this import: https://github.com/Kong/khcp-ui/blob/main/client/cypress/support/commands.ts#L26

If you `import 'cypress/vue'` or any of its exports in a file, it will recognize the types.

adamdehaven: Ah sweet let me try adding that in one of the global tsconfig files instead of here locally

adamdehaven: @kaiarrowood I added this on the `main` branch - if you rebase from main, you shouldn't need the `ts-ignore` for `Cypress.vueWrapper`

adamdehaven (approved):

PR #119, opened 2023-02-07
feat(portal): add id to headings [TDX-2849]

codelemur:
This is a convenience that I think folks would expect to find, and allows for linking section headings directly via anchor tags

### DIFF
--- a/packages/portal/document-viewer/src/components/nodes/Heading.vue
+++ b/packages/portal/document-viewer/src/components/nodes/Heading.vue
@@ -1,5 +1,8 @@
 <template>
-  <component :is="tag">
+  <component
+    :is="tag"
+    :id="slug"
+  >
     <slot />
   </component>
 </template>
@@ -19,8 +22,13 @@ const props = defineProps({
     },
     default: 1,
   },
+  slug: {
+    type: String,
+    default: '',
+  },
 })
 
+const slug = `${props.slug}`
 const tag = `h${props.level}`
 </script>
 
--- a/packages/portal/document-viewer/src/utils/addUniqueHeadingSlugs.ts
+++ b/packages/portal/document-viewer/src/utils/addUniqueHeadingSlugs.ts
@@ -1,0 +1,29 @@
+import { toSlug } from './toSlug'
+export function addUniqueHeadingSlugs<T = any>(children: Array<T>): Array<T> {
+
+  if (!children) return children
+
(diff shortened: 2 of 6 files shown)


packages/portal/document-viewer/src/components/nodes/Heading.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <component :is="tag">
+  <component
+    :is="tag"
+    :id="slug"
```
Mierenga, line 4: question: if we want better assurance of unique element ids across the consuming application, should we add a specific prefix to the slugs so it would be like `doc-${tag}-${slug}`? Not sure if that adds a lot of value for the tradeoff more visually complex URLs.

adamdehaven: We have a pattern of generating an id with the `uuid` package if needed

packages/portal/document-viewer/src/utils/addUniqueHeadingSlugs.ts
```diff
@@ -0,0 +1,27 @@
+import { toSlug } from './toSlug'
+export function addUniqueHeadingSlugs<T = any>(children: Array<T>): Array<T> {
+
+  const slugMap = new Map<string, number>()
+  // loop through the children and their children recursively
```
Mierenga, line 5: nit: we went with not needing to loop recursively
```suggestion
  // loop through the children
```

packages/portal/document-viewer/src/components/nodes/Heading.vue
codelemur: After some discussion I think we should keep it human-readable, but with a docs prefix (but ability for folks to pass their own if they're into that kind of thing). Leaves us the option of putting the tag in it if we want to consume it, yeah?

Mierenga (approved):
LGTM

Mierenga (dismissed):

packages/portal/document-viewer/src/components/nodes/Heading.vue
Mierenga: The main thing is we want a deterministic value that will persist across page loads and is derived from the content itself, to enable shareable links and consistent reloads.

adamdehaven (approved):

adamdehaven (approved):

PR #144, opened 2023-02-15
fix(spec-renderer): untagged items [khcp-6150]

kaiarrowood:
# Summary

Untagged operations should be added to a `default` section to match Swagger UI behavior, `default` section displayed last.
Addresses [KHCP-6150](https://konghq.atlassian.net/browse/KHCP-6150).

image

## PR Checklist

[KHCP-6150]: https://konghq.atlassian.net/browse/KHCP-6150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/portal/spec-renderer/src/components/SpecOperationsList.vue
```diff
@@ -180,6 +199,7 @@ const uid = computed<string>(() => uuidv1())
 
 const filterQuery = ref<string>('')
 const taggedItems = ref<OperationListItem[]>([])
+const untaggedItemsSectionName = ref('default')
```
adamdehaven, line 202: question: If this doesn't change from `default` does it need to be a `ref`?

kaiarrowood: Moved to `const.ts`

packages/portal/spec-renderer/src/components/SpecOperationsList.vue
```diff
@@ -130,6 +149,7 @@ import type { OperationListFilterFunction, Operation, OperationListItem, Tag } f
 import { v1 as uuidv1 } from 'uuid'
 import clonedeep from 'lodash.clonedeep'
 import composables from '../composables'
+import { DEFAULT_UNTAGGED_SECTION_NAME } from '../const'
```
adamdehaven, line 152: I don't think this one string really warrants a new file import, you could just define it statically `const var = 'default` here in the template

Related: we typically don't abbreviate the filenames here -- e.g. `constants.ts` 

but I don't think you need the separate file here since it's only used here

kaiarrowood: I was following the pattern here: https://github.com/Kong/public-ui-components/blob/main/packages/core/copy-uuid/src/const.ts

I put it there because it IS a constant that we don't want people randomly changing without thought. I will rename to `constants.ts`.

adamdehaven: Ah ok makes sense 👍🏼 thanks for renaming 😄

kaiarrowood: I also renamed the `const.ts` in `copy-uuid`

adamdehaven (approved):

Mierenga (dismissed):
LGTM

packages/portal/spec-renderer/src/components/SpecOperationsList.vue
```diff
@@ -79,23 +79,42 @@
...
+            <div :id="getSectionContentId(untaggedItemsSectionName)">
+              <template
+                v-for="item in untaggedItems"
+                :key="`${item.path}-${item.method}`"
+              >
+                <slot
+                  :item="item"
+                  name="untagged-item"
+                  :section="untaggedItemsSectionName"
+                  :select="() => handleSelection(item)"
```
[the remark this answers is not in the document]
kaiarrowood: Turns out we DO need this so the function isn't executed until the event fires

Mierenga (approved):

PR #150, opened 2023-02-16
feat: [KHCP_6176] Expose i18n Translation component for import without Vue plugin

ValeryG:
# Summary

Now `i18n-t` can be used can be used from component libraries without it being registered via plugin on application level.

### DIFF
--- a/packages/core/i18n/README.md
+++ b/packages/core/i18n/README.md
@@ -267,7 +267,36 @@ In some cases we do not have access to the Vue `app` and cannot relay on registe
   const i18nT = i18nTComponent(i18n)
 
 </script>
+```
 
+Or in old `defineComponent` way
+
+```html
+
+<template>
+  <i18n-t
+    :i18n="i18n"
+    tag="h1"
+    keypath="global.default">
+      <a href="https://google.com">Google</a>
+  </i18n-t>
+</template>
+
+
+<script lang="ts">
+  import { defineComponent } from 'vue'
+  import { createI18n, i18nTComponent } from '../src'
+  import english from './locales/en.json'
+
+  export default defineComponent({
+    components: { i18nT: i18nTComponent() },
+    setup() {
+      return {
+        i18n: createI18n('en-us', english)
+      }
+    }
+  })
+</script>
 ```
 
 ## Formatting numbers, dates and times
--- a/packages/core/i18n/README.md
+++ b/packages/core/i18n/README.md
@@ -161,7 +161,7 @@ _Please, do not use this method in any new code. This prevents using parameters/
 
 ### HTML safe formatting with `<i18n-t>`
 
-#### With registered i18nPlugin
+#### When registered as Vue Plugin
 
 Sometimes it is needed to render translated message with HTML as part of the parameters. For this Provided component can be used.
 
--- a/packages/core/i18n/README.md
+++ b/packages/core/i18n/README.md
@@ -260,11 +260,11 @@ In some cases we do not have access to the Vue `app` and cannot relay on registe
 </template>
 
 <script setup lang="ts">
-  import { createI18n, Translation } from '@kong-ui-public/i18n'
+  import { createI18n, Translation, i18nTComponent} from '@kong-ui-public/i18n'
   import english from './locales/en.json'
 
   const i18n = createI18n('en-us', english)
-  const i18nT = createI18nTComponent(i18n)
+  const i18nT = i18nTComponent(i18n)
 
 </script>
 
--- a/packages/core/i18n/README.md
+++ b/packages/core/i18n/README.md
@@ -260,7 +260,7 @@ In some cases we do not have access to the Vue `app` and cannot relay on registe
 </template>
 
 <script setup lang="ts">
-  import { createI18n, Translation, i18nTComponent} from '@kong-ui-public/i18n'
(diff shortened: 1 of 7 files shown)


packages/core/i18n/README.md
```diff
@@ -243,6 +245,31 @@ And then, anywhere in application code where `i18n` is needed
 <h1><b>Morning</b>, my name is <i>Val</i>. And I am <i>Money Asker</i>. I want <div class="red">$1,000.00</div></h1>
 ```
 
+#### With direct use of i18nT component
+
+In some cases we do not have access to vue `app`  and cannot relay on registered i18nT plugin. Working on stadalone components in `public-ui-components` is of those cases. And for this your component will look like:
```
adamdehaven, line 250:
```suggestion
In some cases we do not have access to the Vue `app` and cannot relay on registered i18nT plugin. Working on standalone components in `public-ui-components` is of those cases. And for this your component will look like:
```

packages/core/i18n/README.md
```diff
@@ -243,6 +245,31 @@ And then, anywhere in application code where `i18n` is needed
...
+      <a href="https://google.com">Google</a>
+  </i18n-t>
+</template>
+
+<script setup lang="ts">
+  import { createI18n, Translation } from '@kong-ui-public/i18n'
+  import english from './locales/en.json'
+
+  const i18n = createI18n('en-us', english)
+  const i18nT = createI18nTComponent(i18n)
```
adamdehaven, line 267: This `createI18nTComponent` function doesn't exist in the example (missing the import)

packages/core/i18n/src/Translation.ts
```diff
@@ -1,10 +1,14 @@
 import { defineComponent, h } from 'vue'
-import type { VNodeChild, App } from 'vue'
+import type { VNodeChild, App, PropType } from 'vue'
 import type { IntlShapeEx } from './types'
 
-const TranslationComponent = (i18n: IntlShapeEx) => defineComponent({
+export const createI18nTComponent = (i18n: IntlShapeEx | null = null) => defineComponent({
```
adamdehaven, line 5: suggestion: Can we name this function `i18nTComponent` so that to use it, you can do this? 

```ts
const i18nT = i18nTComponent(i18nLocal)
```

adamdehaven: I think having two functions named `create...` might be confusing (I got asked 3 times yesterday how things worked and had to keep pointing them to the docs)

packages/core/i18n/README.md
ValeryG: fixed

packages/core/i18n/src/Translation.ts
ValeryG: done

packages/core/i18n/README.md
```diff
@@ -161,6 +161,8 @@ _Please, do not use this method in any new code. This prevents using parameters/
 
 ### HTML safe formatting with `<i18n-t>`
 
+#### With registered i18nPlugin
```
adamdehaven, line 164:
```suggestion
#### When registered as Vue Plugin
```

packages/core/i18n/sandbox/App.vue
```diff
@@ -25,11 +25,30 @@
           </div>
         </template>
       </i18n-t>
+
+      <p />
+      This is direct use of i18n-t (no plugin registration ):
+      <p />
```
adamdehaven, line 31:
```suggestion
      <p>
      This is direct use of i18n-t (no plugin registration ):
      </p>
```

packages/core/i18n/src/Translation.ts
```diff
@@ -1,10 +1,14 @@
...
 import type { IntlShapeEx } from './types'
 
-const TranslationComponent = (i18n: IntlShapeEx) => defineComponent({
+export const i18nTComponent = (i18n: IntlShapeEx | null = null) => defineComponent({
   name: 'I18nT',
   props: {
+    i18n: {
+      type: Object as PropType<IntlShapeEx>,
+      default: null,
+    },
```
adamdehaven, line 11: question: now that I'm looking at this logic, we always pass in the `i18n` object as a parameter as shown here: 

```ts
const i18nT = i18nTComponent(i18nLocal)
```

So it looks like we never actually pass in the `i18n` prop in any of our documented use-cases because it has to be passed into the function when registered in the component.

**I think this makes sense, because if you used the prop, you would have to pass in the `i18n` object in your `<template>` usage every single time, which would not be ideal.**

So I think we can actually remove the `prop`, and then in the function call above, expect the `i18n: IntlShapeEx` to be a required parameter, meaning we also don't need to fallback to the prop below.

ValeryG: in case of you do old `defaultComponent` way you might need to use i18n as prop
```
<template>
  <i18n-t :i18n="i18n"/>
</template>

<script lang="ts">
import {i18nTcomponent, createI18n} from ... 
import english from ... 

defineComponent({
  components: {i18nTcomponent()},
  setup() {
    return {
      i18n: createI18n(English)
    }
})
</script>

adamdehaven: So then should we document this usage example as well?

ValeryG: sure. documented...

adamdehaven (approved):
lgtm

PR #156, opened 2023-02-17
fix(spec-renderer): correct path for the id to fix scrolling (TDX-2877)

davidma415:
# Summary
Swagger was doing some weird escaping on the path, so this replaces the space with an underscore instead (which is what the ID actually is being rendered with)

## PR Checklist


packages/portal/swagger-ui-web-component/src/utils.js
```diff
@@ -1,18 +1,22 @@
...
+    'operations',
+    operation.tag ? operation.tag : 'default',
+    operation.operationId,
+  ]
+}
 
-export const operationToSwaggerThingId = (operation) => escapeSwaggerThing(
-  operationToSwaggerThingArray(operation).join('-'),
-)
+export const operationToSwaggerThingId = (operation) => {
```
adamdehaven, line 18: suggestion: is it possible to provide a type interface for `operation`?

andrewwylde (changes requested):
mostly nits but one change requested. Thanks for the work!

davidma415:
@andrewwylde @adamdehaven so I was trying to refactor `utils.js` to be a TypeScript file, but I was missing the import of `helpers` from `swagger-client`. Once I added that back in, I was getting build errors since `helpers` isn't typed - hope we're okay with leaving it for now?

adamdehaven (dismissed):

andrewwylde (approved):

andrewwylde (approved):
Gr8, m8

adamdehaven (approved):
🚢

packages/portal/spec-renderer/src/components/SpecDetails.vue
```diff
@@ -80,8 +80,15 @@ const showAndScrollToOperation = (): void => {
     return
   }
 
-  swaggerRef.value.showOperation(props.activeOperation)
-  swaggerRef.value.scrollToOperation(props.activeOperation)
+  const activeOperation = { ...props.activeOperation }
+
+  if (activeOperation && activeOperation.tag) {
+    const newStr: string = decodeURIComponent(activeOperation.tag.trim().replaceAll(/\s/g, '%20'))
```
andrewwylde, line 86: can you just add a note that this is the same logic from the other repo? I know they're tightly coupled.

PR #193, opened 2023-03-02
fix(document-viewer): allow codeblocks to render xml and fix image tags (TDX-2917) (TDX-2920)

davidma415:
# Summary
`CodeBlock` was only escaping characters that would be tags for html language, but changed that logic to apply to all code. Also refactored the `Image` component to not render as a figure, and remove the `<figcaption>`. 

Includes various styling fixes 

## PR Checklist

### DIFF
--- a/packages/portal/document-viewer/src/components/nodes/Image.vue
+++ b/packages/portal/document-viewer/src/components/nodes/Image.vue
@@ -1,14 +1,9 @@
 <template>
-  <figure>
-    <img
-      :alt="alt"
-      :src="url"
-      :title="title"
-    >
-    <figcaption>
-      <slot />
-    </figcaption>
-  </figure>
+  <img
+    :alt="alt"
+    :src="url"
+    :title="title"
+  >
 </template>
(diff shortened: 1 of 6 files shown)


packages/portal/document-viewer/src/components/nodes/Image.vue
```diff
@@ -1,14 +1,9 @@
...
-    >
-    <figcaption>
-      <slot />
-    </figcaption>
-  </figure>
+  <img
+    :alt="alt"
+    :src="url"
+    :title="title"
+  >
```
zekth, line 6:
```suggestion
  />
```
isn't this the right way?

davidma415: Screenshot 2023-03-02 at 7 19 01 AM

TIL

zekth: TIL too

zekth (approved):

PR #213, opened 2023-03-07
chore: add tracking-id to github-star component [KHCP-6451]

ValeryG:
# Summary

adding tracking id for DD_RUM to pickup the click. after https://github.com/Kong/khcp-ui/pull/4471 is applied, click on every element with `data-tracking-id` attribute will be logged with value of this attribute instead of innerHTML: 

```
  <div data-tracking-id="github-star"></div>
```

image


adamdehaven (approved):
lgtm

PR #263, opened 2023-03-20
chore: upload bundle analyzer result as artifact [KHCP-6645]

ValeryG:
# Summary

upload bundle analyzer result as artifact


adamdehaven (approved):
lgtm

PR #273, opened 2023-03-22
fix(metric-cards): increase rounding factor for trend [MA-1404]

mihai-peteu:
# Summary

Noticed an issue in connect where the Metric Card trend returns a neutral gray color. 
This is due to the rounding precision, since the below `Number(0.002343.toFixed(2))` evaluates to `0`:

Screen Shot 2023-03-22 at 8 41 57 AM

This PR bumps the rounding factor to 4 digits.

## PR Checklist


packages/analytics/metric-cards/src/utilities/trend-display.ts
```diff
@@ -0,0 +1,51 @@
+export const DECIMAL_DISPLAY = 2
+export const DECIMAL_ROUNDING_PRECISION = DECIMAL_DISPLAY + 1
```
adorack, line 2: Sorry, I mis-typed in my message to you.  This should be +2; I also think it would help to explain that it's for percentages.

packages/analytics/metric-cards/src/components/MetricCardContainer.cy.ts
```diff
@@ -80,4 +81,66 @@ describe('<MetricCardContainer />', () => {
...
 
   })
+
+  // Should display up arrow icon, and green trend value text
+  it('Upward trend displayed for diff above threshold', () => {
+    cy.mount(MetricCardContainer, {
+      props: {
+        cards: [
+          {
+            currentValue: 0.3014,
```
adorack, line 91: How is this test case working given that currently it's only looking at 3 decimals when determining the change?  Is it because of the `val * 100`?  I'm pretty unclear on the context for this...

adorack (changes requested):

packages/analytics/metric-cards/src/components/MetricCardContainer.cy.ts
mihai-peteu: yes, that's correct. If we don't pass in `val * 100` to `formatChangeFn` this test would fail.

adorack (approved):

PR #320, opened 2023-04-20
chore: catch and fail CI when PR preview publishing fails [KHCP-7061]

ValeryG:
# Summary

fail PR CI with preview publish did not publish anything

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -181,6 +181,11 @@ jobs:
             npm_instructions="${npm_instructions}${pkg}"
           done
 
+          if [[ -z "${npm_instructions}" ]]; then
+            echo "Error publishing packages"
+            exit -1
+          fi
+
           echo "npm_instructions<<EOF" >> $GITHUB_ENV
           echo -e "$npm_instructions" >> $GITHUB_ENV
           echo "EOF" >> $GITHUB_ENV


.github/workflows/ci.yaml
```diff
@@ -170,9 +170,17 @@ jobs:
...
           lerna version prerelease --preid ${preid} --allow-branch ${{ github.head_ref }} --conventional-prerelease --yes --amend
 
-          npm_instructions=""
           echo "//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}" > .npmrc
-          for pkg in $(pnpm --stream --filter "${{needs.install-dependencies.outputs.filter}}" publish  --no-git-checks --access public --report-summary --tag "${tag}" | grep "+ "| sed 's/+ //')
+
+          pnpmPublish=$(pnpm --stream --filter "${{needs.install-dependencies.outputs.filter}}" publish  --no-git-checks --access restricted --report-summary --tag "${tag}")
+
+          if [[ ! -z $(echo "${pnpmPublish}"|grep "There are no new packages that should be published") ]]; then
+            echo "There are no new packages that should be published"
```
adamdehaven, line 178: suggestion: is line `177` checking for a string that is being echoed on line `178`? If yes, can we move the echoed string to a variable above the if statement so that there's no chance they won't match if someone edits the file?

ValeryG: 178 is just random text we are echoing, so 178 and 177 do not need to match

adamdehaven: Oh so `There are no new packages that should be published` comes from lerna. ok

PR #375, opened 2023-05-02
feat: do not run tests if PR and main are identical [KHCP-7173]

ValeryG:
# Summary

- Using `up-to-date` shared action to figure out if tests are needed on push
- each job in workflow takes a time to start and waits for free runner. so we can combine two jobs into one and shave all this time from the workflow run

### DIFF
--- a/.github/actions/up-to-date/action.yml
+++ b/.github/actions/up-to-date/action.yml
@@ -14,10 +14,6 @@ outputs:
 runs:
   using: composite
   steps:
-    - uses: actions/checkout@v2
-      with:
-        fetch-depth: 0
-
     - name: Get PR Info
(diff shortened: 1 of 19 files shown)


adamdehaven (dismissed):

adamdehaven (commented):
Is it possible to start adding comments in these action script sections to make the code more readable for people besides us?

ValeryG:
> Is it possible to start adding comments in these action script sections to make the code more readable for people besides us?

I am all for comments myself but between the step `name` and run definition - it's more or less clear, so I am afraid in this case comments will just pollute and be on the way of reading. but I I'll add some it questionable places.

adamdehaven (approved):

PR #518, opened 2023-06-12
feat: ability to init i18n with IntlConfig params overridden [KHCP-7670]

ValeryG:
# Summary

Details in:  https://kongstrong.slack.com/archives/C04A6LD904E/p1686302933215479


packages/core/i18n/README.md
```diff
@@ -96,6 +98,32 @@ const i18n = useI18n<typeof english>()
 </script>
 ```
 
+### Override config while create instance
+
+For some applications there is a need to create Intl instance with config settings overridden. ONe example of this is when application needs to define error/warning handlers. In this case use `createI18nEx` function.
```
johncowen, line 103:
```suggestion
For some applications there is a need to create Intl instance with config settings overridden. One example of this is when application needs to define error/warning handlers. In this case use `createI18nEx` function.
```

Just a nit/typo I saw

packages/core/i18n/README.md
```diff
@@ -96,6 +98,32 @@ const i18n = useI18n<typeof english>()
...
+
+```ts
+import { createI18nEx } from '@kong-ui-public/i18n'
+import english from './locales/en.json'
+
+...
+
+// this will create Application instance of Intl object
+createI18nEx<typeof english>({
+    locale'en-us',
```
johncowen, line 117:
```suggestion
    locale: 'en-us',
```

Just a little nit/typo

johncowen (approved):
Code here LGTM, guessing you'll want a proper approval from other folks on your team tho.

Just out of interest more than anything, how come the extra function? Back compat with previous versions or something?

Also I've not tried this in place myself, let me know if you want me to before you merge this tho (I just didn't try it as it looks pretty good to me from the sandbox and you've added a test for proof etc)

packages/core/i18n/src/i18n.ts
```diff
@@ -11,11 +11,12 @@ const intlCache = createIntlCache()
 // typed as any since we don't have access to MessageSource here
 let globIntl: any
 
-export const createI18n = <MessageSource extends Record<string, any>>(locale: SupportedLocales, messages: MessageSource, isGlobal: boolean = false): IntlShapeEx<MessageSource> => {
+export const createI18nEx = <MessageSource extends Record<string, any>>(config: IntlConfigEx<MessageSource>, isGlobal: boolean = false): IntlShapeEx<MessageSource> => {
```
adamdehaven, line 14: suggestion: we need to add TS Doc formatted comment above each function here so that users can properly identity which one they want to import.

In my opinion, the two exported functions here are named too similarly. Could we change the name of the new function to something like `createI18nWithConfig` instead? (I think that is what it's actually doing, right?)
```suggestion
/**
 * {Description of this function}
 * @param {IntlConfigEx<MessageSource>} config {Description of config param}
 * @param {boolean} isGlobal {Description of isGlobal param)
 *
*/
export const createI18nEx = <MessageSource extends Record<string, any>>(config: IntlConfigEx<MessageSource>, isGlobal: boolean = false): IntlShapeEx<MessageSource> => {
```

adamdehaven:
@ValeryG as an alternative to adding a new function, could we keep using the **existing** function, but add a third parameter of the config object that defaults to the default config? Then consuming apps wouldn’t have to change their usage, but you could optionally pass in a config object?

adamdehaven:
We could do a breaking change (bumping the major version) to refactor how we pass `messages` and `locale` so that it's part of a config object, and then an app that wants to pass __more__ parameters can choose to do so.

Would this be the better solution so that we don't have two similar functions exported?

ValeryG:
not a big fan of doing breaking changes - this is the case when we can avoid it ... :) here is what I am proposing instead (changes are committed in this PR): 
```ts
// still works , creates local instance (same as before)
createI18n<typeof english>('en-us', english) 

// still works , creates or returns global instance (same as before)
createI18n<typeof english>('en-us', english, true) 

// new way , creates or returns global instance, with possibility to define additional IntlConfig settings
createI18n<typeof english>('en-us', english, {isGLobal: true, onError: (err: any) {console.error(err)}} ) 
```

hope we like it :) 
```

johncowen (commented):
Just had another quick look here as I saw things had changed from your link in Slack Val, thanks for that! This is fine by me, but guessing you want final approval from this team. Wanted to say thanks for doing the extra work here 👍 and just so you know I'd looked again

adamdehaven (approved):
The code LGTM; did not functionally test

PR #565, opened 2023-06-22
feat(analytics-chart): add analytics chart to public-ui-components [MA-1789]

filipgutica:
https://konghq.atlassian.net/browse/MA-1789

# Summary

## PR Checklist


packages/analytics/analytics-chart/vite.config.ts
```diff
@@ -0,0 +1,32 @@
...
+const packageName = 'analytics-chart'
+const sanitizedPackageName = sanitizePackageName(packageName)
+
+// Merge the shared Vite config with the local one defined below
+const config = mergeConfig(sharedViteConfig, defineConfig({
+  build: {
+    lib: {
+      // The kebab-case name of the exposed global variable. MUST be in the format `kong-ui-{package-name}`
+      // Example: name: 'kong-ui-demo-component'
+      name: `kong-ui-${sanitizedPackageName}`,
```
adamdehaven, line 15:
```suggestion
      name: `kong-ui-public-${sanitizedPackageName}`,
```

adamdehaven (changes requested):

packages/analytics/analytics-chart/package.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "@kong-ui-public/analytics-chart",
+  "version": "0.8.6",
```
adamdehaven, line 3: I think this is fine, but we may need to watch the CI on release to ensure it bumps the version from this starting point if that's what is intended

filipgutica: ahh should I just start it from 0.0.1 ?

adamdehaven: Since it's a new package name, you could. Alternatively, maybe bump it to `0.9.0` so you can tell the difference.

adamdehaven (approved):

PR #609, opened 2023-07-13
feat(analytics-chart): lock tooltip on click + max height [MA-1729]

filipgutica:
# Summary

- Implement max height + scroll for tooltips to prevent tooltip from growing too large.
- Freeze tooltip on click in order to be able to interact with it (eg: scrolling)
- Save the last drawn vertical line in the vertical line plugin to continue highlighting which part of the chart the tooltip is currently referring to
- Allow dragging the tooltip when it is locked

https://konghq.atlassian.net/browse/MA-1729

https://loom.com/share/73db510db09341969911598a6731ce9a

## PR Checklist


sichvoge:
@filipgutica I wonder if we should have some help text that highlights you need to click to "fix" the tooltip and then something to explain you can drag + click again on the chart to dismiss the "fix". WDYT?

packages/analytics/analytics-chart/src/components/chart-plugins/ChartTooltip.vue
```diff
@@ -105,15 +182,31 @@ watch(tooltipEl, value => {
...
   min-width: 250px;
-  pointer-events: none;
+  overflow-y: scroll;
   position: absolute;
+  scrollbar-width: thin;
   transition: all 0.3s cubic-bezier(0.25, 0.8, 0.25, 1);
   z-index: 999;
+
+  &::-webkit-scrollbar-track {
+    background-color: var(--white);
```
mihai-peteu, line 192: this might seem like overkill, but can we pls add a fallback:

`background-color: var(--white, #ffffff);`

packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
```diff
@@ -247,6 +252,19 @@ const tooltipDimensions = ({ width, height }) => {
...
 
+const handleChartClick = () => {
+  tooltipData.locked = !tooltipData.locked
+
+  if (chartInstance.value) {
+    if (tooltipData.locked) {
+      verticalLinePlugin.clickedSegment = chartInstance.value.chart.tooltip?.dataPoints[0]
+    } else {
+      verticalLinePlugin.clickedSegment = undefined
+    }
```
mihai-peteu, line 263: can you simplify this into a ternary pls:

```
verticalLinePlugin.clickedSegment = tooltipData.locked
      ? chartInstance.value.chart.tooltip?.dataPoints[0]
      : undefined
```

packages/analytics/analytics-chart/src/composables/useLineChartOptions.ts
```diff
@@ -60,7 +60,7 @@ export default function useLinechartOptions(chartOptions: LineChartOptions) {
 
   Tooltip.positioners.lineChartTooltipPosition = function(elements, position) {
     // Happens when nothing is found
-    if (!elements.length) {
+    if (!elements.length || chartOptions.tooltipState.locked) {
```
mihai-peteu, line 63: Shouldn't we do conditionally check for `tooltipState?.locked` throughout?  Since `locked` is defined as optional in chart-options.ts...

mihai-peteu (changes requested):
Code aesthetics and one functional question

packages/analytics/analytics-chart/src/composables/useLineChartOptions.ts
filipgutica: Why? If you don't use it, it'll just be undefined in which case the locked check here will essentially behave like it's not locked, as expected.

filipgutica: TooltipState?.locked 

This conditional is for the case that tooltipState might be undefined

packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
filipgutica: done

packages/analytics/analytics-chart/src/composables/useLineChartOptions.ts
filipgutica: not necessary

filipgutica:
> 

That's not how scrolling works, padding/margin will only be applied, not to the last "visible" element, but to the real last element that's behind the "scroll"

mihai-peteu (approved):

sichvoge (approved):
Since this is a massive improvement already and would fix a couple bugs, let's get this in. We should review any other improvement with Helen during our analytics-chart review session.

PR #610, opened 2023-07-13
feat(pageheader): new component [khcp-7898]

kaiarrowood:
# Summary

Add new `AppPageHeader` component for [KHCP-7898](https://konghq.atlassian.net/browse/KHCP-7898).
Also adds support to use new design tokens.

Figma: https://www.figma.com/file/8k70HMMdPj98APLaexYJyR/Header?type=design&node-id=11%3A7494&mode=dev

image

## PR Checklist

[KHCP-7898]: https://konghq.atlassian.net/browse/KHCP-7898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
```diff
@@ -0,0 +1,121 @@
...
+      .page-header-title-badge {
+        align-self: center;
+        margin-left: $kui-space-40;
+      }
+
+      .page-header-title {
+        color: $kui-color-text;
+        font-size: $kui-font-size-70;
+        font-weight: $kui-font-weight-bold;
+        letter-spacing: -0.48px;
```
adamdehaven, line 102: I'm not sure it makes sense to alter the letter-spacing here; we want a consistent font display throughout all of the apps.

kaiarrowood: Was going off the Figma designs, I'm ok to ignore it if we don't care

adamdehaven: Let's just go with the default for now; gets rid of "custom font" conversations

package.json
```diff
@@ -24,6 +24,7 @@
     "@commitlint/config-conventional": "^17.6.6",
     "@digitalroute/cz-conventional-changelog-for-jira": "^8.0.1",
     "@evilmartians/lefthook": "^1.4.5",
+    "@kong/design-tokens": "^1.5.1",
```
adamdehaven, line 27: can you bump this to the latest please?

packages/core/app-layout/docs/page-header.md
```diff
@@ -0,0 +1,67 @@
...
+- type: `String`
+- required: `true`
+- default: `''`
+
+The title text of the page.
+
+#### `breadcrumbs`
+
+- type: Array as PropType<BreadcrumbItem[]>
+- required: `false`
```
adamdehaven, line 42: question: why would this prop not be required?

packages/core/app-layout/docs/page-header.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Slots
+
+#### `icon`
+
+Content displayed right before the title text, typically an icon.
+
+#### `badge`
+
+Content displayed right after the title text, typically a badge.
```
adamdehaven, line 55: If these slots can be used for other purposes, then should the slot names be more generic? 

Something like `title-before` and `title-after` ?

packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
```diff
@@ -0,0 +1,124 @@
...
+</template>
+
+<script setup lang="ts">
+import { computed, PropType } from 'vue'
+import type { BreadcrumbItem } from '@kong/kongponents'
+
+const props = defineProps({
+  title: {
+    type: String,
+    default: '',
```
adamdehaven, line 66: nit: if the prop is required, there should not be a default
```suggestion
```

packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
```diff
@@ -0,0 +1,124 @@
...
+    required: true,
+  },
+  breadcrumbs: {
+    type: Array as PropType<BreadcrumbItem[]>,
+    default: () => ([]),
+  },
+})
+
+const hasBreadcrumbs = computed((): boolean => !!props.breadcrumbs.length)
```
adamdehaven, line 75: nit: you need to optionally chain here in case the user passes in `undefined`
```suggestion
const hasBreadcrumbs = computed((): boolean => !!props.breadcrumbs?.length)
```

packages/core/app-layout/docs/page-header.md
kaiarrowood: You can have a `PageHeader` with no breadcrumb section if you want.

packages/core/app-layout/docs/page-header.md
```diff
@@ -0,0 +1,67 @@
...
+- [Requirements](#requirements)
+- [Usage](#usage)
+  - [Install](#install)
+  - [Props](#props)
+  - [Slots](#slots)
+
+## Features
+
+- Reactive updates based on `prop` value changes :rocket:
+- Slottable areas for displaying custom content, icons, etc.
```
adamdehaven, line 15: suggestion: Can you add an item here about the breadcrumbs (and that they are optional)?

packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
kaiarrowood: I can't keep straight when we explicityly provide a default and when we don't. For example, all the entities components explicitly specify default values even for non-required props, but I'll remove it.

packages/core/app-layout/docs/page-header.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Requirements
+
+- `vue` must be initialized in the host application
+- `@kong/kongponents` must be available as a `dependency` in the host application, along with the package's style imports. [See here for instructions on installing Kongponents](https://kongponents.konghq.com/#globally-install-all-kongponents). Specifically, the following Kongponents must be available:
+  - `KBreadcrumb`
```
adamdehaven, line 21: I'd remove some text here as KBreadcrumbs also depends on other components
```suggestion
- `@kong/kongponents` must be available as a `dependency` in the host application, along with the package's style imports. [See here for instructions on installing Kongponents](https://kongponents.konghq.com/#globally-install-all-kongponents).
```

packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
adamdehaven: don't know what people did everywhere else, this is just the way to go

kaiarrowood:
> A couple things I noticed that seem like they need to be addressed in the UI:
> 
> * The spacing between the title and the badge seems too close when compared to the Figma
>   image
> * This component doesn't appear to be ideally responsive below certain viewport sizes. Can you please add some media queries (using the `kui-breakpoint-*` tokens for media queries) and tweak the layout a bit so that the items properly stack on mobile?
> * If the title text is too long, it send the action menu off of the page. I think we need to add a flex container around the title and action menu (or whatever the right-side content is called) so that the title can wrap to a new line and the action menu remains visible on the right
>   image

kaiarrowood:
> A couple things I noticed that seem like they need to be addressed in the UI:
> 
> * The spacing between the title and the badge seems too close when compared to the Figma
>   image

Fixed

> * This component doesn't appear to be ideally responsive below certain viewport sizes. Can you please add some media queries (using the `kui-breakpoint-*` tokens for media queries) and tweak the layout a bit so that the items properly stack on mobile?

Added mobile breakpoint styles

> * If the title text is too long, it send the action menu off of the page. I think we need to add a flex container around the title and action menu (or whatever the right-side content is called) so that the title can wrap to a new line and the action menu remains visible on the right
>   image

Truncating the title in that scenario instead as agreed by design. Thanks for the help fixing this one @portikM !

@adamdehaven ready for re-review

packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
```diff
@@ -0,0 +1,135 @@
...
+  .page-header-title-section {
+    align-items: center;
+    display: flex;
+    gap: $kui-space-40;
+    justify-content: space-between;
+
+    .page-header-title-wrapper {
+      align-items: baseline;
+      display: flex;
+      min-width: 0;
```
adamdehaven, line 92: question: why min-width 0?

packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
```diff
@@ -0,0 +1,135 @@
...
+
+  .page-header-section-below {
+    width: 100%;
+  }
+
+  :deep(.k-breadcrumbs) {
+    margin-bottom: $kui-space-0;
+  }
+
+  @media (max-width: $kui-breakpoint-mobile) {
```
adamdehaven, line 125: question: why are you using max-width media queries? Styles should be written mobile-first 

Also, if you **must** use max-width, you have to subtract one pixel from the variable or there is an overlap in style rules.

packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
kaiarrowood: This is an odd one, needed for truncation with flex. See this article @portikM found here: https://css-tricks.com/flexbox-truncated-text/

packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
portikM: I'm with Adam on this one - writing styles mobile-first is industry approach to creating responsive UIs. Even though it doesn't currently fit our use-case as mobile screen size support is not on table, it will be one day and then we all will have to use one pattern.

packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
portikM: Maybe add a comment there explaining why is that `min-width` rule there @kaiarrowood?

packages/core/app-layout/src/components/pageHeader/AppPageHeader.cy.ts
```diff
@@ -0,0 +1,84 @@
+// Cypress component test spec file
+
+import AppPageHeader from './AppPageHeader.vue'
+
+describe('<AppPageHeader />', () => {
```
adamdehaven, line 5: Since the breadcrumb icon slot logic is directly tied to the KBreadcrumbs component, there should be test coverage here to ensure the slot content renders correctly

kaiarrowood: I already added it... https://github.com/Kong/public-ui-components/pull/610/files#diff-37753584988831e9fb98576d5feac57c4849ede0716863f9e9efe4b62e311c98R56

ValeryG (approved):

jillztom (approved):

PR #655, opened 2023-08-03
Feat/add entities shared

Leopoldthecoder:
# Summary

## PR Checklist


.eslintrc.cjs
```diff
@@ -24,6 +24,7 @@ module.exports = {
     'space-before-function-paren': 'off',
     quotes: ['error', 'single', { avoidEscape: true }],
     'no-multi-spaces': 'error',
+    'no-unused-vars': 'off',
```
Leopoldthecoder, line 27: > The TypeScript and ESLint rules were both enabled; this disables the ESLint variation.

Same as the `shared-ui-components` repo.

packages/entities/entities-shared/package.json
```diff
@@ -0,0 +1,74 @@
...
+      "require": "./dist/entities-shared.umd.js"
+    },
+    "./package.json": "./package.json",
+    "./dist/*": "./dist/*"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "peerDependencies": {
+    "@kong-ui-public/i18n": "workspace:^",
```
adamdehaven, line 23: I believe should have a version in the entry

vite.config.shared.ts
```diff
@@ -3,7 +3,7 @@
 /**
  * Shared Vite config settings for all components
  */
-import { defineConfig } from 'vite'
+import { defineConfig, loadEnv } from 'vite'
```
adamdehaven, line 6: There are other changes required in this file, such as [this line](https://github.com/Kong/shared-ui-components/blob/c2dcf89302f1159af9f65417204bc8035b042f19/vite.config.shared.ts#L109) that allows utilizing the root `.env` file(s)

adamdehaven: We also want to externalize the `i18n` package for all packages like we do on these two lines:
- https://github.com/Kong/shared-ui-components/blob/c2dcf89302f1159af9f65417204bc8035b042f19/vite.config.shared.ts#L72
- https://github.com/Kong/shared-ui-components/blob/c2dcf89302f1159af9f65417204bc8035b042f19/vite.config.shared.ts#L79

If added to this shared config, we first need to ensure it does **not** impact the actual `i18n` package itself.

packages/entities/entities-shared/package.json
Leopoldthecoder: This is actually what I got after running `pnpm --filter "@kong-ui-public/entities-shared" add --save-peer @kong-ui-public/i18n@latest`. Plus, I think this [solves dependency conflicts](https://kongstrong.slack.com/archives/C03ANC23AQK/p1680581451818809?thread_ts=1680529934.588149&cid=C03ANC23AQK).

adamdehaven: Easy way to check is to run the build, then `npm pack` and check the outputted file to ensure it has the correct resolved version

Leopoldthecoder: image

Yes it has.

vite.config.shared.ts
```diff
@@ -121,3 +121,90 @@ export default defineConfig({
...
+      headers: {
+        ...konnectAuthHeader,
+      },
+    },
+
+    /**
+     * KONG MANAGER PROXIES
+     */
+    '^/kong-manager': {
+      target: 'https://kong-nightly-admin-api.connectia.one',
```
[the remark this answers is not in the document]
Leopoldthecoder: You're right, we should move these to env variables. Good thing is they need to be authorized😬

vite.config.shared.ts
Leopoldthecoder: Done. How do we make sure the `i18n` package is not affected? I ran `pnpm build` and it looks fine.

adamdehaven: Easiest way would be to check the bundle analyzer for the package before and after adding this setting

Leopoldthecoder: OK, here's the result:
- For `@kong-ui-public/i18n`, the bundle analyzer outputs are identical before and after adding this setting.
- For `@kong-ui-public/entities-shared`, the `i18n` chunk is in the output before adding this setting and disappears after.

LGTM.

adamdehaven (approved):
LGTM

PR #711, opened 2023-08-16
feat(top-n-table): new component [khcp-8612]

kaiarrowood:
# Summary

Create new `TopNTable` component for [KHCP-8612](https://konghq.atlassian.net/browse/KHCP-8612).

image

image

## PR Checklist

[KHCP-8612]: https://konghq.atlassian.net/browse/KHCP-8612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics/analytics-chart/sandbox/App.vue
+++ b/packages/analytics/analytics-chart/sandbox/App.vue
@@ -257,7 +257,10 @@
       </div>
     </div>
     <div class="config-container">
-      <div class="config-container-row">
+      <div
+        v-if="!isTopNTable"
+        class="config-container-row"
+      >
         <KLabel>Colors</KLabel>
         <div
           v-for="([label, color], i) in Object.entries(colorPalette)"
--- a/packages/analytics/analytics-chart/sandbox/App.vue
+++ b/packages/analytics/analytics-chart/sandbox/App.vue
@@ -11,7 +11,6 @@
           <div>
             <KRadio
               v-model="chartType"
-              data-testid="service-radio-btn"
               name="chartType"
               :selected-value="ChartTypes.HORIZONTAL_BAR"
             >
@@ -21,7 +20,6 @@
           <div>
             <KRadio
               v-model="chartType"
-              data-testid="route-radio-btn"
               name="chartType"
               :selected-value="ChartTypes.VERTICAL_BAR"
             >
@@ -31,7 +29,6 @@
           <div>
             <KRadio
               v-model="chartType"
-              data-testid="application-radio-btn"
               name="chartType"
               :selected-value="ChartTypes.TIMESERIES_LINE"
             >
@@ -41,7 +38,6 @@
           <div>
             <KRadio
               v-model="chartType"
-              data-testid="application-radio-btn"
               name="chartType"
               :selected-value="ChartTypes.TIMESERIES_BAR"
             >
@@ -51,7 +47,6 @@
           <div>
             <KRadio
               v-model="chartType"
-              data-testid="application-radio-btn"
               name="chartType"
               :selected-value="ChartTypes.DOUGHNUT"
             >
@@ -61,11 +56,19 @@
           <div>
             <KRadio
               v-model="chartType"
-              data-testid="application-radio-btn"
               name="chartType"
               :selected-value="ChartTypesSimple.GAUGE"
             >
               Gauge
+            </KRadio>
+          </div>
+          <div>
+            <KRadio
+              v-model="chartType"
+              name="chartType"
+              :selected-value="ChartTypesSimple.TOPN"
+            >
+              Top N Table
             </KRadio>
           </div>
         </div>
@@ -83,7 +86,6 @@
           <div>
             <KRadio
               v-model="metricDisplay"
-              data-testid="service-radio-btn"
               name="metricDisplay"
               :selected-value="ChartMetricDisplay.SingleMetric"
             >
@@ -93,7 +95,6 @@
           <div>
             <KRadio
               v-model="metricDisplay"
-              data-testid="application-radio-btn"
               name="metricDisplay"
               :selected-value="ChartMetricDisplay.Full"
             >
@@ -103,7 +104,6 @@
           <div>
             <KRadio
               v-model="metricDisplay"
-              data-testid="route-radio-btn"
               name="metricDisplay"
               :selected-value="ChartMetricDisplay.Hidden"
             >
@@ -125,7 +125,6 @@
           <div>
             <KRadio
               v-model="legendPosition"
-              data-testid="service-radio-btn"
               name="legendPosition"
               :selected-value="ChartLegendPosition.Right"
             >
@@ -135,7 +134,6 @@
           <div>
             <KRadio
               v-model="legendPosition"
-              data-testid="route-radio-btn"
               name="legendPosition"
               :selected-value="ChartLegendPosition.Bottom"
             >
@@ -145,7 +143,6 @@
           <div>
             <KRadio
               v-model="legendPosition"
-              data-testid="application-radio-btn"
               name="legendPosition"
               :selected-value="ChartLegendPosition.Hidden"
             >
@@ -175,16 +172,17 @@
       tooltip-title="tooltip title"
     />
     <SimpleChart
-      v-else
+      v-else-if="!isTopNTable"
       :chart-data="exploreResult"
       :chart-options="simpleChartOptions"
       :legend-position="legendPosition"
     />
     <TopNTable
-      v-if="exploreResultV3"
+      v-if="isTopNTable"
+      class="top-n-sandbox"
       :data="exploreResultV3"
       description="Last 30-Day Summary"
-      title="Top 5 Status Codes"
+      title="Top 5 Routes"
     >
       <template #name="{ record }">
         <a href="#">{{ record.name }}</a>
@@ -192,7 +190,10 @@
     </TopNTable>
 
     <!-- Dataset options -->
-    <div class="dataset-options">
+    <div
+      v-if="!isTopNTable"
+      class="dataset-options"
+    >
       <KButton
         appearance="outline"
         class="first-button"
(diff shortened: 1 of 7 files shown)


packages/analytics/analytics-chart/sandbox/App.vue
```diff
@@ -11,7 +11,6 @@
           <div>
             <KRadio
               v-model="chartType"
-              data-testid="service-radio-btn"
```
kaiarrowood, line 14: Test ids should be unique and you don't need them in sandbox code

packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -0,0 +1,268 @@
...
+          </div>
+        </div>
+      </div>
+    </template>
+  </KCard>
+</template>
+
+<script setup lang="ts">
+import { computed, PropType } from 'vue'
+import { AnalyticsExploreV3Result, AnalyticsExploreV3Meta, AnalyticsExploreRecord, RecordEvent } from '@kong-ui-public/analytics-utilities'
```
adamdehaven, line 100: fix: import type (this will soon be enforced via https://github.com/Kong/public-ui-components/pull/729)
```suggestion
import type { AnalyticsExploreV3Result, AnalyticsExploreV3Meta, AnalyticsExploreRecord, RecordEvent } from '@kong-ui-public/analytics-utilities'
```

packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -0,0 +1,268 @@
...
+            </div>
+          </div>
+        </div>
+      </div>
+    </template>
+  </KCard>
+</template>
+
+<script setup lang="ts">
+import { computed, PropType } from 'vue'
```
adamdehaven, line 99: fix: type imports (soon to be enforced by https://github.com/Kong/public-ui-components/pull/729)
```suggestion
import { computed } from 'vue'
import type { PropType } from 'vue'
```

packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -0,0 +1,268 @@
...
+  },
+  emptyStateTitle: {
+    type: String,
+    default: '',
+  },
+})
+
+const { i18n } = composables.useI18n()
+const meta = computed((): AnalyticsExploreV3Meta => props.data.meta)
+const records = computed((): AnalyticsExploreRecord[] => props.data.records)
```
adamdehaven, line 126: question: why are these props set to computed properties? The props are already reactive if you just reference them below as needed

packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -0,0 +1,268 @@
...
+const hasData = computed((): boolean => !!(records.value?.length))
+const displayKey = computed((): string => {
+  if (!meta.value?.display) {
+    return ''
+  }
+
+  return Object.keys(meta.value.display)?.[0]
```
adamdehaven, line 133: issue: if this is potentially undefined, then you need to fallback to returning an empty string

packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -0,0 +1,268 @@
...
+})
+const displayRecord = computed(() => {
```
adamdehaven, line 135: suggestion: please add a return type

packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  return meta.value.metricNames[0]
+})
+const columnName = computed((): string => {
+  if (columnKey.value) {
+    return i18n.t(`chartLabels.${columnKey.value}` as any) || columnKey.value
+  }
+
+  return ''
```
adamdehaven, line 154: nit/refactor: can you update the order here to match the other computed?
```suggestion
  if (!columnKey.value) {
    return ''
  }

  return i18n.t(`chartLabels.${columnKey.value}` as any) || columnKey.value
```

packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -0,0 +1,268 @@
...
+    if (!Object.keys(displayRecord.value).length) {
+      return i18n.t('topNTable.errors.display')
+    }
+
+    if (!columnKey.value) {
+      return i18n.t('topNTable.errors.metricNames')
+    }
+  }
+
+  return ''
```
adamdehaven, line 172: nit/refactor: same suggestion about exiting early with a string at the top to match the others

packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -0,0 +1,268 @@
...
+  const id = getId(record)
+
+  return String(displayRecord.value[id]) || '–'
+}
+const getValue = (record: AnalyticsExploreRecord): string => {
+  if (!columnKey.value) {
+    return '–'
+  }
+
+  const event = getEvent(record)
```
adamdehaven, line 194: question: why do you need the `getEvent` method? If you already have access to the `record` in both places where this method is used, why can't you just do `record.event`?

packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  return approxNum(val, { capital: true, round: true })
+}
+</script>
+
+<style lang="scss" scoped>
+.kong-ui-public-top-n-table {
+  --kui-border-radius-20: 8px;
+  --kui-font-size-60: 16px;
```
adamdehaven, line 208: question: why are we overriding tokens here? 

For the border-radius, they likely used the "new" border-radius in the Figma, but you should just use the default for now. When Kongponents are updated, it will inherit the correct styles (it's ok if it differs from the Figma since it matches our existing styles).

For the font-size, why the override?

packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -0,0 +1,268 @@
...
+
+  if (!val) {
+    return '–'
+  }
+
+  return approxNum(val, { capital: true, round: true })
+}
+</script>
+
+<style lang="scss" scoped>
```
adamdehaven, line 205: For all of the styles below, can you utilize `@kong/design-tokens` SCSS variables for the values that exist? e.g. `font-size`, `line-height`, `space`, etc.?

packages/analytics/analytics-chart/src/locales/en.json
```diff
@@ -110,5 +110,15 @@
...
     "Weekly": "@timestamp per week"
+  },
+  "topNTable": {
+    "nameLabel": "Name",
+    "defaultEmptyStateTitle": "No data to display",
+    "defaultErrorStateTitle": "Something went wrong",
+    "errors": {
+      "meta": "Error: response must contain `meta`",
+      "display": "Error: `display` response incorrectly formatted (must have at least one key)",
+      "metricNames": "Error: `metricNames` response incorrectly formatted (must have at least one key)"
```
adamdehaven, line 121: issue: you should use single quotes `'` here instead of backticks; it can cause issues in some scenarios

adamdehaven (changes requested):

packages/analytics/analytics-chart/src/components/TopNTable.vue
kaiarrowood: They are computeds just in an attempt to simplify the reuse below so I can say `meta.value` instead of `props.data.meta`. If there is a drawback to setting them as computeds I can remove.

adamdehaven: I'd recommend just referencing the props directly; there's no real point in making Vue go through extra cycles since it's 1:1

packages/analytics/analytics-chart/src/components/TopNTable.vue
kaiarrowood: I didn't do this because it's correctly implied and from talking to the data team, it sounds like this return type will eventually be more specifically typed in the API. If I hard code the type to this generic `Record<string,string>` it just means I would have to change it again down the line after they change it in the API. 

image

packages/analytics/analytics-chart/src/locales/en.json
adorack: When would this cause issues?

packages/analytics/analytics-chart/src/components/TopNTable.vue
kaiarrowood: That's old code from my initial implementation, you're right. I can remove.

packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -0,0 +1,268 @@
...
+                  id: getId(entry),
+                  name: getName(entry),
+                  dimension: displayKey,
+                }"
+              >
+                {{ getName(entry) }}
+              </slot>
+            </div>
+            <div class="column-2">
+              &nbsp; {{ getValue(entry) }}
```
adorack, line 89: Out of curiosity, why do you have the `&nbsp;` here?  Trying to keep the column from collapsing under certain circumstances?

adorack (commented):
Looks good, thank you!  I'm digging into `khcp-ui` to see if I can quickly try it out, but I don't think that's a blocker for merging it.

packages/analytics/analytics-chart/src/locales/en.json
adamdehaven: TBH I can't remember the scenario

adamdehaven: but I know we removed them (on purpose) from `en.json` in `khcp-ui` at some point

kaiarrowood:
> When would this cause an issue.
https://github.com/Kong/public-ui-components/pull/711#discussion_r1301771782

@adorack I assume you mean, why do we have an error if there is no `metricNames`. The first entry in this array is the `key` I use to display the column 2 data - without it I can't display the counts.

adorack (approved):
It doesn't look like we're set up yet to try plugging this in to the UI, but the overall code / layout look good to me.

PR #714, opened 2023-08-17
test(entities-upstreams-targets): fix flaky tests and test results upload workflow

nekolab:
The `cy.wait()` will wait intercepted network events for 5 seconds by default. But in the CI environment, the
`['@getUpstream', '@fetchServices', '@fetchCertificates']` combination will took more than 4 seconds so the test will be flaky if it occasionally execeeds the 5 seconds limitation.

The slow part seems caused by the Cypress's vm hooks. It does not affect the component in development/production mode. So it's safe to loose the timeout to 10 seconds

This commit also fix the test results upload failure. The artifact name could not contains some special characters like `/`, which will occur in `matrix.container`. A new step is added to normalize the artifact name to make sure it can be uploaded successfully. (Verified [here](https://github.com/Kong/public-ui-components/actions/runs/5893716306))

# Summary

## PR Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -312,12 +312,21 @@ jobs:
           install: false
           command: pnpm run test:component:ci --spec ${{ matrix.container }}
 
+      - name: Normalize test results artifact name
+        continue-on-error: true
+        if: failure()
+        id: normalize-name
+        shell: bash
+        run: |
+          packageName=$(echo ${{ matrix.container }} | sed 's/\//-/g')
+          echo "artifactName=component-test-failures-${packageName}" >> $GITHUB_OUTPUT
+
       - name: Upload Component Test Results
         uses: actions/upload-artifact@v3
         continue-on-error: true
         if: failure()
         with:
-          name: component-test-failures-${{ matrix.container }}
+          name: ${{ steps.normalize-name.outputs.artifactName }}
           path: |
             cypress/videos/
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -312,12 +312,21 @@ jobs:
...
+          packageName=$(echo ${{ matrix.container }} | sed 's/\//-/g')
+          echo "artifactName=component-test-failures-${packageName}" >> $GITHUB_OUTPUT
+
       - name: Upload Component Test Results
         uses: actions/upload-artifact@v3
         continue-on-error: true
         if: failure()
         with:
-          name: component-test-failures-${{ matrix.container }}
+          name: ${{ steps.normalize-name.outputs.artifactName }}
```
adamdehaven, line 329: @nekolab can you create a copy of the artifact name changes from here and PR also into `shared-ui-components`?

adamdehaven (approved):
LGTM; please see my CI comment

.github/workflows/ci.yaml
nekolab: Yes! I will create a PR tomorrow

nekolab: I've checked the `shared-ui-components`. It already has a similar mechanism `component-test-failures-${{ steps.prep.outputs.pkgResName }}` in the CI workflow

PR #729, opened 2023-08-22
chore(tsconfig): apply `@vue/tsconfig`

nekolab:
Use an exported `const enum` in vite is forbidden due to the [compiler limitation](https://vitejs.dev/guide/features.html#isolatedmodules). To fix this, some `const enum` exports need to be reshape to `enum`.

Besides the `const enum`, we also need to use the Type-Only Imports and Export syntax to avoid potential problems like type-only imports being incorrectly bundled. See https://vitejs.dev/guide/features.html#transpile-only

These restrictions are all included in `@vue/tsconfig`, so we applied it to the root `tsconfig.json` to prevent further issues.

# Summary

## PR Checklist


adamdehaven (changes requested):
Blocking merge until I have time to dig into what is being proposed in this PR

tsconfig.json
```diff
@@ -29,12 +25,5 @@
...
     "node_modules",
     "packages/*/dist"
-  ],
-  "ts-node": {
-    // these options are overrides used only by ts-node
-    // same as the --compilerOptions flag and the TS_NODE_COMPILER_OPTIONS environment variable
-    "compilerOptions": {
-      "module": "commonjs"
-    }
-  }
```
adamdehaven, line 39: question: why is this being removed? This was specifically for the CLI

.eslintrc.cjs
```diff
@@ -45,6 +45,7 @@ module.exports = {
     }],
     '@typescript-eslint/no-explicit-any': 'off',
     '@typescript-eslint/indent': ['error', 2],
+    '@typescript-eslint/consistent-type-imports': 'error',
```
adamdehaven, line 48: Can you please add the `fixStyle: 'separate-type-imports'` to the rule?

adamdehaven (changes requested):

tsconfig.json
adamdehaven: As long as the CLI `create-package` still works, this should be fine

tsconfig.json
```diff
@@ -1,15 +1,11 @@
 {
+  "extends": "@vue/tsconfig/tsconfig",
```
adamdehaven, line 2: Fully resolve the import as in the docs
```suggestion
  "extends": "@vue/tsconfig/tsconfig.json",
```

tsconfig.json
nekolab: Oh, I thought it was used for cypress. But since `create-package` seems ok I suggest remove this block to keep tsconfig simple

adamdehaven: Sounds good

adamdehaven (approved):
lgtm

sumimakito (approved):
LGTM

PR #736, opened 2023-08-23
fix(swagger-ui-web-component): open section if collapsed for side nav (TDX-3313)

davidma415:
# Summary
The bug was because the div is not visible if the parent container is not open. This bug fix opens the parent container, which then allows us to select the operation that we want to find, and then scrolls to it.

## PR Checklist

### DIFF
--- a/packages/portal/swagger-ui-web-component/src/element.js
+++ b/packages/portal/swagger-ui-web-component/src/element.js
@@ -223,17 +223,17 @@ export class SwaggerUIElement extends HTMLElement {
     }
 
     const operationElementId = operationToSwaggerThingId(operation)
-    let element = this.shadowRoot.getElementById(operationElementId)
-    if (!element) {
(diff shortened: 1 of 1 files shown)


packages/portal/swagger-ui-web-component/src/element.js
```diff
@@ -223,9 +223,22 @@ export class SwaggerUIElement extends HTMLElement {
     }
 
     const operationElementId = operationToSwaggerThingId(operation)
-    const element = this.shadowRoot.getElementById(operationElementId)
+    let element = this.shadowRoot.getElementById(operationElementId)
```
adamdehaven, line 226: suggestion: if it doesn’t have a huge blast radius, could you rename the `element` var to something that isn’t a primitive? Not a huge deal

adamdehaven (dismissed):

adamdehaven (approved):

PR #782, opened 2023-09-13
feat(about-section): new component [khcp-8589]

kaiarrowood:
# Summary

New `AppAboutSection` component for [KHCP-8589](https://konghq.atlassian.net/browse/KHCP-8589).
Also minor fixes for `TopNTable`:
- Fix `KCard` styles
- Add `isLoading` prop

Figma: https://www.figma.com/file/8UBIeLOY3YX3oF0RtDdwau/About-Section?type=design&node-id=55%3A9902&mode=dev

image

## PR Checklist

[KHCP-8589]: https://konghq.atlassian.net/browse/KHCP-8589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/app-layout/src/components/aboutSection/AppAboutSection.vue
```diff
@@ -0,0 +1,186 @@
...
+    default: '',
+  },
+  modified: {
+    type: String,
+    default: '',
+  },
+  modifiedLabel: {
+    type: String,
+    default: '',
+  },
```
adamdehaven, line 100: question: why would the `createdLabel` or `modifiedLabel` ever be different text?

packages/core/app-layout/src/components/aboutSection/AppAboutSection.vue
```diff
@@ -0,0 +1,186 @@
...
+<style lang="scss">
+.kong-ui-app-about-section.kong-card.border {
+  border-radius: $kui-border-radius-20;
+  padding: $kui-space-70;
+
+  .k-card-header {
+    align-items: baseline;
+    margin-bottom: $kui-space-0 !important;
+  }
+}
```
adamdehaven, line 185: question: why are we overriding KCard here? If it's just to match the Figma, can we please remove this override and when the Kongponent is reskinned, it will automatically inherit the correct styles.

adamdehaven (commented):
The responsiveness overall looks great, but one issue with the created/modified timestamps

image

Can we drop these to a new line, under the title, on smaller screens?

adamdehaven:
image


I’m worried a bit about the amount of content being displayed at the bottom of the component. Is this accounted for in the responsive style rules?

Obviously mobile is going to be a stretch, but even smaller laptop screens are going to have to wrap/stack some of this content.

Could you update the sandbox to see what this looks like and then we can inspect on smaller screens?

packages/core/app-layout/src/components/aboutSection/AppAboutSection.vue
adamdehaven: If that's the case, should the prop names for `created/modified` be more generic?

e.g. `createdLabel: 'Joined'` seems a bit confusing

packages/core/app-layout/src/components/aboutSection/AppAboutSection.vue
adamdehaven: just add a comment to remove overrides once the reskin happens?

packages/core/app-layout/src/components/aboutSection/AppAboutSection.vue
```diff
@@ -0,0 +1,186 @@
+<template>
+  <KCard class="kong-ui-app-about-section">
+    <template
+      v-if="title"
+      #title
+    >
+      <div class="about-section-title">
+        {{ title }}
+      </div>
```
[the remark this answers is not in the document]
kaiarrowood: We are inside `KCard`'s `title` slot, this `div` (which I changed to a `span`) gets created in an `<h4>`

image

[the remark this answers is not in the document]
adamdehaven: `span` 👍🏼

packages/core/app-layout/src/components/aboutSection/AppAboutSection.vue
kaiarrowood: Do you have a suggestion, it maps to `created_at`/`updated_at` fields most of the time.

packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -233,18 +243,22 @@ const getValue = (record: AnalyticsExploreRecord): string => {
     }
 
     .column-1 {
-      padding: 0 $kui-space-80 $kui-space-60 0;
+      padding: 0 $kui-space-80 $kui-space-50 0;
```
adamdehaven, line 246: nit: can you use `$kui-space-0` for the `0` values? Whole PR

packages/core/app-layout/src/components/aboutSection/AppAboutSection.vue
```diff
@@ -0,0 +1,241 @@
...
+    default: 'Modified',
+  },
+  isLoading: {
+    type: Boolean,
+    default: false,
+  },
+})
+
+const displayModified = computed(() => {
+  return !!props.modified && props.modified !== props.created
```
adamdehaven, line 135: suggestion: if you add a check here that `props.created` exists, you can also change the `v-if="created && displayModified"` to just `v-if="displayModified"` in the template above

packages/core/app-layout/src/components/aboutSection/AppAboutSection.vue
```diff
@@ -0,0 +1,241 @@
...
+  .about-section-header-end {
+    display: flex;
+
+    .about-section-timestamps {
+      align-self: center;
+      color: $kui-color-text-neutral;
+      font-size: $kui-font-size-20;
+      line-height: $kui-line-height-20;
+
+      @media (max-width: $kui-breakpoint-phablet) {
```
adamdehaven, line 158: question: why are we using max-width instead of min-width? (mobile first 😄 )

If you need `max-width` you need to subtract a pixel
```suggestion
      @media (max-width: ($kui-breakpoint-phablet - 1px)) {
```

kaiarrowood: I copy pastad and always forget which direction we're supposed to go. I will fix.

packages/core/app-layout/src/components/aboutSection/AppAboutSection.vue
kaiarrowood: Design team mentioned they may decide to only show `modified` on some pages, so I may need `displayModified` to be `true` when `props.created` doesn't exist.

jillztom (approved):

PR #789, opened 2023-09-14
fix(analytics-chart): better legend label truncation [MA-2048]

filipgutica:
https://konghq.atlassian.net/browse/MA-2048

# Summary

Only truncate the label when the legend grows such that it begins to wrap to multiple rows.

calculate the number of row in the grid layout, representing
the legend by comparing the "top" position of each item
incrementing the numberOfRows each time there is a difference.

Use the number of rows to determine if a label should be
truncated or not.

## Truncate when list grows and wraps to a new row.
https://www.loom.com/share/c338d6643f9d4f46baf49de6245e468a

## Truncate on window resize if legend is forced into a new row.
https://www.loom.com/share/bf2b6c0d7a03428ea290f3059fe6d01d

## PR Checklist

### DIFF
--- a/packages/analytics/analytics-chart/src/components/chart-plugins/ChartLegend.vue
+++ b/packages/analytics/analytics-chart/src/components/chart-plugins/ChartLegend.vue
@@ -1,12 +1,13 @@
 <template>
   <ul
     v-if="position !== ChartLegendPosition.Hidden"
+    ref="legendContainerRef"
     class="legend-container"
     :class="positionToClass(position)"
     data-testid="legend"
   >
     <li
-      v-for="{ fillStyle, strokeStyle, text, datasetIndex, index, value } in (itemsRef as any[])"
+      v-for="{ fillStyle, strokeStyle, text, datasetIndex, index, value } in (items as any[])"
       :key="text"
       @click="handleLegendItemClick(datasetIndex, index)"
     >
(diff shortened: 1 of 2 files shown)


packages/analytics/analytics-chart/src/components/chart-plugins/ChartLegend.vue
```diff
@@ -54,8 +57,19 @@ const props = defineProps({
...
+  if (legendContainerRef.value) {
+    const scrollHeight = legendContainerRef.value.scrollHeight
+    if (scrollHeight > parseInt(KUI_SPACE_150, 10)) {
+      shouldTruncate.value = true
+    }
+  }
+}
 
-const itemsRef = toRef(props, 'items')
+watch(() => props.items, checkForWrap, { immediate: true })
```
adorack, line 72: I think this should have a `flush: 'post`` -- experimenting with the sandbox, it doesn't actually trigger wrapping the instant that a dataset starts to trigger scrolling.  At the time this `watch` runs, the DOM hasn't updated yet.

adorack (changes requested):
It looks like there's something a bit off with the alignment of long labels?

image

packages/analytics/analytics-chart/src/components/chart-plugins/ChartLegend.vue
filipgutica: that sounds good

filipgutica:
> It looks like there's something a bit off with the alignment of long labels?
> 
> image

I'm not sure I follow regarding the alignment.

image

mihai-peteu (approved):

mihai-peteu (approved):

PR #817, opened 2023-10-02
feat(config-card-display): json-yaml format [khcp-9212]

aanchalm01:
# Summary
Addresses:
[https://konghq.atlassian.net/browse/KHCP-9212](https://konghq.atlassian.net/browse/KHCP-9212)

Test PR in Konnect: [https://github.com/Kong/khcp-ui/pull/6910](https://github.com/Kong/khcp-ui/pull/6910)
`FF: khcp-8778-json-yaml-configurations`

## PR Checklist

### DIFF
--- a/packages/entities/entities-certificates/src/components/CACertificateConfigCard.vue
+++ b/packages/entities/entities-certificates/src/components/CACertificateConfigCard.vue
@@ -67,7 +67,7 @@ import type { PropType } from 'vue'
 import { computed, ref } from 'vue'
 import type { AxiosError } from 'axios'
 import type { KongManagerCertificateEntityConfig, KonnectCertificateEntityConfig, CACertificateConfigurationSchema, EntityRow } from '../types'
-import { EntityBaseConfigCard, ConfigurationSchemaType, ConfigurationSchemaSection, useStringHelpers, ConfigCardItem } from '@kong-ui-public/entities-shared'
+import { EntityBaseConfigCard, ConfigurationSchemaType, ConfigurationSchemaSection, useStringHelpers, useHelpers, ConfigCardItem } from '@kong-ui-public/entities-shared'
 import endpoints from '../ca-certificates-endpoints'
 import composables from '../composables'
 import '@kong-ui-public/entities-shared/dist/style.css'
@@ -106,6 +106,7 @@ const props = defineProps({
 
 const { i18n: { t, formatUnixTimeStamp } } = composables.useI18n()
 const { convertKeyToTitle } = useStringHelpers()
+const { getPropValue } = useHelpers()
 const { getCertificateData } = composables.useCertificate()
 
 const fetchUrl = computed((): string => endpoints.form[props.config.app].edit)
@@ -120,10 +121,6 @@ const parsedCertData = computed(() => {
 const issuer = computed((): string => parsedCertData.value?.schemaIssuer || '')
 const expiry = computed((): string => parsedCertData.value?.schemaExpiry ? formatUnixTimeStamp(parsedCertData.value?.schemaExpiry) : '')
 const keyUsages = computed((): string[] => parsedCertData.value?.schemaKeyUsages || [])
-
-const getPropValue = (propName: string, slotProps?: Record<string, any>) => {
-  return slotProps?.[propName] || undefined
-}
 
 const getPropItemValue = (propKey: string, slotProps?: Record<string, any>) => {
   const propValue = getPropValue('rowValue', slotProps)
--- a/packages/entities/entities-certificates/src/components/CACertificateConfigCard.vue
+++ b/packages/entities/entities-certificates/src/components/CACertificateConfigCard.vue
@@ -10,23 +10,30 @@
       @fetch:success="handleFetch"
       @loading="(val: boolean) => $emit('loading', val)"
     >
-      <template #cert="{ rowValue }">
+      <template #cert="slotProps">
         <KCodeBlock
+          v-if="getPropValue('rowValue', slotProps)"
           :id="`ca-cert-${config.entityId}-cert-codeblock`"
-          :code="rowValue"
+          :code="getPropValue('rowValue', slotProps)"
           is-single-line
           language="plaintext"
         />
       </template>
-      <template #metadata="{ rowValue }">
+      <template #metadata-label>
+        <KLabel class="metadata-label">
+          Metadata
+        </KLabel>
+      </template>
+
+      <template #metadata="slotProps">
         <ConfigCardItem
-          v-for="propKey in Object.keys(rowValue).filter((prop: string) => !HIDDEN_METADATA.includes(prop))"
+          v-for="propKey in Object.keys(getPropValue('rowValue', slotProps)).filter((prop: string) => !HIDDEN_METADATA.includes(prop))"
           :key="propKey"
           :data-testid="`ca-cert-metadata-${propKey}`"
           :item="{
             key: propKey,
             label: convertKeyToTitle(propKey),
-            value: propKey === 'expiry' ? expiry : propKey === 'key_usages' ? keyUsages : rowValue[propKey],
+            value: getPropItemValue(propKey, slotProps),
             type: propKey === 'key_usages' ? ConfigurationSchemaType.BadgeTag : ConfigurationSchemaType.Text
           }"
         />
@@ -114,6 +121,26 @@ const issuer = computed((): string => parsedCertData.value?.schemaIssuer || '')
 const expiry = computed((): string => parsedCertData.value?.schemaExpiry ? formatUnixTimeStamp(parsedCertData.value?.schemaExpiry) : '')
 const keyUsages = computed((): string[] => parsedCertData.value?.schemaKeyUsages || [])
 
+const getPropValue = (propName: string, slotProps?: Record<string, any>) => {
+  return slotProps?.[propName] || undefined
+}
+
+const getPropItemValue = (propKey: string, slotProps?: Record<string, any>) => {
+  const propValue = getPropValue('rowValue', slotProps)
+
+  if (propKey === 'expiry') {
+    return expiry.value
+  } else if (propKey === 'key_usages') {
+    return keyUsages.value
+  }
+
+  if (propValue) {
+    return propValue[propKey]
+  }
+
+  return undefined
+}
+
 const handleFetch = (entity: Record<string, any>) => {
   record.value = entity as EntityRow
   emit('fetch:success', entity)
@@ -167,6 +194,10 @@ const configSchema = ref<CACertificateConfigurationSchema>({
 .kong-ui-ca-certificate-entity-config-card {
   :deep(.config-badge) {
     margin-right: $kui-space-20;
+  }
+
+  .metadata-label {
+    font-size: $kui-font-size-40;
   }
 }
 </style>
--- a/packages/entities/entities-shared/docs/config-card-display.md
+++ b/packages/entities/entities-shared/docs/config-card-display.md
@@ -1,0 +1,63 @@
+# ConfigCardDisplay.vue
+
+A wrapper component for flexible display of entity configuration fields.
+
+- [Requirements](#requirements)
+- [Usage](#usage)
+  - [Install](#install)
+  - [Props](#props)
+  - [Usage example](#usage-example)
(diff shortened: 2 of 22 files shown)


packages/entities/entities-shared/src/components/entity-base-config-card/EntityBaseConfigCard.vue
```diff
@@ -187,6 +256,30 @@ const { axiosInstance } = composables.useAxios({
...
+    label: 'JSON',
+    value: 'json',
+  },
+  {
+    label: 'YAML',
+    value: 'yaml',
+  },
+]
+
+const konnectJsonYamlEnabled = computed(() => props.config.app === 'konnect' && props.config.jsonYamlEnabled)
```
ValeryG, line 275: what is the purpose of computed there?

aanchalm01: I am looking at this now and think I can use a method call instead of computed here since we don't require caching for this variable. What do you think ?

packages/entities/entities-shared/src/components/entity-base-config-card/ConfigCardDisplay.vue
```diff
@@ -0,0 +1,137 @@
...
+import yaml from 'js-yaml'
+
+const props = defineProps({
+  item: {
+    type: Object,
+    required: false,
+    default: () => null,
+  },
+  format: {
+    type: String,
```
adamdehaven, line 92: suggestion: you can type this a little stronger if these are the only valid values
```suggestion
    type: String as PropType<'structured' | 'yaml' | 'json'>,
```

packages/entities/entities-shared/src/components/entity-base-config-card/ConfigCardDisplay.vue
```diff
@@ -0,0 +1,137 @@
...
+    required: false,
+    default: 'structured',
+    validator: (val: string) => ['structured', 'yaml', 'json'].includes(val),
+  },
+  propListTypes: {
+    type: Array as PropType<String[]>,
+    required: false,
+    default: () => ([]),
+  },
```
adamdehaven, line 101: question: this simply accepts an array of strings -- can the user pass any string? If not, can we provide an interface here (similar to the suggestion for the `format` prop)

packages/entities/entities-shared/src/components/entity-base-config-card/ConfigCardDisplay.vue
```diff
@@ -0,0 +1,137 @@
...
+const jsonContent = ref('')
+const yamlContent = ref('')
+
+watch(() => props.format, () => {
+  if (props.format === 'json') {
+    jsonContent.value = JSON.stringify(props.record, null, 2)
+  }
+  if (props.format === 'yaml') {
+    yamlContent.value = yaml.dump(props.record)
+  }
```
adamdehaven, line 123: suggestion: utilize a function param, and then reference the param inside the watcher instead of `props.format`

Also, since it can only be one or the other, use if/else
```suggestion
watch(() => props.format, (format: 'structured' | 'yaml' | 'json') => {
  if (format === 'json') {
    jsonContent.value = JSON.stringify(props.record, null, 2)
  } else if (format === 'yaml') {
    yamlContent.value = yaml.dump(props.record)
  }
```

adamdehaven, line 123: There is no logic here for `structured` -- assuming that's intentional

packages/entities/entities-shared/src/components/entity-base-config-card/ConfigCardDisplay.vue
```diff
@@ -0,0 +1,137 @@
...
+
+}, { immediate: true })
+
+</script>
+
+<style lang="scss" scoped>
+.config-card-prop-section-title {
+  color: $kui-color-text;
+  font-size: $kui-font-size-40;
+  font-weight: 600;
```
adamdehaven, line 133: nit: use token
```suggestion
  font-weight: $kui-font-weight-semibold;
```

packages/entities/entities-shared/src/components/entity-base-config-card/EntityBaseConfigCard.vue
```diff
@@ -17,7 +17,38 @@
...
+          v-if="showBookButton"
+          appearance="btn-link"
+          class="book-icon"
+          @click="handleDocumentationClick"
+        >
+          <template #icon>
+            <KIcon
+              color="#0044f4"
+              icon="book"
+            />
```
adamdehaven, line 44: suggestion: instead of using `KIcon` can you please import the icon from `@kong/icons`

packages/entities/entities-shared/src/components/entity-base-config-card/EntityBaseConfigCard.vue
adamdehaven: I think computed should be fine since it's evaluating two props

packages/entities/entities-shared/src/components/entity-base-config-card/ConfigCardDisplay.vue
aanchalm01: Yes, because for `structured` we use `ConfigCardItem` component to display the content.

packages/entities/entities-shared/src/components/entity-base-config-card/ConfigCardDisplay.vue
aanchalm01: I see this error when I add the type check, I tried doing v-if but that didnt resolve it. 
image 1592

adamdehaven: Ew 😢 

Ok just make sure the validator covers it; then just resolve

packages/entities/entities-shared/src/components/entity-base-config-card/ConfigCardDisplay.vue
aanchalm01: Didnt update this as this was giving same issue like `format` prop.

packages/entities/entities-shared/src/components/entity-base-config-card/ConfigCardDisplay.vue
aanchalm01: This gives an error as well: `No overload matches this call. The last overload gave the following error.ts(2769)`;
so updated this to `watch(() => props.format, (format: string) => { `instead of `watch(() => props.format, (format: 'structured' | 'yaml' | 'json') => {`

packages/entities/entities-shared/docs/config-card-display.md
```diff
@@ -0,0 +1,63 @@
...
+
+The base item configuration when format is structured.
+
+#### `format`
+
+- type: `String`
+- required: `false`
+- default: `'structured'`
+
+Format to be displayed in the Config Card.
```
kaiarrowood, line 39:
```suggestion
Format to be displayed in the Config Card. Can be `structured`, `json`, or `yaml`.
```

packages/entities/entities-shared/docs/config-card-display.md
```diff
@@ -0,0 +1,63 @@
...
+
+The base record configuration when format is either JSON or YAML.
+
+### Usage example
+
+Please refer to the [sandbox](../sandbox/pages/ConfigCardDisplayPage.vue).
+
+## TypeScript interfaces
+
+Currently we don't have any typescript interfaces exported from this component
```
kaiarrowood, line 63:
```suggestion
Currently, we don't have any typescript interfaces exported from this component.
```

packages/entities/entities-certificates/src/components/CACertificateConfigCard.vue
```diff
@@ -114,6 +121,26 @@ const issuer = computed((): string => parsedCertData.value?.schemaIssuer || '')
 const expiry = computed((): string => parsedCertData.value?.schemaExpiry ? formatUnixTimeStamp(parsedCertData.value?.schemaExpiry) : '')
 const keyUsages = computed((): string[] => parsedCertData.value?.schemaKeyUsages || [])
 
+const getPropValue = (propName: string, slotProps?: Record<string, any>) => {
```
kaiarrowood, line 124: Can you make this method available in a composable so you don't need to duplicate it in each entity?

aanchalm01: Updated.

kaiarrowood (dismissed):
LGTM! 🚀

packages/entities/entities-shared/src/components/entity-base-config-card/EntityBaseConfigCard.vue
```diff
@@ -187,6 +278,29 @@ const { axiosInstance } = composables.useAxios({
   headers: props.config?.requestHeaders,
 })
 
+const configFormatItems = [
+  {
+    label: 'Structured',
```
Leopoldthecoder, line 283: Can `'Structured'` be internationalized? `'JSON'` and `'YAML'` are fine because I think they are universal in all languages.

packages/entities/entities-shared/src/components/entity-base-config-card/EntityBaseConfigCard.vue
```diff
@@ -176,6 +243,30 @@ const props = defineProps({
...
+    required: false,
+  },
+  /**
+   * Boolean to determine whether to display the documentation button
+   */
+  hideConfigCardDoc: {
+    type: Boolean,
+    default: false,
+    required: false,
+  },
```
Leopoldthecoder, line 261: Do we need another prop to determine whether to display the documentation button? We already have `configCardDoc`, so I think we can control the visibility of the documentation button by checking if `configCardDoc` is undefined or an empty string.

packages/entities/entities-certificates/src/components/CACertificateConfigCard.vue
```diff
@@ -10,23 +10,30 @@
       @fetch:success="handleFetch"
       @loading="(val: boolean) => $emit('loading', val)"
     >
-      <template #cert="{ rowValue }">
+      <template #cert="slotProps">
```
Leopoldthecoder, line 13: Question: why are we changing `{ rowValue }` and `{ row }` to `slotProps` and `getPropValue`? IMHO `{ rowValue }` is more straightforward.

packages/entities/entities-shared/src/components/entity-base-config-card/EntityBaseConfigCard.vue
Leopoldthecoder: Sorry I wasn't clear before. What I meant was since we already have `configCardDoc`, do we still need `hideConfigCardDoc`? We can control the button's visibility by checking `configCardDoc` and remove `hideConfigCardDoc`.

aanchalm01: Added both just in case to not run into issue if someone didnt provide the doc url and still wanted to show the button. I can remove `hideConfigCardDoc` if it's not making sense here.

packages/entities/entities-certificates/src/components/CACertificateConfigCard.vue
kaiarrowood: The problem is `slotProps` can contain either `rowValue` or `row` or both or neither and TS isn't smart enough to determine which properties will be available to it. In order to avoid TS squawking at us we needed to move the logic that pulls the prop out of the slot props out of the template.

packages/entities/entities-shared/src/components/entity-base-config-card/EntityBaseConfigCard.vue
aanchalm01: I will remove the `hideConfigCardDoc`.

kaiarrowood (approved):
LGTM! 🚀

packages/entities/entities-certificates/src/components/CACertificateConfigCard.vue
Leopoldthecoder: Got it. Thank you both!

Leopoldthecoder (approved):
LGTM! Thank you for this feature!

PR #837, opened 2023-10-14
fix(service): interpreting 0 as a boolean true

Leopoldthecoder:
# Summary

This PR fixes a bug where zero values are interpreted as `true`.

Before:
```ts
// when data.read_timeout is 0, the following evaluates to true
(data?.read_timeout || data?.read_timeout === 0) ?? 60000
```

After:
```ts
// when data.read_timeout is 0, the following evaluates to 0
(data?.read_timeout || data?.read_timeout === 0) ? data?.read_timeout : 60000
```

Fixes FTI-5437

## PR Checklist


kaiarrowood (approved):
LGTM! 🚀 

Very interesting, I had to read up on nullish operator again to understand how we were getting `true` from this. It looks like a valid solution would also have been to just remove the `=== 0`, but this works too.

image

Leopoldthecoder:
Ah yes, `data?.read_timeout ?? 60000` should be enough.

PR #848, opened 2023-10-19
feat(metric-cards): new v2.0 layout [MA-2139]

mihai-peteu:
# Summary

https://konghq.atlassian.net/browse/MA-2139

## PR Checklist


packages/analytics/analytics-metric-provider/package.json
```diff
@@ -50,7 +50,7 @@
   "dependencies": {
     "@kong-ui-public/analytics-utilities": "workspace:^0.8.4",
     "@kong-ui-public/core": "1.1.0",
-    "@kong-ui-public/metric-cards": "0.2.14",
+    "@kong-ui-public/metric-cards": "workspace:*",
```
mihai-peteu, line 53: @adamdehaven The goal here was to be able to update both `metric-cards` and `analytics-metric-provider` in a single PR, without first needing to publish a new version of `metric-cards`.

I ran `add metric-cards@latest` and ended up with the `workspace:*` diff. Noticed that all other workspace version instances in this repo are `workspace:^`.  Safe to leave as-is, or use the hat here also?

adamdehaven: @adorack this sounds familiar to something we chatted about a long time ago -- do you remember? 😬

mihai-peteu: Screenshot 2023-10-23 at 3 42 30 PM

adorack: Sorry, I didn't see your ping.  (Working on fixing on-call issues from this weekend, so I haven't had time to comb through Github notifications.)  I don't have any additional information, unfortunately.

packages/analytics/metric-cards/src/utilities/trend-display.ts
```diff
@@ -51,14 +53,14 @@ export const calculateChange = (curr: number, prev: number) => {
 /**
  * Determines whether to display an upward or downward trend, or no change
  */
-export const defineIcon = (polarity: number, thisIsBad: boolean = false): string => {
+export const defineIcon = (polarity: number, thisIsBad: boolean = false) => {
```
adamdehaven, line 56:
```suggestion
export const defineIcon = (polarity: number, thisIsBad: boolean = false): TrendUpIcon | TrendDownIcon | IndeterminateSmallIcon => {
```

packages/analytics/metric-cards/src/components/MetricCardContainer.vue
```diff
@@ -93,7 +97,7 @@ const formatCardValues = (card: MetricCardDef): MetricCardDisplayValue => {
     metricValue: card.formatValueFn ? card.formatValueFn(card.currentValue) : approxNum(card.currentValue, { capital: true, round: true }) || '0',
     metricChange: card.formatChangeFn ? card.formatChangeFn(change) : metricChange(change, props.hasTrendAccess, props.fallbackDisplayText),
     changePolarity: polarity,
-    icon: defineIcon(polarity, card.increaseIsBad),
+    trendIcon: defineIcon(polarity, card.increaseIsBad) as any,
```
adamdehaven, line 100: If you implement my other suggestion (providing a return type on the function) you should be able to get rid of the `any` here
```suggestion
    trendIcon: defineIcon(polarity, card.increaseIsBad),
```

packages/analytics/metric-cards/src/utilities/trend-display.ts
adamdehaven: You may have to do `typeof TrendUpIcon | typeof TrendDownIcon | typeof IndeterminateSmallIcon` since they are not types

mihai-peteu: Last commit did the trick. Thanks!

packages/analytics/metric-cards/src/components/MetricCardContainer.vue
```diff
@@ -47,11 +50,15 @@ import type { PropType } from 'vue'
 import { computed } from 'vue'
 import approxNum from 'approximate-number'
 import { MetricCardSize } from '../constants'
+import type { MetricCardType } from '../enums'
 import type { MetricCardDef, MetricCardDisplayValue } from '../types'
 import { changePolarity, metricChange, defineIcon, calculateChange } from '../utilities'
 import MetricsCard from './display/MetricsCard.vue'
 import MetricCardLoadingSkeleton from './display/MetricCardLoadingSkeleton.vue'
 
+// Only needed for type definitions
```
adamdehaven, line 59: suggestion: update comment
```suggestion
// Import any one of the `@kong/icons` components in order to access the interface (they are all the same); then alias as a generic to provide the icon interface to the prop types.
```

adorack (changes requested):
I'd like to review.

packages/analytics/analytics-metric-provider/src/components/MetricsConsumer.vue
```diff
@@ -39,13 +39,16 @@ const { traffic, latency } = providerData.data
...
+  cardType: MetricCardType.TRAFFIC,
   title: computed(() => providerData.longCardTitles
     ? i18n.t('metricCard.long.traffic')
     : i18n.t('metricCard.short.traffic')),
+  description: providerData.description,
   record: traffic.mapped,
   hasError: traffic.hasError,
   lookupKey: props.lookupKey,
   sumGroupedValues: ALL_STATUS_CODE_GROUPS,
+  trendRange: computed(() => providerData.trendRange.value),
```
adorack, line 51: This `computed` isn't doing anything -- if you'd like a ref, you should just be able to pass `providerData.trendRange`.

packages/analytics/analytics-metric-provider/src/composables/useMetricCardBuilder.ts
```diff
@@ -1,17 +1,20 @@
...
+  cardType: MetricCardType,
   title: Ref<string>,
+  description?: string,
   record: Ref<ChronologicalMappedMetrics>,
   hasError: Ref<boolean>,
   lookupKey?: string,
   sumGroupedValues?: string[],
   increaseIsBad?: boolean,
   formatValueFn?: (rawValue: number) => string,
+  trendRange?: Ref<string> | undefined,
```
adorack, line 17: This should just be `trendRange?: Ref<string>` -- the `?` implies the `| undefined`.

packages/analytics/analytics-utilities/package.json
```diff
@@ -60,6 +60,7 @@
     "date-fns-tz": "^2.0.0"
   },
   "devDependencies": {
+    "@kong-ui-public/i18n": "workspace:^",
```
adorack, line 63: This is more than a dev dependency, right?  I think this needs to be a `peerDependency` as well.  `analytics-metric-provider` has an example; if I recall correctly, you need to add it to `peerDependencies`, `devDependencies`, and make some changes to `vite.config`.

However, see other comments; you might not need i18n in `timeframes`.

packages/analytics/analytics-utilities/src/timeframes.ts
```diff
@@ -26,9 +28,13 @@ const adjustForTz = (d: Date, tz: string) => {
   return new Date(d.getTime() - getTimezoneOffset(tz, d))
 }
 
+const { i18n } = composables.useI18n()
+
 export class Timeframe {
   readonly timeframeText: string
 
+  readonly rangeDisplayText: string
```
adorack, line 36: One thought...

All timeframes have unique keys.  Instead of adding i18n to `timeframes` (which might cause some issues down the line, since this library is used server-side), maybe the metric card provider can do something like

```
i18n.t(`timeframe.rangeDisplay.${timeframe.key}`)
```

with appropriate adjustments for `custom` timeframes.

Sorry that I didn't think of this earlier when you were talking about it; I hadn't dug into this yet.

packages/analytics/metric-cards/src/components/MetricCardContainer.vue
```diff
@@ -33,10 +33,13 @@
         :key="index"
         v-bind="formatCardValues(card)"
         :card-size="cardSize"
+        :card-type="(card.cardType as MetricCardType)"
```
adorack, line 36: Out of curiosity, is this cast still needed?

packages/analytics/metric-cards/src/components/display/MetricsCard.vue
```diff
@@ -55,15 +97,48 @@
 
 <script setup lang="ts">
 import type { PropType } from 'vue'
+import { computed } from 'vue'
+import {
+  KUI_COLOR_BORDER_DANGER_STRONG,
+  // KUI_COLOR_BACKGROUND_NEUTRAL,
```
adorack, line 103: Please remove dead code.

packages/analytics/metric-cards/src/enums/metric-card-size.enum.ts
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable no-unused-vars */
```
adorack, line 1: Why is this enum here when you already have `MetricCardSize` in `constants.ts`?

packages/analytics/metric-cards/src/types/metric-card.ts
```diff
@@ -1,21 +1,25 @@
...
+  trendRange?: string
 }
 
 export interface MetricCardDisplayValue {
   metricValue: string
   metricChange: string
   changePolarity: number
-  icon?: string
-  cardSize?: string
+  trendIcon?: any
```
adorack, line 21: Is it possible to type this as `GenericIcon`?  No worries if not, just checking.

packages/analytics/analytics-metric-provider/README.md
```diff
@@ -118,3 +118,4 @@ The provider and consumer support use in contexts where metric cards are not app
 - `additionalFilter`: commonly used to apply additional filters to the metric query: for example, to retrieve many entities within a given scope.
 - `queryReady`: if this property is set and is `false`, the metric cards will not issue a query -- they will remain in a "loading" state.  When this property becomes `true`, the query will fire.  Useful for when the component renders before the page has all of the necessary information to construct a query.
 - `longCardTitles`: whether to show the long or short translation for the title of the card.
+- `trendRange`: a description of the time period involved in the calculation, determined internally by the MetricsProvider
```
adorack, line 121: I think this isn't actually an "other property" -- it's not exposed to the consumer of this component.

packages/analytics/analytics-metric-provider/src/components/MetricsProviderInternal.vue
```diff
@@ -79,8 +97,10 @@ provide(METRICS_PROVIDER_KEY, {
     traffic: trafficData,
     latency: latencyData,
   },
+  description: props.description || '',
```
adorack, line 100: If you need description to be a string, I think you can just set a default on line 39.  But you might not need to have this fallback, since `description` is typed as optional in the provider data interface?  I'd say either set a default and make it required in the provider data, or don't say `|| ''` and leave the types as-is.

adorack (changes requested):

packages/analytics/analytics-utilities/package.json
mihai-peteu: ty for the heads up on that, fixing.

mihai-peteu: Installed with `add --save-peer`. That makes it pop up in both devDependencies` and `peerDependencies` - which seems to be the expected behaviour, judging by other instances in this repo 👍

packages/analytics/analytics-metric-provider/src/components/MetricsProviderInternal.vue
mihai-peteu: You speak the truth :) This was added in a frenzy of attempts to pass the doggone typecheck a few days ago.

Runs cleanly without, removing the fallback.

packages/analytics/analytics-metric-provider/src/components/MetricsConsumer.vue
mihai-peteu: Well, i'll be damned... Thanks for the heads up.

packages/analytics/metric-cards/src/enums/metric-card-size.enum.ts
mihai-peteu: Unused file, i'll remove.

Context: I attempted to move the MetricCardSize to an `enum` earlier in the PR, and that caused the typecheck to explode. So moved it back to `constants.ts`.

filipgutica (changes requested):
Just a couple high level questions. 

Any updates to `analytics-metric-provider` README file to reflect any of the new changes?

Are there any breaking changes for any of the packages, be it `metric-card` or `analytics-metric-provider` that require the breaking changes flag to be set?

packages/analytics/metric-cards/src/components/MetricCardContainer.vue
mihai-peteu: Yup 😬 

Screenshot 2023-10-27 at 8 57 35 AM

mihai-peteu:
@filipgutica I don't think there are breaking changes - I tested the new metric-provider temp package on `konnect-portal` and nothing blew up in my face:

Screenshot 2023-10-27 at 9 06 32 AM

Screenshot 2023-10-27 at 9 06 59 AM

adorack (approved):

packages/analytics/metric-cards/package.json
```diff
@@ -43,6 +43,7 @@
   },
   "devDependencies": {
     "@kong/design-tokens": "^1.11.6",
+    "@kong/icons": "^1.7.8",
```
adorack, line 46: Just to double-check... I would have guessed this would be a `peerDependency` -- why is it a dev dependency?

packages/analytics/metric-cards/src/components/MetricCardContainer.vue
adorack: This means you're coercing a possibly undefined value into a prop that expects it to be defined.  Somewhere, you need to have a default value.

adorack (changes requested):

packages/analytics/metric-cards/package.json
mihai-peteu: Good question. I just did what @adamdehaven told me to do, which is install it as a devDependency :)

Now that I look at the rest of `public-ui-components`, `@kong-icons` is used under `dependencies` in 2 other places:

https://github.com/Kong/public-ui-components/blob/main/packages/core/error-boundary/package.json#L65
https://github.com/Kong/public-ui-components/blob/main/packages/entities/entities-shared/package.json#L72

adamdehaven: you need it as an actual dependency since you're using the rendered SVG; technically it _could_ be a peerDep, but the package is so light it doesn't really matter TBH

packages/analytics/metric-cards/src/components/MetricCardContainer.vue
mihai-peteu: I'll default it to the Generic Count card type then

adorack (approved):

adamdehaven (approved):
✅ based on @adorack's signoff 
The code lgtm

PR #852, opened 2023-10-19
fix(entities-shared): use ktooltip for truncated item [MA-2088]

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-2088

  - We want to use KTooltip for the truncated text tooltip. However, KTooltips with an empty label will still render a tooltip with an empty box. So we need to conditionally render KTooltip.

image

## PR Checklist


packages/entities/entities-shared/src/components/entity-base-config-card/ConfigCardItem.vue
```diff
@@ -103,14 +103,26 @@
...
-            <span
-              v-else
-              class="attrs-data-text"
-              :class="{ 'truncated': truncated }"
-              :title="truncated ? item.value : undefined"
-            >
-              {{ componentAttrsData.text }}
-            </span>
+            <div v-else>
+              <KTooltip
```
adamdehaven, line 107: TBH, I think the better solution here would be to utilize a dynamic component or something that would/would not render the tooltip instead of injecting elements into the DOM.

Alternatively, we could also alter KTooltip to not render something if the `label` is `undefined`

filipgutica: I updated it to not render things in the DOM, I figured out the break point where the label starts being truncated

packages/entities/entities-shared/src/components/entity-base-config-card/ConfigCardItem.vue
```diff
@@ -131,6 +143,9 @@ import InternalLinkItem from './InternalLinkItem.vue'
 import StatusBadge from './StatusBadge.vue'
 import '@kong-ui-public/copy-uuid/dist/style.css'
 
+// 20ch + ellipsis(3ch) + 1
+const MAX_LABEL_LENGTH = 24
```
Leopoldthecoder, line 147: This may not be accurate. Because we are not using a monospaced font, different characters will have very different widths, e.g. 'l' and 'W'. A better way would be to render a `<span>` with the same style off-screen and check its `clientWidth` and `scrollWidth`.

That being said, I can accept this as a temporary solution.

adamdehaven: Would it be helpful to modify KTooltip so that it does not render the popup if the `label` is empty?

Leopoldthecoder: > Would it be helpful to modify KTooltip so that it does not render the popup if the label is undefined?
Yes, it will be helpful.

If KTooltip does not render the popup when `label` is undefined, we still need to know when the text is being truncated.

kaiarrowood: Big +1 that it would be nice to not render the tooltip if `label` is `undefined`. We end up writing conditional code to support this scenario A LOT.

adamdehaven: Updated
https://github.com/Kong/kongponents/pull/1783

Leopoldthecoder (dismissed):
LGTM🚀

packages/entities/entities-shared/src/composables/useTruncationDetector.ts
```diff
@@ -0,0 +1,26 @@
...
+    if (textContent.value && !truncationCalculated.value) {
+      textOffsetWidth.value = (textContent.value as HTMLElement).offsetWidth
+      textScrollWidth.value = (textContent.value as HTMLElement).scrollWidth
+      truncationCalculated.value = true
+    }
+  }
+
+  watch(textContent, detectTruncation)
+
+  const isTruncated = computed(() => {
```
adamdehaven, line 19:
```suggestion
  const isTruncated = computed((): boolean => {
```

packages/entities/entities-shared/src/components/entity-base-config-card/ConfigCardItem.vue
```diff
@@ -103,14 +103,17 @@
...
 
-            <span
-              v-else
-              class="attrs-data-text"
-              :class="{ 'truncated': truncated }"
-              :title="truncated ? item.value : undefined"
-            >
-              {{ componentAttrsData.text }}
-            </span>
+            <div v-else>
```
adamdehaven, line 106: question: why can’t the v-else go directly on the KTooltip?

packages/entities/entities-shared/src/components/entity-base-config-card/ConfigCardItem.vue
```diff
@@ -293,6 +296,11 @@ const componentAttrsData = computed((): ComponentAttrsData => {
       }
   }
 })
+
+const textContent = ref<HTMLElement>()
+
+const { isTruncated } = composables.useTruncationDetector(textContent as Ref<HTMLElement>)
```
adamdehaven, line 302: question: @Leopoldthecoder does it seem odd to pass in a `ref` to the function instead of the underlying value?

packages/entities/entities-shared/src/index.ts
```diff
@@ -12,13 +12,13 @@ import EntityFormSection from './components/entity-form-section/EntityFormSectio
...
 // Extract specific composables to export
-const { useAxios, useDeleteUrlBuilder, useErrors, useFetchUrlBuilder, useFetcher, useDebouncedFilter, useStringHelpers, useHelpers, useGatewayFeatureSupported } = composables
+const { useAxios, useDeleteUrlBuilder, useErrors, useFetchUrlBuilder, useFetcher, useDebouncedFilter, useStringHelpers, useHelpers, useGatewayFeatureSupported, useTruncationDetector } = composables
 
 // Components
 export { EntityBaseConfigCard, ConfigCardItem, ConfigCardDisplay, InternalLinkItem, EntityBaseForm, EntityBaseTable, EntityDeleteModal, EntityFilter, EntityToggleModal, PermissionsWrapper, EntityFormSection }
 
 // Composables
-export { useAxios, useDeleteUrlBuilder, useErrors, useFetchUrlBuilder, useFetcher, useDebouncedFilter, useStringHelpers, useHelpers, useGatewayFeatureSupported }
+export { useAxios, useDeleteUrlBuilder, useErrors, useFetchUrlBuilder, useFetcher, useDebouncedFilter, useStringHelpers, useHelpers, useGatewayFeatureSupported, useTruncationDetector }
```
adamdehaven, line 21: question: since the truncation composable is only used in this package, why does it need to be exported?

packages/entities/entities-shared/src/components/entity-base-config-card/ConfigCardItem.vue
Leopoldthecoder: I think it's acceptable. We have other composables (e.g. `useFetcher`) that take a `MaybeRef` param. Besides in the `useTruncationDetector` function, we need to watch this `textContent` param.

adamdehaven: Wondering if that same maybe logic should be used here for safety

Leopoldthecoder: Ah yeah I think it'd be nice. More unified with other composables.

packages/entities/entities-shared/src/index.ts
filipgutica: I'm using it in `khcp-ui`

adamdehaven: Should that be a helper composable exported from https://github.com/Kong/public-ui-components/tree/main/packages/core/core instead?

adamdehaven: Meh nevermind; keep as-is 🤷🏼

packages/entities/entities-shared/src/components/entity-base-config-card/ConfigCardItem.vue
adamdehaven: @filipgutica can you add the logic to handle if a user passes in a `ref` or just the HTML element? The composable should be able to handle either

filipgutica: but if I watch the `unRefed` value does that break the reactivity?

filipgutica: done

adamdehaven: Yes. If you explicitly want a `Ref` I would add block comments above the `export default function useTruncationDetector` line so that it's shown during code hinting; something like:

```ts
/**
 * Detect when blah blah blah
 * @param Ref<HTMLElement> textContent The description of this param... 
 * @returns ... the thing
 */
export default function useTruncationDetector(textContent: Ref<HTMLElement>) {
```

filipgutica: 👍

packages/entities/entities-shared/src/composables/useTruncationDetector.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  const detectTruncation = () => {
+    if (textContent.value && !truncationCalculated.value) {
+      textOffsetWidth.value = (textContent.value as HTMLElement).offsetWidth
+      textScrollWidth.value = (textContent.value as HTMLElement).scrollWidth
+      truncationCalculated.value = true
+    }
+  }
+
+  watch(textContent, detectTruncation)
```
adamdehaven, line 24: Since you're using a watcher here, it would be better to utilize the data initialized internally:
```suggestion
  const detectTruncation = () => {
    if (textContent.value && !truncationCalculated.value) {
      textOffsetWidth.value = (textContent.value as HTMLElement).offsetWidth
      textScrollWidth.value = (textContent.value as HTMLElement).scrollWidth
      truncationCalculated.value = true
    }
  }

  watch(textContent, (content: HTMLElement) => {
    if (content && !truncationCalculated.value) {
      textOffsetWidth.value = content.offsetWidth
      textScrollWidth.value = content.scrollWidth
      truncationCalculated.value = true
    }
  })
```

packages/entities/entities-shared/src/composables/useTruncationDetector.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  const isTruncated = computed<boolean>(() => {
+    return textOffsetWidth.value < textScrollWidth.value
+  })
```
adamdehaven, line 28: question: this could return `false` before `truncationCalculated` has been set to `true` -- is that an issue?

adamdehaven (approved):

PR #870, opened 2023-10-26
feat(plugins): plugin select unification [khcp-9487]

kaiarrowood:
# Summary

Unification for the `Plugins` -> `Select` page. For [KHCP-9487](https://konghq.atlassian.net/browse/KHCP-9512).

## PR Checklist

[KHCP-9487]: https://konghq.atlassian.net/browse/KHCP-9487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/entities/entities-plugins/src/plugins-endpoints.ts
```diff
@@ -9,14 +9,26 @@ export default {
       forEntity: '/{workspace}/{entityType}/{entityId}/plugins',
     },
   },
+  select: {
+    konnect: {
+      availablePlugins: '/api/runtime_groups/{controlPlaneId}/v1/available-plugins',
+    },
+    kongManager: {
+      availablePlugins: '/{workspace}/kong',
```
Leopoldthecoder, line 17: Non-blocking: just realized if we want to adopt `PluginSelect` in the [open-source Kong Manager](https://github.com/Kong/kong-manager), we'll need to update this block because Kong Gateway CE has a different endpoint to return available plugins. But don't worry, the KM team can handle this when we start to uptake `PluginSelect`.

packages/entities/entities-plugins/src/components/PluginForm.vue
```diff
@@ -0,0 +1,236 @@
+<template>
+  <div class="kong-ui-entities-plugin-form">
+    <EntityBaseForm
```
Leopoldthecoder, line 3: Question: is this just a placeholder or are we using `EntityBaseForm` to build plugin forms?

packages/entities/entities-plugins/docs/plugin-select.md
```diff
@@ -0,0 +1,190 @@
...
+- default: `{}`
+
+Plugins that should be disabled and their disabled messages.
+Example:
+
+```json
+{
+  'acl': 'ACL is not supported for this entity type',
+}
+```
```
Leopoldthecoder, line 157: I tried adding `:disabled-plugins="{ acl: 'ACL is not supported for this entity type' }"` in the sandbox, but clicking the `acl` card still takes me to the plugin form page:

Kapture 2023-10-30 at 17 47 21

packages/entities/entities-plugins/src/components/DeleteCustomPluginSchemaModal.vue
```diff
@@ -0,0 +1,205 @@
...
+      :title="t('delete.title')"
+      :visible="isDeleteModalVisible"
+      @cancel="hideDeleteModal"
+      @proceed="confirmDelete"
+    /> -->
+          <!-- TODO: just use KPrompt??? -->
+          <!-- TODO: i18n-t element instead
+            <p class="confirm-text">
+            {{ helpText.confirm_1 }} "<strong>{{ plugin?.name }}</strong>" {{ helpText.confirm_2 }}.
+          </p> -->
```
adamdehaven, line 60: question: is this commented component code intentional?

packages/entities/entities-plugins/src/components/DeleteCustomPluginSchemaModal.vue
```diff
@@ -0,0 +1,205 @@
...
+        >
+          {{ t('actions.go_to_plugins') }}
+        </KButton>
+        <KButton
+          v-else
+          appearance="danger"
+          data-testid="delete-custom-plugin-form-submit"
+          :disabled="isDeleteButtonDisabled"
+          form="delete-custom-plugin-form"
+          :icon="isLoading ? 'spinner' : undefined"
```
adamdehaven, line 99: suggestion: instead of using the `icon` prop that is going away, can you just slot in the `ProgressIcon` from `@kong/icons`? Same comment if you're using any other `icon` props in the PR.

packages/entities/entities-plugins/src/components/DeleteCustomPluginSchemaModal.vue
```diff
@@ -0,0 +1,205 @@
...
+})
+
+const isLoading = ref(false)
+const errorMessage = ref('')
+const customPluginNameFormValue = ref('')
+const isPluginSchemaInUse = ref(false)
+
+const handleCancel = (): void => {
+  emit('closed')
+}
```
adamdehaven, line 139: suggestion: instead of creating a function here for a one-time-use emit, can you just put the emit inline in the template?

packages/entities/entities-plugins/src/components/DeleteCustomPluginSchemaModal.vue
```diff
@@ -0,0 +1,205 @@
...
+
+const handleSubmit = async (): Promise<void> => {
+  isLoading.value = true
+  errorMessage.value = ''
+
+  if (!props.plugin?.id) {
+    return
+  }
```
adamdehaven, line 147: question: here you are setting loading to true but then exiting if there is no plugin id. Should you exit before modifying the loading state?

packages/entities/entities-plugins/src/components/PluginCardSkeleton.vue
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+
+  // TODO: do I need these two? They would be deep()
+  .skeleton-card-column {
+    width: calc(33% - $kui-space-60);
+  }
+  .skeleton-card {
+    height: 240px;
+  }
```
adamdehaven, line 60: not sure I understand the question here

packages/entities/entities-plugins/src/components/PluginCustomGrid.vue
```diff
@@ -0,0 +1,274 @@
...
+        v-if="modifiedCustomPlugins.length <= pluginsPerRow"
+        #trigger
+      >
+        &nbsp;
+      </template>
+      <template #visible-content>
+        <div class="plugin-card-container">
+          <PluginSelectCard
+            v-for="(plugin, index) in getPluginCards('visible')"
+            :key="index"
```
adamdehaven, line 43: nit: indexes by themselves don't make good keys; can you use a plugin id/name or something instead?

Same comment for the other v-for below

packages/entities/entities-plugins/src/components/PluginCustomGrid.vue
```diff
@@ -0,0 +1,274 @@
...
+            :can-delete-custom="canDeleteCustom"
+            :can-edit-custom="canEditCustom"
+            :config="config"
+            :no-route-change="noRouteChange"
+            :plugin="plugin"
+            @custom-plugin-delete="() => handleCustomPluginDelete(plugin)"
```
adamdehaven, line 49: question: why do you `() => callbackFn` instead of just `@custom-plugin-delete="handleCustomPluginDelete(plugin)"`

packages/entities/entities-plugins/src/components/PluginCustomGrid.vue
```diff
@@ -0,0 +1,274 @@
...
+          @plugin-clicked="emitPluginData"
+        />
+      </div>
+    </KCollapse>
+
+    <DeleteCustomPluginSchemaModal
+      v-if="openDeleteModal && selectedPlugin"
+      :plugin="selectedPlugin"
+      @closed="handleClose"
+      @proceed="() => handleClose(true)"
```
adamdehaven, line 73: question: why not just `@proceed="handleClose(true)"`

packages/entities/entities-plugins/src/components/PluginCustomGrid.vue
```diff
@@ -0,0 +1,274 @@
...
+   */
+  pluginsPerRow: {
+    type: Number,
+    default: 4,
+  },
+})
+
+const emit = defineEmits<{
+  (e: 'plugin-clicked', plugin: PluginType): void,
+  (e: 'revalidate'): void,
```
adamdehaven, line 149: question: what is this `revalidate` event used for?

Also, I don't see it in the docs

packages/entities/entities-plugins/src/components/PluginCustomGrid.vue
```diff
@@ -0,0 +1,274 @@
...
+
+  // ADD CUSTOM_PLUGIN_CREATE as the first card if allowed creation
+  return props.canCreateCustom && !props.noRouteChange && props.config.createCustomRoute
+    ? [{
+      id: 'custom-plugin-create',
+      name: t('plugins.select.tabs.custom.create.name'),
+      available: true,
+      group: PluginGroup.CUSTOM_PLUGINS,
+      description: t('plugins.select.tabs.custom.create.description'),
+    } as PluginType].concat(customPlugins)
```
adamdehaven, line 174: question: why do you have to force `as PluginType` here? Why doesn't this already follow the interface?

packages/entities/entities-plugins/src/components/PluginCustomGrid.vue
```diff
@@ -0,0 +1,274 @@
...
+    return modifiedCustomPlugins.value.slice(0, props.pluginsPerRow)
+  }
+
+  return modifiedCustomPlugins.value.slice(props.pluginsPerRow)
+}
+
+// text for plugin group "view x more" label
+const triggerLabel = computed((): string => {
+  const totalCount = getPluginCards('all')?.length || 0
+  const hiddenCount = getPluginCards('hidden')?.length || 0
```
adamdehaven, line 191: question: it looks like `getPluginCards` always returns an array, right?

If yes, I'm not sure you need the fallbacks to `0` here?

packages/entities/entities-plugins/src/components/PluginCustomGrid.vue
```diff
@@ -0,0 +1,274 @@
...
+  openDeleteModal.value = true
+  selectedPlugin.value = {
+    id: plugin.id,
+    name: plugin.name,
+  }
+}
+
+const handleClose = (revalidate?: boolean): void => {
+  if (revalidate) {
+    emit('revalidate')
```
adamdehaven, line 213: question: why do we emit this?

packages/entities/entities-plugins/src/components/PluginCustomGrid.vue
```diff
@@ -0,0 +1,274 @@
...
+      grid-template-columns: repeat(2, 1fr);
+    }
+
+    @media (min-width: $kui-breakpoint-tablet) {
+      grid-template-columns: repeat(3, 1fr);
+    }
+
+    @media (min-width: $kui-breakpoint-laptop) {
+      grid-template-columns: repeat(4, 1fr);
+    }
```
adamdehaven, line 271: mobile first 🎉

packages/entities/entities-plugins/src/components/PluginForm.vue
```diff
@@ -0,0 +1,236 @@
...
+  name: '',
+  tags: '',
+  entity_id: '',
+})
+
+/**
+ * Is the form submit button enabled?
+ * If the form.fields and formFieldsOriginal are equal, always return false
+ */
+const canSubmit = computed((): boolean => JSON.stringify(form.fields) !== JSON.stringify(formFieldsOriginal) && !!form.fields.name)
```
adamdehaven, line 132: suggestion: could you use your `objectsAreEqual` helper here?

packages/entities/entities-plugins/src/components/PluginForm.vue
```diff
@@ -0,0 +1,236 @@
...
+
+const initForm = (data: Record<string, any>): void => {
+  form.fields.name = data?.name || ''
+  form.fields.tags = data?.tags?.join(', ') || ''
+  // TODO:
+  // form.fields.entity_id = data?.certificate?.id || ''
+
+  // Set initial state of `formFieldsOriginal` to these values in order to detect changes
+  Object.assign(formFieldsOriginal, form.fields)
```
adamdehaven, line 141: You may want to check that things remain reactive and are correctly mapped here -- probably fine, but double-check

packages/entities/entities-plugins/src/components/PluginSelect.vue
```diff
@@ -0,0 +1,497 @@
...
+            <PluginCustomGrid
+              :can-create-custom="userCanCreateCustom"
+              :can-delete-custom="userCanDeleteCustom"
+              :can-edit-custom="userCanEditCustom"
+              :config="config"
+              :no-route-change="noRouteChange"
+              :plugin-list="filteredPlugins"
+              :plugins-per-row="pluginsPerRow"
+              @plugin-clicked="(val: PluginType) => $emit('plugin-clicked', val)"
+              @revalidate="() => pluginsList = buildPluginList()"
```
adamdehaven, line 95: question: why not just `@revalidate="pluginsList = buildPluginList()"`

packages/entities/entities-plugins/src/components/PluginSelect.vue
```diff
@@ -0,0 +1,497 @@
...
+})
+
+const tabs = props.config.app === 'konnect'
+  ? [
+    {
+      hash: '#kong',
+      title: t('plugins.select.tabs.kong.title'),
+    },
+    {
+      hash: '#custom',
```
adamdehaven, line 264: question: should these be i18n strings that get sanitized to alpha and `-` only?

packages/entities/entities-plugins/src/components/PluginSelect.vue
```diff
@@ -0,0 +1,497 @@
...
+watch(() => props.ignoredPlugins, (val, oldVal) => {
+  if (!objectsAreEqual(val, oldVal) && !isLoading.value) {
+    pluginsList.value = buildPluginList()
+  }
+})
+
+const userCanCreateCustom = ref(false)
+const userCanEditCustom = ref(false)
+const userCanDeleteCustom = ref(false)
+onBeforeMount(async () => {
```
adamdehaven, line 411: nit: spacing
```suggestion
const userCanDeleteCustom = ref(false)

onBeforeMount(async () => {
```

packages/entities/entities-plugins/src/components/PluginSelect.vue
```diff
@@ -0,0 +1,497 @@
...
+  // Evaluate the user permissions
+  userCanCreateCustom.value = await props.canCreateCustom()
+  userCanEditCustom.value = await props.canEditCustom()
+  userCanDeleteCustom.value = await props.canDeleteCustom()
+})
+
+onMounted(async () => {
+  isLoading.value = true
```
adamdehaven, line 419: suggestion: if we're setting `isLoading` to true immediately, would it be easier just to init the ref as `true` where it is defined?

packages/entities/entities-plugins/src/components/PluginSelect.vue
```diff
@@ -0,0 +1,497 @@
...
+  hasError.value = false
+  fetchErrorMessage.value = ''
```
adamdehaven, line 421: If `hasError` and `fetchErrorMessage` were initialized in these states, there's no reason to need to set them again at the root of the `onMounted` hook

packages/entities/entities-plugins/src/components/PluginSelect.vue
```diff
@@ -0,0 +1,497 @@
...
+
+  emit('loading', isLoading.value)
```
adamdehaven, line 423: suggestion: instead of calling the emit every time you flip the state, you could just throw a watcher on it:

```ts
watch(isLoading), (loading: boolean) => {
  emit('loading', loading)
})
```

packages/entities/entities-plugins/src/components/PluginSelect.vue
```diff
@@ -0,0 +1,497 @@
...
+
+  try {
+    const res = await axiosInstance.get(availablePluginsUrl.value)
```
adamdehaven, line 426: suggestion: since you're only accessing data, go ahead and destructure
```suggestion
    const { data } = await axiosInstance.get(availablePluginsUrl.value)
```

packages/entities/entities-plugins/src/components/PluginSelect.vue
```diff
@@ -0,0 +1,497 @@
...
+  } catch (error: any) {
+    hasError.value = true
+    fetchErrorMessage.value = getMessageFromError(error)
+  }
+
+  // fetch scoped entity to check for pre-existing plugins
+  if (fetchEntityPluginsUrl.value) {
+    try {
+      const res = await axiosInstance.get(fetchEntityPluginsUrl.value)
+      const { data } = res.data
```
adamdehaven, line 445: suggestion: since you're only accessing data, go ahead and destructure
```suggestion
      const { data } = await axiosInstance.get(fetchEntityPluginsUrl.value)
```

You may have to rename this one though if you do the suggestion above as well, e.g.:

```ts
const { data: myOtherDataVarName } = await axiosInstance.get(fetchEntityPluginsUrl.value)
```

packages/entities/entities-plugins/src/components/PluginSelect.vue
```diff
@@ -0,0 +1,497 @@
...
+
+      if (data?.length) {
```
adamdehaven, line 447: If you destructure `data` above you won't need the optional chaining here

packages/entities/entities-plugins/src/components/PluginSelectCard.vue
```diff
@@ -0,0 +1,310 @@
...
+    }"
+    :label="plugin.disabledMessage"
+    position-fixed
+  >
+    <KCard
+      class="plugin-card-content"
+      :data-testid="plugin.name"
+      :disabled="!plugin.available || plugin.disabledMessage"
+      has-hover
+      @click="isCustomPlugin ? undefined : noRouteChange ? emitPluginData() : handleCreateClick()"
```
adamdehaven, line 15: suggestion: instead of the conditional logic inline here, can you map this to a function and then call whatever is appropriate conditionally there instead?

packages/entities/entities-plugins/src/components/PluginSelectCard.vue
```diff
@@ -0,0 +1,310 @@
...
+      >
+        <div class="header-wrapper">
+          <KBadge v-if="!isCreateCustomPlugin">
+            {{ t('plugins.select.custom_badge_text') }}
+          </KBadge>
+        </div>
+      </template>
+
+      <template
+        v-if="isCustomPlugin && !isCreateCustomPlugin && !noRouteChange && controlPlaneId && (canDeleteCustom || canEditCustom)"
```
adamdehaven, line 29: suggestion: this is a lot of logic; can we move to `computed`?

packages/entities/entities-plugins/src/components/PluginSelectCard.vue
```diff
@@ -0,0 +1,310 @@
...
+              class="actions-trigger"
+              data-testid="overflow-actions-button"
+              size="small"
+            >
+              <template #icon>
+                <KIcon
+                  :color="KUI_COLOR_TEXT_NEUTRAL_STRONGER"
+                  icon="more"
+                  :size="KUI_ICON_SIZE_30"
+                />
```
adamdehaven, line 47: suggestion: can you import from `@kong/icons` instead of using `KIcon`? Applies to the whole PR whenever you have a `KIcon` implemented

packages/entities/entities-plugins/src/components/PluginSelectCard.vue
```diff
@@ -0,0 +1,310 @@
...
+import { computed, type PropType } from 'vue'
+import { useRouter } from 'vue-router'
+import { PluginGroup, type KongManagerPluginFormConfig, type KonnectPluginFormConfig, type PluginType } from '../types'
+import { KUI_ICON_SIZE_30, KUI_COLOR_TEXT_NEUTRAL_STRONGER } from '@kong/design-tokens'
+import composables from '../composables'
+import PluginIcon from './PluginIcon.vue'
+
+const emit = defineEmits<{
+  (e: 'plugin-clicked', plugin: PluginType) : void,
+  (e: 'custom-plugin-delete'): void,
```
adamdehaven, line 120: I didn't see this event documented anywhere (sorry if I missed it)?

packages/entities/entities-plugins/src/components/PluginSelectCard.vue
```diff
@@ -0,0 +1,310 @@
...
+}
+</script>
+
+<style lang="scss" scoped>
+.plugin-card {
+  color: initial;
+  display: flex;
+  flex-basis: 100%;
+  flex-flow: row-wrap;
+  max-width: 335px;
```
adamdehaven, line 207: question: why this specific max-width?

packages/entities/entities-plugins/src/components/PluginSelectGrid.vue
```diff
@@ -0,0 +1,217 @@
...
+
+          <div class="plugin-card-container">
+            <PluginSelectCard
+              v-for="(plugin, index) in getPluginCards(group, 'hidden')"
+              :key="index"
+              :config="config"
+              :no-route-change="noRouteChange"
+              :plugin="plugin"
+              @plugin-clicked="emitPluginData"
+            />
```
adamdehaven, line 64: same comments here for `key` using `index`

packages/entities/entities-plugins/src/components/PluginSelectGrid.vue
```diff
@@ -0,0 +1,217 @@
...
+  const plugins = nonCustomPlugins.value[group as keyof PluginCardList] || []
+
+  if (type === 'all') {
+    return plugins
+  } else if (type === 'visible') {
+    return plugins.slice(0, props.pluginsPerRow)
+  }
+
+  return plugins.slice(props.pluginsPerRow)
+}
```
adamdehaven, line 152: This function looks identical to the `getPluginCards` implementation in the other file; could it be extracted to a composable in this package and shared?

packages/entities/entities-shared/src/composables/useHelpers.ts
```diff
@@ -10,7 +10,47 @@ export default function useHelpers() {
     return slotProps?.[propName] ?? undefined
   }
 
+  /**
+ * Check if 2 objects are equal
+ * @param {Object} a first object to compare
+ * @param {Object} b second object to compare
+ * @returns {Boolean} whether or not the objects are equal
+ */
```
adamdehaven, line 18: nit: formatting

packages/entities/entities-shared/src/composables/useHelpers.ts
```diff
@@ -10,7 +10,47 @@ export default function useHelpers() {
...
+    }
+  }
+
+  /**
+ * A comparator function that given a key, compares object values with that key, and returns the results of
+ * localCompare on those values (see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare for reference)
+ * Also checks for undefined, nulls and sub-Arrays.
+ * @param {String} property the key to sort on
+ * @returns {Function} a comparator function
+ */
```
adamdehaven, line 33: nit: formatting

packages/entities/entities-plugins/docs/plugin-select.md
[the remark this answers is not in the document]
kaiarrowood: Because ignored plugins are completely not shown. Disabled plugins are shown but we must display a tooltip explaining why it's disabled.

packages/entities/entities-plugins/src/components/DeleteCustomPluginSchemaModal.vue
```diff
@@ -0,0 +1,205 @@
...
+            appearance="danger"
+          />
+        </div>
+        <form
+          id="delete-custom-plugin-form"
+          @submit.prevent="handleSubmit"
+        >
+          <p>
+            {{ t('delete.confirmModalText1') }}
+            <strong>{{ plugin?.name }}</strong>?
```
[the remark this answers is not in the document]
kaiarrowood: This file wasn't ready when I asked you to review, sorry about that, I had missed refactoring it. Cleaned all of this up

packages/entities/entities-plugins/src/components/DeleteCustomPluginSchemaModal.vue
kaiarrowood: See above

packages/entities/entities-plugins/src/components/DeleteCustomPluginSchemaModal.vue
kaiarrowood: See above

packages/entities/entities-plugins/src/components/PluginCardSkeleton.vue
kaiarrowood: Note to myself from before I did a pass to clean up TODO's, removed now

packages/entities/entities-plugins/src/components/PluginCustomGrid.vue
kaiarrowood: Because passing it that way evaluates it immediately instead of when the event fires - nevermind, I remember now we researched this and that is a React rule, not a Vue rule. I'll switch.

packages/entities/entities-plugins/src/components/PluginCustomGrid.vue
kaiarrowood: It's internal only (used in `PluginSelect.vue`), we currently aren't doc'ing this component because it isn't exposed for consumption. You have the ability to delete custom plugins from the select page which means we need to refresh the plugins list afterwards.

packages/entities/entities-plugins/src/components/PluginCustomGrid.vue
kaiarrowood: It does follow the interface, but when I didn't cast it TS was saying I was trying to append an untyped object to a typed object array.

packages/entities/entities-plugins/src/components/PluginCustomGrid.vue
kaiarrowood: Emitted to `PluginSelect` to refresh custom plugin list after deleting a custom plugin.

packages/entities/entities-plugins/src/components/PluginSelect.vue
kaiarrowood: I'm not sure I follow, these are the actual tab hashes?

packages/entities/entities-plugins/src/components/PluginSelect.vue
kaiarrowood: This one is `data.data`, wouldn't I still need to check that it's defined and has a non-zero length?

packages/entities/entities-plugins/src/components/PluginSelectCard.vue
kaiarrowood: This is a child component that we don't expose to consuming apps. The emit is only used by `CustomPluginGrid` which is also not exposed to consuming apps.

packages/entities/entities-plugins/docs/plugin-select.md
```diff
@@ -0,0 +1,190 @@
...
+
+#### `pluginsPerRow`
+
+- type: `number`
+- required: `false`
+- default: `4`
+
+Number of plugins to always have visible (never will be collapsed).
```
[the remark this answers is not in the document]
kaiarrowood: I think we can't do that. We can't use standard wrapping for this because what this actually is, is the number of plugins in the "always visible" section of the KCollapse and I don't think the logic to determine that on the fly would be worth it, it isn't the same thing as the number of plugins to display in a row before wrapping. We would end up needing a staging area like we have for KMultiselect to determine how many plugins we can fit in a single row and then have to do this for each plugin section.

packages/entities/entities-plugins/src/components/PluginCustomGrid.vue
adamdehaven: Can you just add a note that the revalidate event is internal-only?

packages/entities/entities-plugins/src/components/PluginCustomGrid.vue
adamdehaven: suggestion: Can you add that has a note here, and that it's internal-only

packages/entities/entities-plugins/src/components/PluginSelect.vue
adamdehaven: Ignore this; I was going down a road when I reviewed but the implementation ended up being diff.

packages/entities/entities-plugins/src/components/PluginSelect.vue
adamdehaven: If it's `data.data` then you can do this:

```ts
const { data: { data } } = await axiosInstance.get(fetchEntityPluginsUrl.value)
```

packages/entities/entities-plugins/src/components/PluginSelectCard.vue
adamdehaven: Same suggestion here; can you just add a comment about what the event is for and that it's internal-only?

packages/entities/entities-plugins/src/components/PluginForm.vue
kaiarrowood: For now just a placeholder, I didn't mean to actually check this file in, sorry about that!

packages/entities/entities-plugins/src/components/PluginSelect.vue
kaiarrowood: Right, but wouldn't I still need the same conditional? This just means I can remove line `450` right? 
image

adamdehaven: yes

packages/entities/entities-plugins/src/components/DeleteCustomPluginSchemaModal.vue
kaiarrowood: There is now additional logic in this function that wasn't there before

packages/entities/entities-plugins/src/components/DeleteCustomPluginSchemaModal.vue
kaiarrowood: This actually won't matter because the component won't even be rendered if `plugin` isn't passed in.

packages/entities/entities-plugins/src/components/PluginForm.vue
kaiarrowood: You can completely ignore this file, it's a placeholder and I'm going to rip it back out

packages/entities/entities-plugins/src/components/PluginSelect.vue
kaiarrowood: Looks like the scope should allow me to get away without renaming data

packages/entities/entities-plugins/docs/plugin-select.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Events
+
+#### plugin-clicked
+
+An `@plugin-clicked` event is emitted when a plugin in the selection grid is clicked. The event payload is the plugin object.
```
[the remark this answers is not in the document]
kaiarrowood: I don't think so because I think that would mislead the user into thinking it's fired whenever someone clicks anywhere in the `PluginSelect` (or if they for some reason did want to configure a click event on the component it wouldn't do what they expected).

packages/entities/entities-plugins/docs/plugin-select.md
```diff
@@ -0,0 +1,190 @@
...
+
+A synchronous or asynchronous function, that returns a boolean, that evaluates if the user can edit a given custom plugin.
+
+#### `noRouteChange`
+
+- type: `boolean`
+- required: `false`
+- default: `false`
+
+If true, let consuming component handle event when clicking on a plugin. Used in conjunction with `@plugin-clicked` event.
```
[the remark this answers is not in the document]
kaiarrowood: We can't do this because the handler being defined on `PluginSelect` is different than the handler defined on the `PluginSelectCard` component which is the component that actually evaluates what to do on click.

I will change the name.

packages/entities/entities-plugins/src/components/PluginSelectCard.vue
kaiarrowood: I'm assuming we don't want the cards themselves to be resizing based on the space available. This is the width that seems to look best for the design?

image

adamdehaven: Could we use CSS grid to handle the size instead of flex?

Leopoldthecoder:
Can we adjust the size of skeletons so they align with the size of actual plugin cards? Right now they are a lot bigger. Also I think the divider can be moved up a little, and its color can be lighter.

image

packages/entities/entities-plugins/docs/plugin-select.md
Leopoldthecoder: Confirmed this is fixed.

packages/entities/entities-plugins/src/components/PluginSelectCard.vue
kaiarrowood: I don't have time to refactor the CSS right now and this isn't something I know how to do off the top of my head. Could it be a ticket for the backlog?

jillztom (approved):

PR #883, opened 2023-11-01
feat(plugins): form unification [khcp-9513]

kaiarrowood:
# Summary

Unify plugin forms. Includes support for rendering plugins and auth credentials (consumer credentials). 
For [KHCP-9513](https://konghq.atlassian.net/browse/KHCP-9513).

Tests will be added in [KHCP-9775](https://konghq.atlassian.net/browse/KHCP-9775)

## PR Checklist

[KHCP-9513]: https://konghq.atlassian.net/browse/KHCP-9513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[KHCP-9775]: https://konghq.atlassian.net/browse/KHCP-9775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/entities/entities-plugins/src/plugins-endpoints.ts
```diff
@@ -23,12 +23,18 @@ export default {
...
+      entityGetAll: '/api/runtime_groups/{controlPlaneId}/{entity}?size=1000',
     },
     kongManager: {
       create: '/{workspace}/plugins',
       edit: '/{workspace}/plugins/{id}',
       pluginSchema: '/{workspace}/schemas/plugins/{plugin}',
       validate: '/{workspace}/schemas/plugins/validate',
+      // VFG endpoints
+      entityGetOne: '/{workspace}/{entity}/{id}',
+      entityGetAll: '/{workspace}/{entity}?size=1000',
```
Leopoldthecoder, line 37:
```suggestion
      entityGetAll: '/{workspace}/{entity}',
```

When calling `getAll`, `FieldAutoSuggest` has [a pre-defined page size](https://github.com/Kong/public-ui-components/blob/a7e83306e2b303b719219693dc6155f9d132e886/packages/core/forms/src/generator/fields/advanced/FieldAutoSuggest.vue#L222), so here we need to remove `size=1000`, otherwise there'll be two `size` params and Admin API will return 500.

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -0,0 +1,699 @@
...
+      }
+
+      updateModel(props.record.config, 'config')
+    }
+  }
+
+  // Check if incoming field exists in current model and add if so update
+  if (entityIdField.value && props.schema) {
+    const updateFields: Record<string, any> = {}
+    const key = entityIdField.value === 'consumer_group_id' ? 'consumer_group-id' : JSON.stringify(JSON.parse(entityIdField.value)).replace('_', '-')
```
Leopoldthecoder, line 517:
```suggestion
    const key = entityIdField.value === 'consumer_group_id' ? 'consumer_group-id' : entityIdField.value.replace('_', '-')
```

I think `entityIdField.value` is a string.

packages/entities/entities-plugins/src/composables/usePluginMeta.ts
```diff
@@ -1,7 +1,19 @@
...
+import aclsCredentialsSchema from './plugin-schemas/credentials/mockedAclSchema.json'
+import { basicAuthSchema } from './plugin-schemas/BasicAuth'
+import basicAuthCredentialsSchema from './plugin-schemas/credentials/mockedBasicAuthSchema.json'
+import { keyAuthSchema } from './plugin-schemas/KeyAuth'
+import keyAuthCredentialsSchema from './plugin-schemas/credentials/mockedKeyAuthSchema.json'
+import { hmacAuthSchema } from './plugin-schemas/HMAC'
+import hmacAuthCredentialsSchema from './plugin-schemas/credentials/mockedHmacAuthSchema.json'
+import { jwtSchema } from './plugin-schemas/JWT'
+import jwtCredentialsSchema from './plugin-schemas/credentials/mockedJwtSchema.json'
+import OAuth2Schema from './plugin-schemas/OAuth2'
```
Leopoldthecoder, line 15: Besides `OAuth2`, KM has another credential type `KeyAuthEnc`:
```suggestion
import OAuth2Schema from './plugin-schemas/OAuth2'
import KeyAuthEncSchema from './plugin-schemas/KeyAuthEnc'
```

packages/core/forms/src/index.ts
```diff
@@ -1,8 +1,8 @@
-import type { App, Plugin } from 'vue'
-import validators from './generator/utils/validators'
+import type { App } from 'vue'
+// import validators from './generator/utils/validators'
```
adamdehaven, line 2: nit: if some of the code isn't needed in this file (commented out) can you remove it?

packages/entities/entities-plugins/docs/plugin-form.md
```diff
@@ -83,19 +103,67 @@ The base konnect or kongManger config.
...
 
+#### `hideScopeSelection`
+
+- type: `Boolean`
+- required: `false`
+- default: `false`
+
+Manually toggle visibility of plugin scope selection control.
+
+#### `isCredential`
```
adamdehaven, line 114: nit: can we move away from prefixed prop names? `is*` and `has*`

At least for future as well

packages/entities/entities-plugins/docs/plugin-form.md
```diff
@@ -83,19 +103,67 @@ The base konnect or kongManger config.
...
+
+Support instance names for plugins. This can be removed when KHCP-5872-custom-names-for-plugins is removed. Enabled by default for KM.
+
 ### Events
 
 #### error
 
 An `@error` event is emitted when form validation fails. The event payload is the response error.
 
+#### fetch-schema:error
```
adamdehaven, line 144: suggestion: since this is a type of error, I think the syntax should be reversed?
```suggestion
#### error:fetch-schema
```

packages/entities/entities-plugins/docs/plugin-form.md
```diff
@@ -83,19 +103,67 @@ The base konnect or kongManger config.
...
 #### loading
 
 A `@loading` event is emitted when loading state changes. The event payload is a boolean.
 
 #### update
 
-A `@update` event is emitted when the form is saved. The event payload is the plugin object.
+An `@update` event is emitted when the form is saved. The event payload is the plugin object.
+
+#### model-updated
```
adamdehaven, line 156: question: is this event bound to a `v-model`? If yes, then it needs to be `update:model-value`

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -0,0 +1,717 @@
...
+    default: () => ({}),
+  },
+  isEditing: {
+    type: Boolean,
+    default: false,
+  },
+  isCredential: {
+    type: Boolean,
+    default: false,
+  },
```
adamdehaven, line 110: Same comment here on prefixed prop names

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -0,0 +1,717 @@
...
+})
+
+const { axiosInstance } = useAxios({
+  headers: props.config?.requestHeaders,
```
adamdehaven, line 114: `props.config` is required, so you shouldn't need to optionally chain. Same comment for whole PR

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -0,0 +1,717 @@
...
+
+/**
+ * @param {entityType} string - the entity query path WITHOUT leading/trailing slash (ex. 'routes')
+ * @returns {Promise<import('axios').AxiosResponse<T>>}
+ */
+const getAll = (entityType: string, params: AxiosRequestConfig['params']): Promise<AxiosResponse> => {
+  const url = buildGetAllUrl(entityType)
+
+  // TODO: currently hardcoded to fetch 1000 records, and filter
+  // client side. If more than 1000 records, this won't work
```
adamdehaven, line 171: question: why is this TODO? It doesn't sound like we can change it?

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -0,0 +1,717 @@
...
+    if (type === 'object' && value && value.length && scheme.type === 'input') {
+      formModel[modelKey] = value.join(', ')
+      originalModel[modelKey] = value.join(', ')
+
+      return
+    }
+
+    formModel[modelKey] = value
+    originalModel[modelKey] = value
+  })
```
adamdehaven, line 288: suggestion: throughout the PR, there are some complex logic blocks; can you add comments appropriately to help provide context for someone else to understand what's happening and why?

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -0,0 +1,717 @@
...
+
+  emit('model-updated', {
+    model: formModel,
+    originalModel,
+    data: getModel(),
+  })
+}
+
+const getModel = (): Record<string, any> => {
```
adamdehaven, line 297: nit: you're calling this function above before it is defined; you should move the function definition above the `updateModel` block

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -0,0 +1,717 @@
...
+    if (Array.isArray(value)) {
+      formModel[modelKey] = JSON.parse(JSON.stringify(value))
+      originalModel[modelKey] = JSON.parse(JSON.stringify(value))
+
+      return
+    }
+
+    if (type === 'object' && value && !value.length) {
+      return updateModel(value, modelKey)
+    }
```
adamdehaven, line 277: question: there's some recursion here; can you explain it with some comments?

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -0,0 +1,717 @@
...
+        }
+
+        if (field.model) schema[field.model] = field
+      })
+      delete schema[customField]
+    }
+  })
+
+  // Iterate over each field, transform each field value to match the
+  // Kong Admin APIs value expectations
```
adamdehaven, line 330: question: how do we know the logic below matches what the Admin API is expecting?

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
adamdehaven, line 297: suggestion: can you provide a better return type here?

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -0,0 +1,717 @@
...
+          service_id: props.record.service_id,
+          route_id: props.record.route_id,
+          consumer_id: props.record.consumer_id,
+          consumer_group_id: props.record.consumer_group_id,
+          enabled: props.record.enabled,
+        })
+      }
+
+      updateModel(props.record.config, 'config')
+    }
```
adamdehaven, line 520: question: what is this block doing?

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -0,0 +1,717 @@
...
+  }
+
+  loading.value = false
+}
+
+watch(loading, (newLoading) => {
+  emit('loading', newLoading)
+})
+
+watch(() => props.schema, (newSchema) => {
```
adamdehaven, line 556: question: the schema is an object; do you need a `deep` watcher here?

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -0,0 +1,717 @@
...
+
+  .fade-enter-from, .fade-leave-to {
+    opacity: 0;
+  }
+}
+</style>
+
+<style lang="scss">
+.kong-ui-entities-plugin-form {
+  .vue-form-generator {
```
adamdehaven, line 595: suggestion: instead of using unscoped styles here, could you use the `:deep` selector to target the child component(s)?

packages/entities/entities-plugins/src/types/plugin.ts
```diff
@@ -46,6 +53,12 @@ export enum PluginScope {
   CONSUMER_GROUP = 'consumer_group',
 }
 
+export interface PluginEntityInfo {
+  entity: PluginScope,
+  entityEndpoint: EntityType,
```
adamdehaven, line 58: nit: commas aren't needed
```suggestion
  entity: PluginScope
  entityEndpoint: EntityType
```

adamdehaven (commented):
There is a lot of complex code/logic in this PR. It would be great if you could go through and add context/comments in a lot of areas.

The other concern I have is that a lot of objects are just typed as `Record<string, any>` which doesn't provide much type safety; however, given the schemas that are passed in I'm unsure how possible better type safety could be provided as-is

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
Leopoldthecoder: I checked the file history and these lines can be removed:
```diff
    if (props.record.data) {
-      const newModel = props.record.data
-      newModel.client_certificate = newModel.client_certificate?.id
      updateModel(props.record.data)
    } else if (props.record.config) {
      if (props.record.consumer_id || props.record.service_id || props.record.route_id || props.record.consumer_group_id) {
        updateModel({
          service_id: props.record.service_id,
          route_id: props.record.route_id,
          consumer_id: props.record.consumer_id,
          consumer_group_id: props.record.consumer_group_id,
          enabled: props.record.enabled,
        })
      }

      updateModel(props.record.config, 'config')
    }
```

They are from [this PR](https://github.com/Kong/kong-admin/pull/2031/files#diff-67c30c7f791df36e84579a8d81db0c0bd180c11ade995771b87c3656a3528d7bR447) and have nothing to do with plugins.

packages/entities/entities-plugins/docs/plugin-form.md
kaiarrowood: Nope, this is just the name used everywhere in VFG for the update event to any fields in the form. I understand it's confusing but it's used in so many places throughout the code in both repos I'm hesitant to change it.

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
kaiarrowood: Because we make it? This is the logic to transform the formdata into the request structure expected for the API

packages/entities/entities-plugins/src/components/PluginForm.vue
```diff
@@ -90,130 +147,687 @@ const props = defineProps({
...
+  // ex. 'acl'
+  name: {
+    default: props.pluginType,
+    type: 'input',
+    inputType: 'hidden',
+    styleClasses: 'd-none',
+  },
+  // plugin scoping
+  selectionGroup: {
+    type: props.hideScopeSelection || (formType.value === EntityBaseFormType.Create && props.config.entityId) ? 'foreign' : 'selectionGroup',
```
Leopoldthecoder, line 285: The comment for `hideScopeSelection` says `Rather than using internal logic to determine whether or not to hide scope, force it`, so could we make `hideScopeSelection` take precedence over internal checking?
```suggestion
    type: !props.hideScopeSelection ? 'selectionGroup' : ((formType.value === EntityBaseFormType.Create && props.config.entityId) ? 'foreign' : 'selectionGroup'),
```

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -0,0 +1,734 @@
...
+      // scope and top level fields
+      if (props.record.consumer_id || props.record.service_id || props.record.route_id || props.record.consumer_group_id) {
+        updateModel({
+          service_id: props.record.service_id,
+          route_id: props.record.route_id,
+          consumer_id: props.record.consumer_id,
+          consumer_group_id: props.record.consumer_group_id,
+          enabled: props.record.enabled,
+        })
+      }
```
Leopoldthecoder, line 535: When editing a plugin that is scoped to both a service and a route, the scope input is blank. Not sure if the following is the right fix, but it does avoid the blank input:
```suggestion
      if (props.record.consumer?.id || props.record.service?.id || props.record.route?.id || props.record.consumer_group?.id) {
        updateModel({
          'service-id': props.record.service?.id,
          'route-id': props.record.route?.id,
          'consumer-id': props.record.consumer?.id,
          'consumer_group-id': props.record.consumer_group?.id,
          enabled: props.record.enabled,
        })
      }
```

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -0,0 +1,734 @@
...
+
+      // main plugin configuration
+      updateModel(props.record.config, 'config')
+    }
+  }
+
+  // scoping logic, convert _ to - for form model
+  // Check if incoming field exists in current model and if so update
+  if (entityIdField.value && props.entityId && props.schema) {
```
Leopoldthecoder, line 544: This `if` block is for creating a plugin from within a specific entity, right? If so, we need to change the condition:
```suggestion
  if (entityIdField.value && props.entityId && props.schema && !props.editing) {
```

Because in the above scenario (editing a plugin that is scoped to both a service and a route), `entityIdField.value` is `route_id`, and `props.entityId` is the ID of the service (due to the order of route and service when assigning `entity` and `id` [here](https://github.com/Kong/public-ui-components/blob/fe14ce40824b00b92ec37ca1bba2c47b55579e8b/packages/entities/entities-plugins/src/components/PluginForm.vue#L243-L261)), so this block will assign the service ID to `updateFields.route_id`, which confuses the scope input.

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -0,0 +1,734 @@
...
+
+  :deep(.vue-form-generator) {
+    & > fieldset {
+      border: none;
+      padding: $kui-space-0;
+    }
+
+    .field-switch .field-wrap label {
+      .label {
+        background-color: $kui-color-background-neutral-weak;
```
Leopoldthecoder, line 620: This line makes the switch gray even when it is checked. Is it intentional?
image

portikM (approved):
Didn't do a full review, approving to unblock merging based on feedback from other devs here and in the Slack thread 👍

PR #889, opened 2023-11-03
feat(entities-shared): fuzzy search filter

Leopoldthecoder:
# Summary

This PR adds fuzzy search support to `useDebouncedFilter` when in Kong Manager if we have more than 100 records.

## PR Checklist

### DIFF
--- a/packages/entities/entities-shared/src/composables/useDebouncedFilter.ts
+++ b/packages/entities/entities-shared/src/composables/useDebouncedFilter.ts
@@ -114,16 +114,16 @@ export default function useDebouncedFilter(
           } else {
             results.value = []
           }
-        } else { // Admin API supports filtering on specific fields
+        } else if (query) { // Admin API supports filtering on specific fields
           const promises = []
 
-          if (isValidUuid(query || '') && keys.searchKeys.includes('id')) {
+          if (isValidUuid(query) && keys.searchKeys.includes('id')) {
             // If query is a valid UUID, do the exact search
             promises.push((async () => {
-              const res: Record<string, any> = await axiosInstance.get(`${url}/${query}`)
-              return [res[keys.fetchedItemsKey]]
+              const { data } = await axiosInstance.get(`${url}/${query}`)
+              return [data[keys.fetchedItemsKey] ?? data]
             })())
-          } else if (query) {
+          } else {
             // Search on fields with backend filtering support
             promises.push(
               ...keys.searchKeys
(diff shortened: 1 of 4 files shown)


packages/entities/entities-shared/src/composables/useDebouncedFilter.ts
```diff
@@ -93,20 +98,58 @@ export default function useDebouncedFilter(
...
+              ...keys.searchKeys
+                .filter(key => key !== 'id')
+                .map(async key => {
+                  const { data } = await axiosInstance.get(`${url}?${key}=${query}`)
+                  return data[keys.fetchedItemsKey]
+                }),
+            )
+          }
+
+          if (query) {
```
nekolab, line 138: Maybe we can hoist this `if` to line 118 to prevent issuing unnecessary network requests

packages/entities/entities-shared/src/composables/useDebouncedFilter.ts
```diff
@@ -93,20 +98,58 @@ export default function useDebouncedFilter(
...
-        const { data }: Record<string, any> = await axiosInstance.get(`${currUrl}?size=${size}`)
+          const { data }: Record<string, any> = await axiosInstance.get(`${currUrl}?size=${size}`)
 
-        if (keys.fetchedItemsKey in data) {
-          results.value = data[keys.fetchedItemsKey]
-        } else if (data?.id) { // exact match
-          results.value = [data]
-        } else {
-          results.value = []
+          if (keys.fetchedItemsKey in data) {
```
nekolab, line 110: do we still need this `if` block since KoKo supports exact match only?

packages/entities/entities-shared/src/composables/useDebouncedFilter.ts
```diff
@@ -93,20 +98,58 @@ export default function useDebouncedFilter(
...
+            results.value = []
+          }
+        } else { // Admin API supports filtering on specific fields
+          const promises = []
+
+          if (isValidUuid(query || '') && keys.searchKeys.includes('id')) {
+            // If query is a valid UUID, do the exact search
+            promises.push((async () => {
+              const res: Record<string, any> = await axiosInstance.get(`${url}/${query}`)
+              return [res[keys.fetchedItemsKey]]
```
nekolab, line 124: Not sure but maybe it should be `return [res.data]` here. I think it happens to work because the `fetchedItemsKey` defaults to `"data"`.

nekolab (changes requested):

nekolab:
Nit: I found `searchKeys` in `packages/entities/entities-consumer-groups/src/components/ConsumerGroupForm.vue` doesn't contain `'id'`, not sure if it's an intended behavior or not 😕

packages/entities/entities-shared/src/composables/useDebouncedFilter.ts
Leopoldthecoder: If `query` is empty, the `promises` array will be empty as well. But hoisting this `if` up does make the code cleaner, so I did it.

Leopoldthecoder:
> nit: I found `searchKeys` in `packages/entities/entities-consumer-groups/src/components/ConsumerGroupForm.vue` doesn't contain `'id'`, not sure if it's an intended behavior or not 😕

Good catch. I added `id`.

packages/entities/entities-shared/src/composables/useDebouncedFilter.ts
Leopoldthecoder: You are right. I changed this to
``` js
const { data } = await axiosInstance.get(`${url}/${query}`)
return [data[keys.fetchedItemsKey] ?? data]
```

nekolab (approved):
LGTM 👍

PR #891, opened 2023-11-04
feat(entities-routes): add filter api for routes list

jillztom:
# Summary

Utilizing the filter API for entity routes
```
us.api.konghq.tech/konnect-api/api/runtime_groups/ff7e9ee5-5cc5-48f6-818a-154faef8270e/routes/foo
```
will be changed to 
```
us.api.konghq.tech/v2/control-planes/ff7e9ee5-5cc5-48f6-818a-154faef8270e/core-entities/routes?filter[name]=foo
```

## PR Checklist


vite.config.shared.ts
```diff
@@ -160,11 +161,21 @@ export const getApiProxies = (pathToRoot: string = '../../../.') => {
         ...konnectAuthHeader,
       },
     }
+    // @ts-ignore
+    regionalV2Proxies[`^/${region}/konnect-v2`] = {
```
adamdehaven, line 165: question: Is there a reason you moved away from the `kong-api/` proxied URL? We have this as a standard across most of the repos

adamdehaven: Previously, it was set to this:

`'/us/kong-api/konnect-api'`

It’s now being set to this:

`'/us/konnect-v2'`

Instead, you should use something like this to keep kong-api in the proxied URL:

`'/us/kong-api/v2/konnect-api'`

QueaT-kong: @jillztom could u refer to the question?

QueaT-kong: because I'm not sure about answer to that

jillztom: We do not have a region specific proxying for `v2` endpoints if I'm not mistaken. 

`'^/kong-api/v2'` is hitting the `https://global.api.konghq.tech`

adamdehaven: This is for the global one: https://github.com/Kong/public-ui-components/blob/5cf51320246505fba2958ea229aa9f7a71a5a5a5/vite.config.shared.ts#L187-L194

So then something like `'^/kong-api/us/v2' maybe? 

When Val and I set up all the proxying, we tried to standardize on `kong-api` in the path just to make it easy to debug, expectations, etc.

ValeryG:
if we change the line https://github.com/Kong/public-ui-components/blob/main/vite.config.shared.ts#L155 to read 
```
-regionalProxies[`^/${ region }/kong-api/konnect-api`]
+regionalProxies[`^/${ region }/kong-api`]
```
it will handle both `konnect-api` and `v2` and we do not need to change apiBaseUrl in the konnectConfigs...

vite.config.shared.ts
jillztom: We can do 

```
regionalV2Proxies[`^/${region}/kong-api/v2`] = {
      target: (env.VITE_KONNECT_API ?? '').replace(/\{geo\}/, region),
      rewrite: (path: string) => path.replace(`/${region}/kong-api`, ''),
      changeOrigin: true,
      headers: {
        ...konnectAuthHeader,
      },
    }
```

Just have to confirm it won't collide with `'^/kong-api/v2'`. It shouldn't because we are having this regional proxy before the global proxy.

adamdehaven: Correct, the region comes first so it won't conflict

jillztom:
> if we change the line https://github.com/Kong/public-ui-components/blob/main/vite.config.shared.ts#L155 to read
> 
> ```
> -regionalProxies[`^/${ region }/kong-api/konnect-api`]
> +regionalProxies[`^/${ region }/kong-api`]
> ```
> 
> it will handle both `konnect-api` and `v2`

There we go! Even better :)

vite.config.shared.ts
```diff
@@ -149,10 +149,11 @@ export const getApiProxies = (pathToRoot: string = '../../../.') => {
   // Add additional regions as they become available
   const availableRegions = ['au', 'eu', 'us']
   const regionalProxies = {}
+  const regionalV2Proxies = {}
   // Build the regional API proxies
   for (const region of availableRegions) {
     // @ts-ignore
-    regionalProxies[`^/${ region }/kong-api/konnect-api`] = {
+    regionalProxies[`^/${ region }/kong-api`] = {
```
adamdehaven, line 156: Some formatting seems off in this file; can you run `pnpm run lint:fix` from the root and push up any changes, if there are any?

QueaT-kong: no file changes after command

PR #897, opened 2023-11-06
refactor(TopN): remove support for old explore v3 format [MA-2094]

adorack:
- Remove union type for explore v3 display blob
- Remove shims for old display blob support

BREAKING CHANGE: removed support for old display blob style.

# Summary

## PR Checklist


adamdehaven (commented):
FWIW, this package is still on a `0.x` version, so a breaking commit will only bump the `minor` version (meaning, I don't think it makes much of a difference with the current state of the component).

Once the component is bumped to `1.x` then this **would** trigger a major version bump, which I'm not sure is warranted based on the change, but I'll leave that up to the Analytics team

filipgutica (approved):

PR #906, opened 2023-11-10
MA-2278: entity link shared entity

mihai-peteu:
# Summary

Needed for Gateway Manager MFE. 

https://konghq.atlassian.net/browse/MA-2278

### Intro
`EntityLink` has been moved here from Konnect, along with a `useExternalLinkCreator` composable which builds a full URL from the provided array of sequential route parameters.

### Why? 
We cannot rely on the Vue router to build the links internally in the MFE world, because MFEs are not aware of each others L2 routes.  Therefore, we have to pass in the absolute URL from the consuming app.

### Changes
- EntityLink now contains an `<ExternalLink>` instead of a `<router-link>`
- EntityLink uses a simpler notification on UUID copy (tooltip hover text changes), to avoid using the Toaster
- The consuming app will need to import `useExternalLinkCreator()` from this package, which is used to build the full URL, then passing that as a string to `EntityLink`

Please refer to matching Konnect PR for implementation:
https://github.com/Kong/khcp-ui/pull/7138/files

output

## PR Checklist


packages/entities/entities-shared/src/components/entity-link/EntityLink.vue
```diff
@@ -0,0 +1,134 @@
+<template>
```
filipgutica, line 1: needs a component test

packages/entities/entities-shared/src/composables/useExternalLinkCreator.ts
```diff
@@ -0,0 +1,18 @@
+export default function useExternalLinkCreator(routeParams: string[]): string {
```
filipgutica, line 1: needs a unit test

filipgutica (changes requested):
just need to add some tests, looks good otherwise,

packages/entities/entities-shared/src/components/entity-link/EntityLink.vue
```diff
@@ -0,0 +1,134 @@
...
+    type: String,
+    required: true,
+  },
+})
+
+const textContent = ref<HTMLElement>()
+const { isTruncated } = composables.useTruncationDetector(textContent as Ref<HTMLElement>)
+
+const { i18n: { t } } = composables.useI18n()
+const { notify } = composables.useToaster()
```
adamdehaven, line 70: question: can we move away from using a toaster here? 

I would think just changing the tooltip text on click would be enough, or this component could just emit an event to the host app.

packages/entities/entities-shared/src/components/entity-link/EntityLink.vue
```diff
@@ -0,0 +1,134 @@
...
+  if (notifyParam.type === 'success') {
+    notify({
+      appearance: 'success',
+      message: t('global.actions.copied_id', { id: entityUuid.value }),
+    })
+  } else {
+    notify({
+      appearance: 'danger',
+      message: t('global.actions.copyToClipboardFailed'),
+    })
```
adamdehaven, line 98: See my other comment on removing the toaster altogether, but I also don't think we need to show something if copying fails; it's close to near-impossible and there's nothing we can do about it

packages/entities/entities-shared/src/components/entity-link/EntityLink.vue
mihai-peteu: If we emit an event to the host app, then we end up with duplicated code in multiple places in the consuming app...

I prefer the tooltip text change on click, if this is an established pattern

adamdehaven: Yes, this is the pattern we use for copy events, etc.

packages/entities/entities-shared/src/composables/useExternalLinkCreator.ts
mihai-peteu: Added some basic tests. Because we create the base url inside `useExternalLinkCreator`, the only way to make this function fail is by providing an empty route param, or one that doesn't match the regex I defined.


(Note: I used the same regex as https://github.com/Kong/public-ui-components/blob/main/packages/core/forms/src/generator/utils/validators.ts#L209)

packages/entities/entities-shared/src/components/entity-link/EntityLink.vue
```diff
@@ -0,0 +1,159 @@
...
+          <CopyIcon
+            class="copy-icon"
+            :color="KUI_COLOR_TEXT_PRIMARY"
+            data-testid="copy-id"
+            :size="16"
+          />
+        </span>
+      </KClipboardProvider>
+    </KTooltip>
+    <!-- <CopyUuid
```
filipgutica, line 54: nit: should we not just delete this block?

packages/entities/entities-shared/src/components/entity-link/EntityLink.cy.ts
```diff
@@ -0,0 +1,119 @@
+import EntityLink from './EntityLink.vue'
+// import type { EntityLinkData } from '../../types'
```
filipgutica, line 2: nit: do we need the commented out stuff checked in?

filipgutica (changes requested):
last 2 comments looks all good otherwise.

packages/entities/entities-consumer-groups/package.json
```diff
@@ -64,7 +64,7 @@
     "extends": "../../../package.json"
   },
   "distSizeChecker": {
-    "errorLimit": "400KB"
+    "errorLimit": "410KB"
```
mihai-peteu, line 67: @Leopoldthecoder | @adamdehaven Had to bump this slightly, the build was unexpectedly failing otherwise:

https://github.com/Kong/public-ui-components/actions/runs/6897551205

filipgutica (approved):

mihai-peteu:
@sumimakito Apologies, I meant this PR in public-ui 

( the [khcp-ui PR](https://github.com/Kong/khcp-ui/pull/7138) will get merged in after this one )

packages/entities/entities-shared/src/types/entity-link.ts
```diff
@@ -0,0 +1,5 @@
+export interface EntityLinkData {
+  id: string,
+  label?: string,
+  deleted: boolean
```
sumimakito, line 4:
```suggestion
  deleted?: boolean
```

trivia: Maybe we can make this field optional?

packages/entities/entities-shared/src/components/entity-link/EntityLink.vue
```diff
@@ -0,0 +1,149 @@
...
+const textContent = ref<HTMLElement>()
+const { isTruncated } = composables.useTruncationDetector(textContent as Ref<HTMLElement>)
+const { i18n: { t } } = composables.useI18n()
+
+const tooltipDefaultText = t('global.actions.copyId')
+const copyUuidTooltipText = ref(tooltipDefaultText)
+
+// If the dimension value is the same as the label, it means the entity has been deleted
+const isDeletedEntity = computed(() => {
+  return props.entityLinkData.deleted
```
sumimakito, line 86:
```suggestion
  return props.entityLinkData.deleted ?? false
```

… if we make `deleted` optional

sumimakito (dismissed):
The PR appears to be in good shape overall. I've noticed that Kong Manager and other shared components aren't using `<EntityLink />`. These comments are non-blocking. Please feel free to merge it whenever you're ready. 👍

sumimakito (approved):
LGTM

PR #913, opened 2023-11-14
feat(*): add support to EntityBaseTable to use action button outside [KHCP-9618]

mptap:
# Summary

https://konghq.atlassian.net/browse/KHCP-9618

### Entities Updated:

- Gateway Services
- Routes
- Consumers
- ConsumerGroups
- ConsumerCredentials
- Plugins
- Upstreams
- Targets
- Certificates
- CACertificates
- SNIs
- Vaults
- Keys
- KeySets

## PR Checklist

### DIFF
--- a/packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
+++ b/packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
@@ -49,7 +49,6 @@
         data-testid="page-header-actions"
       >
         <slot name="actions" />
-        <div id="kong-ui-app-page-header-action-button" />
       </div>
     </div>
 
--- a/packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
+++ b/packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
@@ -49,7 +49,7 @@
         data-testid="page-header-actions"
       >
         <slot name="actions" />
-        <div id="action-button" />
+        <div id="kong-ui-app-page-header-action-button" />
       </div>
     </div>
 
--- a/packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
+++ b/packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
@@ -49,6 +49,7 @@
         data-testid="page-header-actions"
       >
         <slot name="actions" />
+        <div id="action-button" />
       </div>
     </div>
 
--- a/packages/entities/entities-certificates/src/components/CACertificateList.vue
+++ b/packages/entities/entities-certificates/src/components/CACertificateList.vue
@@ -32,14 +32,16 @@
           to="#kong-ui-app-page-header-action-button"
         >
           <PermissionsWrapper :auth-function="() => canCreate()">
-            <KButton
-              appearance="primary"
-              data-testid="toolbar-add-ca-certificate"
-              icon="plus"
-              :to="config.createRoute"
-            >
-              {{ t('ca-certificates.list.toolbar_actions.new_ca_certificate') }}
-            </KButton>
+            <KTooltip :label="t('ca-certificates.list.empty_state.description')">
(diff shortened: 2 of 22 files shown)


packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
```diff
@@ -49,6 +49,7 @@
         data-testid="page-header-actions"
       >
         <slot name="actions" />
+        <div id="action-button" />
```
adamdehaven, line 52: this is a pretty generic class - it may make sense to change this to something more specific, e.g. `kong-ui-app-page-header-action-button`

packages/entities/entities-shared/src/components/entity-base-table/EntityBaseTable.vue
```diff
@@ -44,7 +44,16 @@
               v-if="$slots['toolbar-button']"
               class="toolbar-button-container"
             >
-              <slot name="toolbar-button" />
+              <Teleport
+                v-if="!useActionOutside"
+                to="#action-button"
```
adamdehaven, line 49: will need to change the selector here as well

packages/core/app-layout/src/components/pageHeader/AppPageHeader.vue
mptap: Updated

packages/entities/entities-shared/src/components/entity-base-table/EntityBaseTable.vue
mptap: Updated

packages/entities/entities-gateway-services/src/components/GatewayServiceList.vue
```diff
@@ -27,13 +28,16 @@
...
       <template #toolbar-button>
         <PermissionsWrapper :auth-function="() => canCreate()">
-          <KButton
-            appearance="primary"
-            icon="plus"
-            :to="config.createRoute"
-          >
-            {{ t('gateway_services.list.toolbar_actions.new_gateway_service') }}
-          </KButton>
+          <Teleport to="#kong-ui-app-page-header-action-button">
```
jillztom, line 31: Let's use 

```
<Teleport :to="'#action-buttons-'+cacheIdentifier" :disabled="!useActionOutside">
  <PermissionsWrapper :auth-function="() => canCreate()">
    <KButton
      appearance="primary"
      data-testid="toolbar-add-route"
      icon="plus"
      :to="config.createRoute"
   >
    {{ t('routes.list.toolbar_actions.new_route') }}
   </KButton>
  </PermissionsWrapper>
</Teleport>
```

I just used `#action-buttons-'+cacheIdentifier` as the `id`. Feel free to anything you like.

mptap: Updated

jillztom:
https://github.com/Kong/public-ui-components/assets/6960750/b7bcaf58-07cf-4e26-8be3-5d9e1bb21ebf


The call to action buttons are jumping from one place to another when you navigate around. This is probably because of the disabled state of the Teleport and hasData.

`:disabled="!useActionOutside || !hasData"` this logic needs to be updated. When the component is getting mounted, the `hadData` is false and the teleport will be disabled causing the jump.

packages/entities/entities-certificates/src/components/CACertificateList.vue
```diff
@@ -13,6 +13,7 @@
       preferences-storage-key="kong-ui-entities-ca-certificates-list"
       :query="filterQuery"
       :table-headers="tableHeaders"
+      :use-action-outside="useActionOutside"
```
Leopoldthecoder, line 16: @mptap Should this line be removed? Because `EntityBaseTable` does not have the `use-action-outside` prop.

PR #920, opened 2023-11-16
fix(entities-list): show response error message in entity list

nekolab:
The current implementation just display general error message when error happens, and `<EntityBaseTable />` doesn't support set error message for the list (only error title allowed).

To display the message returned from server, prop `errorMessage` in `<EntityBaseTable />` now supports both string form (title only) and object form (title and message).

All kinds of entity lists are modified to set the error message when available.

# Summary

## PR Checklist

### DIFF
--- a/packages/entities/entities-routes/src/components/RouteList.vue
+++ b/packages/entities/entities-routes/src/components/RouteList.vue
@@ -203,6 +203,7 @@ import type {
   ExactMatchFilterConfig,
   FilterFields,
   FuzzyMatchFilterConfig,
+  TableErrorMessage,
 } from '@kong-ui-public/entities-shared'
 import '@kong-ui-public/entities-shared/dist/style.css'
 
@@ -349,7 +350,7 @@ const resetPagination = (): void => {
 /**
  * loading, Error, Empty state
  */
-const errorMessage = ref('')
+const errorMessage = ref<TableErrorMessage>(null)
 
 /**
  * Copy ID action
@@ -483,14 +484,19 @@ const confirmDelete = async (): Promise<void> => {
  */
 watch(fetcherState, (state) => {
   if (state.status === FetcherStatus.Error) {
-    errorMessage.value = t('errors.general')
+    errorMessage.value = {
+      title: t('errors.general'),
+    }
+    if (state.error?.response?.data?.message) {
+      errorMessage.value.message = state.error.response.data.message
+    }
     // Emit the error for the host app
     emit('error', state.error)
 
     return
   }
 
-  errorMessage.value = ''
+  errorMessage.value = null
 })
 
 // Initialize the empty state options assuming a user does not have create permissions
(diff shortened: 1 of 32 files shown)


Leopoldthecoder (dismissed):
LGTM👍

packages/entities/entities-routes/src/components/RouteList.vue
```diff
@@ -349,7 +349,7 @@ const resetPagination = (): void => {
 /**
  * loading, Error, Empty state
  */
-const errorMessage = ref('')
+const errorMessage = ref<{ title?: string, message?: string } | null>(null)
```
adamdehaven, line 352: If you're using this interface everywhere, instead of copy/paste, can we export the type from the `entities-shared` package and use it as needed instead of replicating?

packages/entities/entities-shared/src/components/entity-base-table/EntityBaseTable.vue
```diff
@@ -239,6 +240,22 @@ const emit = defineEmits<{
...
+      title: props.errorMessage,
+      message: undefined,
+    }
+  } else {
+    return {
+      hasError: !!props.errorMessage,
+      title: props.errorMessage?.title,
+      message: props.errorMessage?.message,
+    }
+  }
```
adamdehaven, line 256: IMO this interface is confusing; how does each component know whether to utilize a `string` or the `object` interface? Shouldn't we standardize on one or the other?

packages/entities/entities-routes/src/components/RouteList.vue
```diff
@@ -483,14 +483,19 @@ const confirmDelete = async (): Promise<void> => {
  */
 watch(fetcherState, (state) => {
   if (state.status === FetcherStatus.Error) {
-    errorMessage.value = t('errors.general')
+    errorMessage.value = {
+      title: t('errors.general'),
+    }
+    if (state.error?.response?.data?.message) {
+      errorMessage.value.message = state.error.response.data.message
+    }
```
adamdehaven, line 491: question here, but applies to all instances: are there times we would not want to surface the error message from the API to the user? Isn't that the point of the more generic messaging?

How does a downstream component know whether to display the `errorMessage.value.title` or the `errorMessage.value.message`? How to know which is appropriate?

sumimakito: (Just) Some ideas for better UX: Maybe we can try to fold the "technical information" and show a general error message by default sometime in the future when we decide not to surface the detailed error. 

_Context: I'm recently fixing the OpenID Connect feature and playing with Okta_

image

packages/entities/entities-shared/src/components/entity-base-table/EntityBaseTable.vue
nekolab: This design is for the backward compatibility, I don't want introduce a breaking change here because I'm not sure if we have some project using `<EntityBaseTable />` directly

packages/entities/entities-routes/src/components/RouteList.vue
nekolab: @adamdehaven Good question.
My opinion is, for the Kong Manager, we should always display the message returned by server. Since KM is designed as the frontend of the Admin API, anything returned from Admin API need to be present to the user.

For Konnect, it could be more complicated. I agree with @sumimakito, we can always use pre-defined phrase for these scenarios and hide server response in "technical details" for better user experience

Leopoldthecoder (approved):

PR #933, opened 2023-11-20
fix(entity-link): more robust url regex

mihai-peteu:
# Summary

URL regex should allow dev preview URLs / .tech / .com domains, but also`localhost` with `http` protocol, followed 

## PR Checklist


packages/entities/entities-shared/src/composables/useExternalLinkCreator.ts
```diff
@@ -7,11 +7,12 @@ export default function useExternalLinkCreator(routeParams: string[]): string {
     const fullUrl = `${baseUrl}${path}`
     const hasEmptyParams = routeParams.some(item => item.trim() === '')
 
-    /* Test validity of generated URL */
-    const urlRegex = /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,4}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g // eslint-disable-line no-useless-escape
+    // Test validity of generated URL; only allows `http` protocal if domain is `localhost`
+    /* eslint-disable-next-line no-useless-escape */
+    const validUrlRegex = /^(https:\/\/(www\.)?([a-zA-Z0-9-]+\.){1,}[a-zA-Z]{2,}(:[0-9]+)?(\/[^\/]*)*)$|^http:\/\/localhost(:[0-9]+)?(\/[^\/]*)*$/
```
hfukada, line 12: technically i think the last `[^\/]*` should probably be `+` if you want to avoid doubleslash?
```
> t = /^(\/[^\/]*)*$/
/^(\/[^\/]*)*$/
> t.test('/a//')
true
> t.test('//a')
true
> t = /^(\/[^\/]+)*$/
/^(\/[^\/]+)*$/
> t.test('//a')
false
> t.test('/a//')
false
```

mihai-peteu: Yes, but we do want to allow a single slash, since we might want to land on `/gateway-manager/176e2ec5-d3c0-4268-ac25-d8716ed8c14f/routes` for example.

mihai-peteu: Scratch that, we don't do trailing slashes. Pushed a commit w/ `[^\/]+`.

packages/entities/entities-shared/src/composables/useExternalLinkCreator.ts
```diff
@@ -7,11 +7,12 @@ export default function useExternalLinkCreator(routeParams: string[]): string {
     const fullUrl = `${baseUrl}${path}`
     const hasEmptyParams = routeParams.some(item => item.trim() === '')
 
-    /* Test validity of generated URL */
-    const urlRegex = /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,4}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g // eslint-disable-line no-useless-escape
+    // Test validity of generated URL; only allows `http` protocal if domain is `localhost`
+    /* eslint-disable-next-line no-useless-escape */
+    const validUrlRegex = /^(https:\/\/(www\.)?([a-zA-Z0-9-]+\.){1,}[a-zA-Z]{2,}(:[0-9]+)?(\/[^\/]+)*)$|^(https|http):\/\/localhost(:[0-9]+)?(\/[^\/]+)*$/
```
adamdehaven, line 12: Would it make sense to add a unit test for this composable to test happy and bad paths?

mihai-peteu: Good call - pushed a commit with a few more failing cases

adamdehaven (approved):
lgtm, defer to @hfukada for final

PR #940, opened 2023-11-22
feat(csv-export-modal): migrate from Konnect [KHCP-8746]

mihai-peteu:
# Summary
https://konghq.atlassian.net/browse/KHCP-9746

Refer to khcp-ui PR below which pulls in this package, and replaces the existing component
https://github.com/Kong/khcp-ui/pull/7283

Screenshot 2023-12-11 at 7 12 47 PM

## PR Checklist


adamdehaven:
One thing I'm noticing, `@kong-ui-public/analytics-chart` and `@kong-ui-public/dashboard-renderer` appear to have circular dependencies on each other; this needs to be resolved for one or the other

adamdehaven:
Other things I'm noticing:

- https://github.com/Kong/public-ui-components/blob/00763fd2a5564bc568dd9b5fbc29d06af3f88677/packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts#L84-L94 You're exporting constants from the `/types` directory alongside interfaces instead of from something like `/src/constants.ts` - not ideal TBH
- The only way I can get the unit test to pass for `dashboard-renderer` is to comment out [this line](https://github.com/Kong/public-ui-components/blob/00763fd2a5564bc568dd9b5fbc29d06af3f88677/packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts#L48) which is importing `ChartMetricDisplay` (and its corresponding import at the top of the file

mihai-peteu:
With `dashboard-renderer-types`, the idea was to derive types from the JSON schema, which is why those are objects, not interfaces. 

I'll let @adorack have a look at this, there's got to be a simple fix. Would be nice to continue leveraging `FromSchema` [ [link](https://github.com/ThomasAribart/json-schema-to-ts#stop-typing-twice-%EF%B8%8F) ].

packages/analytics/analytics-chart/src/components/VueJsonCsv.vue
```diff
@@ -18,7 +18,19 @@ import { unparse } from 'papaparse'
...
+  Number = 'number',
+  Boolean = 'boolean',
+  Object = 'object',
+  Function = 'function',
+  Undefined = 'undefined',
+  Symbol = 'symbol'
+}
+
+const isType = (value: any, type: ValidType) => {
+  return typeof value === type
```
mihai-peteu, line 32: @filipgutica this is the line giving me trouble, even though I'm using that enum up above

`Typeof comparisons should be to string literals`

packages/analytics/analytics-chart/src/components/VueJsonCsv.vue
```diff
@@ -118,22 +99,19 @@ const exportableData = computed(() => {
...
-      !isType(labels, ValidType.Function) &&
-      !isType(labels, ValidType.Object)
-  ) {
+
+  if (![ValidType.Undefined, ValidType.Function, ValidType.Object].includes(typeof labels as ValidType)) {
     throw new Error('Labels needs to be a function(value,key) or object.')
   }
 
-  if (isType(labels, ValidType.Function)) {
+  if (typeof labels as ValidType === ValidType.Function) {
```
filipgutica: Rather than doing all these changes, if this works, then you can just keep the original function and update it to be 

```
const isType = (value: any, type: ValidType) => {
  return typeof value as ValidType === type
}
```

packages/analytics/analytics-chart/src/components/VueJsonCsv.vue
```diff
@@ -9,35 +9,24 @@
...
-  Object = 'object',
-  Function = 'function',
-  Undefined = 'undefined',
-  Symbol = 'symbol'
-}
-
-const isType = (value: any, type: ValidType) => {
-  return typeof value === type
-}
-
```
filipgutica:
```
const isType = (value: any, type: ValidType) => {
  return typeof value as ValidType === type
}
```

This works and passes the eslint check

filipgutica (changes requested):

packages/analytics/analytics-chart/src/components/vue-json-csv/VueJsonCsv.vue
```diff
@@ -0,0 +1,178 @@
...
+  if (typeof props.fields as ValidType !== ValidType.Object && !Array.isArray(fields)) {
+    throw new Error('Fields needs to be an array of strings.')
+  }
+
+  if (Array.isArray(fields)) {
+    // Keep only requested object properties
+    return (item: Function) => {
+      return pick(item, fields)
+    }
+  }
```
mihai-peteu, line 126: Useful to keep this code that optionally truncates each row, keeping only the requested `fields` (aka columns)?

adorack (changes requested):
I would like to review.

adorack (changes requested):
I'd like to review.

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
```diff
@@ -184,6 +210,8 @@ watch(() => props.height, (newHeight) => {
   heightRef.value = newHeight
 })
 
+const rawChartData = computed(() => toRef(props, 'chartData'))
```
adorack, line 213: You don't need to wrap this in a `computed` -- you can just say `composables.useChartSelectedRange(toRef(props, 'chartData'))`.  (`toRef` takes care of ensuring that you still have a reactive thing to pass to the function.)

packages/analytics/analytics-chart/src/components/mockData.ts
adorack, line 1: Adam commented at one point that putting this in the `src` directory may increase the bundle size of the package.  @adamdehaven, where is the best place to put mock data?  Does it still increase the bundle size even if it's not included in any module tree exported from the package?

adorack (changes requested):

packages/analytics/analytics-chart/src/components/mockData.ts
adamdehaven: If it's not exported it shouldn't impact the package; however, you could create a `/packages/analytics/analytics-chart/fixtures/` directory

packages/analytics/analytics-chart/src/components/AnalyticsChart.cy.ts
```diff
@@ -723,6 +238,57 @@ describe('<AnalyticsChart />', () => {
...
+        chartTitle: 'Requests',
+      },
+    })
+    cy.getTestId('csv-export-button').should('not.exist')
+  })
+
+  it('does not render an "Export" button if chart data is present but prop is set to `false`', () => {
+    cy.mount(AnalyticsChart, {
+      props: {
+        allowCsvExport: true,
```
adorack, line 258: Based on the test description, this should be `false` and the `chartData` prop should be `exploreResult`.

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
```diff
@@ -175,6 +201,7 @@ const props = defineProps({
 
 const { i18n } = composables.useI18n()
 const heightRef = ref<string>(props.height)
+const csvFilename = computed<string>(() => props?.chartTitle ? props.chartTitle : i18n.t('csvExport.defaultFilename'))
```
adorack, line 204: Experimenting with this in Konnect: because we don't set a chart title in reports, this way of deriving `csvFilename` means that reports have an empty filename prefix.  I think we need a new optional property on AnalyticsChart so that we can properly set the export filename -- e.g., first try `props.exportFileNamePrefix`, then try `props.chartTitle`, then fall back to a default.

adorack (changes requested):

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
mihai-peteu: Good call. Pushed a commit.

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
```diff
@@ -201,7 +206,7 @@ const props = defineProps({
 
 const { i18n } = composables.useI18n()
 const heightRef = ref<string>(props.height)
-const csvFilename = computed<string>(() => props?.chartTitle ? props.chartTitle : i18n.t('csvExport.defaultFilename'))
+const csvFilename = computed<string>(() => props?.chartTitle ?? props?.filenamePrefix ?? i18n.t('csvExport.defaultFilename'))
```
adorack, line 209: Shouldn't it be `props.filenamePrefix ?? props.chartTitle || defaultFilename`?  It'd be weird if you set both `filenamePrefix` and `chartTitle` and the component ignored the filename prefix when determining the filename, and it'd be weird if an empty string chart title prevented the export from falling back to the default.

(I also don't think you need the question mark in `props?` -- that should always be defined, even if the exact prop is falsey.)

adorack (changes requested):
(Please don't forget about the comment on the test.)

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
mihai-peteu: Good point :)

mihai-peteu: Updated the order.
 
```
props.filenamePrefix ?? (props.chartTitle || i18n.t('csvExport.defaultFilename'))
```

adorack (changes requested):
LGTM aside from my earlier comment about the test in `AnalyticsChart.cy`.

packages/analytics/analytics-chart/src/components/AnalyticsChart.cy.ts
mihai-peteu: Copy / Pasta at its finest. Thanks for pointing out.

adamdehaven (approved):

PR #941, opened 2023-11-23
feat(dashboard-renderer): add grid layout components

filipgutica:
# Summary

Grid layout
https://konghq.atlassian.net/browse/MA-2268

Updated `DashboardRenderer` to use `GridLayout` for laying out the `Tiles` 

Updated `DashboardDefinition` and `TileDefinition` schemas to include positioning and sizing info that is passed off to the `GridLayout` component for display.

See the following example:
```vue
<DashboardRenderer
  :context="context"
  :config="config"
/>
```
```typescript
const context: DashboardRendererContext = {
  filters: [],
  timeSpec: {
    type: 'relative',
    time_range: '15M',
  },
}

const config: DashboardConfig = {
  // 5 x 5 grid
  gridSize: {
    cols: 5,
    rows: 5,
  }
  tiles: [
    {
      definition: {
        chart: {
          type: ChartTypes.HorizontalBar,
        },
        query: {},
      },
      layout: {
        // Position at column 0, row 0
        position: {
          col: 1,
          row: 1,
        },
        // Spans 3 columns and 2 rows
        size: {
          col: 3,
          row: 2,
        }
      }
    },
    {
      definition: {
        chart: {
          type: ChartTypes.Gauge,
          metricDisplay: ChartMetricDisplay.Full,
          reverseDataset: true,
          numerator: 0,
        },
        query: {},
      },
      layout: {
        // Position at column 0, row 2
        position: {
          col: 0,
          row: 2,
        },
        // Spans 3 columns and 2 rows
        size: {
          col: 3,
          row: 2,
        }
      }
    },
  ],
}
```
https://www.loom.com/share/d1860d63881049e8b1d5feeff8ed054a
image
image
image

## PR Checklist


adamdehaven:
I thought this was going to be a POC, not actually merged into the package?

packages/analytics/dashboard-renderer/src/components/layout/GridLayout.vue
```diff
@@ -0,0 +1,114 @@
...
+      // mark all spanned positions as occupied
+      if (tile) {
+        for (let i = 0; i < tile.size.rows; i++) {
+          for (let j = 0; j < tile.size.cols; j++) {
+            occupied.add(`row-${row + i}-col-${col + j}`)
+          }
+        }
+      }
+    }
+  }
```
adamdehaven, line 79: What's the performance here look like? You have two sets of nested loops within the a parent loop

packages/analytics/dashboard-renderer/src/components/layout/GridLayout.vue
```diff
@@ -0,0 +1,114 @@
...
+    min-width: v-bind('tileWidthPixels');
+  }
+
+  .tile-content {
+    height: 100%;
+    width: 100%;
+  }
+}
+
+@media (max-width: 750px) {
```
adamdehaven, line 109: suggestion: use the breakpoints from the design tokens package

packages/analytics/dashboard-renderer/src/components/layout/GridLayout.vue
```diff
@@ -0,0 +1,114 @@
+<template>
+  <div class="kong-ui-public-grid-layout">
+    <div
+      v-for="cell in totalCells"
+      :key="cell.key"
+      class="grid-cell"
+      :style="{
+        gridRow: `span ${cell.tile?.size.rows}`,
+        gridColumn: `span ${cell.tile?.size.cols}`,
```
adamdehaven, line 9: I'm not sure the optional chaining makes sense in the style attributes here; if they are undefined it will render invalid CSS

packages/analytics/dashboard-renderer/src/components/SimpleChartRenderer.vue
```diff
@@ -7,6 +7,7 @@
     <SimpleChart
       :chart-data="data"
       :chart-options="mappedChartOptions"
+      :height="`${props.height}px`"
```
adamdehaven, line 10: you don't use `props` in the template 
```suggestion
      :height="`${height}px`"
```

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
```diff
@@ -64,6 +64,10 @@ export type GaugeChartOptions = FromSchema<typeof gaugeChartSchema>
 export const tileSchema = {
   type: 'object',
   properties: {
+    id: {
+      type: 'string',
+      description: 'Unique identifier for the tile.',
```
adamdehaven, line 69: question: where do these descriptions get utilized?

packages/analytics/dashboard-renderer/src/types/grid-layout-types.ts
```diff
@@ -0,0 +1,20 @@
...
+export interface Tile {
+  id: string,
+  position: TilePosition,
+  size: TileSize,
+}
+
+export interface GridSize {
+  rows: number,
+  cols: number,
+}
```
adamdehaven, line 20: nit: interfaces don't need commas

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -6,21 +6,30 @@
...
-      v-for="(tile, index) in props.definition.tiles"
+    <GridLayout
       v-else
-      :key="index"
-      :definition="tile"
-    />
+      :grid-size="props.definition.gridSize"
+      :tile-height="props.definition.tileHeight"
+      :tile-width="props.definition.tileWidth"
+      :tiles="props.definition.tiles"
```
adamdehaven, line 14: You don't use `props` in the template, just reference the prop directly

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -6,21 +6,30 @@
...
+    >
+      <template #tile="{ tile }">
+        <Tile
+          :definition="(tile as TileDefinition)"
+          :height="(props.definition.tileHeight ?? 150) * tile.size.rows"
```
adamdehaven, line 19: Same comment here: don't reference props

packages/analytics/dashboard-renderer/src/components/layout/GridLayout.vue
```diff
@@ -0,0 +1,114 @@
...
+</template>
+
+<script lang="ts" setup>
+import { defineProps, computed, type PropType } from 'vue'
+import type { GridSize, Tile } from 'src/types'
+
+interface Cell {
+  key: string
+  tile?: Tile
+}
```
adamdehaven, line 29: interfaces should be defined in a types file

packages/analytics/dashboard-renderer/src/components/layout/GridLayout.vue
```diff
@@ -0,0 +1,114 @@
...
+const tileWidthPixels = computed(() => `${props.tileWidth}px`)
+const tileHeightPixels = computed(() => `${props.tileHeight}px`)
+
+</script>
+
+<style lang="scss" scoped>
+.kong-ui-public-grid-layout {
+  box-sizing: border-box;
+  display: grid;
+  gap: 10px;
```
adamdehaven, line 93: nit: can you our design tokens

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
```diff
@@ -99,4 +155,5 @@ export interface RendererProps<T> {
   query: any // TODO: Explore v4
   queryReady: boolean,
   chartOptions: T,
+  height: number,
```
adamdehaven, line 158: No commas in type files

packages/analytics/dashboard-renderer/src/components/layout/GridLayout.vue
filipgutica: I'm assuming we're not going to have massive grids with thousands of cells, with tiles spanning 100s of cells. 

We can put a hard limit on the size of grid we will have to prevent performance issues.

The tile rows/cols nested loop will only loop for how many cells that tile spans, which should never be more than a few col,row spans

filipgutica:
> I thought this was going to be a POC, not actually merged into the package?

I made quick enough progress on the layout that it doesn't need to be a POC, we need to use the layout for the feature. 

Eventually it can be replaced once there's a Kongponent that meets our needs here?

packages/analytics/dashboard-renderer/src/components/layout/GridLayout.vue
adamdehaven: I still have some concerns with the loops 😬

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
filipgutica: https://json-schema.org/draft/2020-12/json-schema-validation#name-title-and-description

filipgutica: I can remove them from here, but eventually this schema will be used to represent dashboards and tiles in the dashboard API. Just getting ahead of myself may be.

packages/analytics/dashboard-renderer/src/components/layout/GridLayout.vue
filipgutica: Ok, reduced the number of loops, now it's just one nested loop.

filipgutica:
> @filipgutica I think it would make sense to look for an existing library to generate the grid instead of creating something like this from scratch. I'm guessing one exists that also supports saving the layout, etc. like we chatted about

This component already does everything we need it to do, as, is in this PR. Can we not move forward with this?

packages/analytics/dashboard-renderer/src/components/layout/GridLayout.vue
```diff
@@ -0,0 +1,130 @@
...
+        'empty-cell': !cell.tile,
+      }"
+      :style="cell.tile && {
+        gridRow: `span ${cell.tile.size.rows}`,
+        gridColumn: `span ${cell.tile.size.cols}`,
+      }"
+    >
+      <slot
+        v-if="cell.tile"
+        class="tile-content"
```
adamdehaven, line 20: Classes aren't typically bound to the slot; is this rendering in the DOM?

packages/analytics/dashboard-renderer/src/components/layout/GridLayout.vue
```diff
@@ -0,0 +1,130 @@
...
+  },
+  tileHeight: {
+    type: Number,
+    required: false,
+    default: DEFAULT_TILE_HEIGHT,
+  },
+  tileWidth: {
+    type: Number,
+    required: false,
+    default: DEFAULT_TILE_WIDTH,
```
adamdehaven, line 46: issue: you can't bind imports to default props values; they have to be imported in a separate script block

packages/analytics/dashboard-renderer/src/components/layout/GridLayout.vue
```diff
@@ -0,0 +1,130 @@
...
+  },
+  tiles: {
+    type: Array as PropType<Tile[]>,
+    required: true,
+  },
+})
+
+const gridCols = ref(props.gridSize.cols)
+const gridRows = ref(props.gridSize.rows)
+const gridContainer = ref<HTMLElement>()
```
adamdehaven, line 56: issue: you have to initialize the ref as `null`, not `undefined`
```suggestion
const gridContainer = ref<HTMLElement>(null)
```

packages/analytics/dashboard-renderer/src/components/layout/GridLayout.vue
```diff
@@ -0,0 +1,130 @@
...
+const localTiles = reactive([...props.tiles])
+
+const totalCells = computed<Cell[]>(() => {
```
adamdehaven, line 59: nit: "total" implies a number but here it's used as an aggregate. Can you rename the prop to something more descriptive?

packages/analytics/dashboard-renderer/src/components/layout/GridLayout.vue
```diff
@@ -0,0 +1,130 @@
...
+      }
+    }
+
+  }
+
+  return cells
+})
+
+const tileWidthPixels = computed(() => `${props.tileWidth}px`)
+const tileHeightPixels = computed(() => `${props.tileHeight}px`)
```
adamdehaven, line 96: suggestion: there is a lot of logic and variables here; can you add comments?

packages/analytics/dashboard-renderer/src/components/layout/GridLayout.vue
```diff
@@ -0,0 +1,130 @@
...
+    min-width: v-bind('tileWidthPixels');
+  }
+
+  .tile-content {
+    height: 100%;
+    width: 100%;
+  }
+}
+
+@media (max-width: $kui-breakpoint-mobile) {
```
adamdehaven, line 120: issue: our breakpoints are mobile-first

You should typically write mobile styles first, and the. Progressively alter styles as the viewport width increases. 

If you **must** use max-width breakpoints, you need to adjust the value as shown here: 
```suggestion
@media (max-width: ($kui-breakpoint-mobile - 1px)) {
```

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <div class="analytics-chart-shell">
+  <div
+    class="analytics-chart-shell"
+    :style="{ height: heightRef, width }"
```
adamdehaven, line 4: question: why are you using inline styles instead of v-bind?

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -6,27 +6,58 @@
...
-      v-for="(tile, index) in props.definition.tiles"
+    <GridLayout
       v-else
-      :key="index"
-      :definition="tile"
-    />
+      :col-num="definition.gridSize.cols"
+      :layout="dashboardLayout"
+      :max-rows="definition.gridSize.rows"
+      :row-height="definition.tileHeight ?? DEFAULT_TILE_HEIGHT"
```
adamdehaven, line 14: question: This only accounts for null, what if it's undefined?

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -6,27 +6,58 @@
...
+import { GridLayout, GridItem } from 'vue3-grid-layout-next'
+import Tile from './DashboardTile.vue'
 
 const props = defineProps<{
   context: DashboardRendererContext,
   definition: DashboardDefinition,
 }>()
 
+const dashboardLayout = computed(() => {
+  return props.definition.tiles.map(tile => {
```
adamdehaven, line 49: suggestion: Can you add return types here

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -11,25 +11,40 @@
...
+  },
+  height: {
+    type: Number,
+    required: false,
+    default: DEFAULT_TILE_HEIGHT,
+  },
+  width: {
+    type: Number,
+    required: false,
+    default: DEFAULT_TILE_WIDTH,
```
adamdehaven, line 38: issue: same comment here on referencing imported const as default prop values

packages/analytics/dashboard-renderer/package.json
```diff
@@ -64,14 +64,15 @@
     "errorLimit": "200KB"
   },
   "peerDependencies": {
-    "@kong-ui-public/analytics-chart": "^0.15.3",
     "@kong-ui-public/i18n": "workspace:^",
     "@kong/kongponents": "^8.126.2",
     "vue": "^3.3.4"
   },
   "dependencies": {
+    "@kong-ui-public/analytics-chart": "^0.16.5",
```
adamdehaven, line 72: issue: shouldn't this point to the workspace?

adorack (changes requested):
I'd like to review.

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
```diff
@@ -36,6 +38,24 @@ export const barChartSchema = {
...
+    stacked: {
+      type: 'boolean',
+    },
+    syntheticsDataKey,
+  },
+  required: ['type'],
+  additionalProperties: false,
+} as const satisfies JSONSchema
+
+export type TimeseriesChartOptions = FromSchema<typeof barChartSchema>
```
adorack, line 57:
```suggestion
export type TimeseriesChartOptions = FromSchema<typeof timeseriesChartSchema>
```

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
```diff
@@ -61,42 +81,116 @@ export const gaugeChartSchema = {
...
+export type TileLayout = FromSchema<typeof tileLayoutSchema>
+
+export const tileConfigSchema = {
+  type: 'object',
+  properties: {
+    id: {
       type: 'string',
     },
+    definition: {
+      allOf: [tileDefinitionSchema],
```
adorack, line 146: You can probably simplify this by just embedding the `tileDefinitionSchema` directly: `definition: tileDefinitionSchema`.

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
```diff
@@ -61,42 +81,116 @@ export const gaugeChartSchema = {
...
+    },
+    layout: {
+      allOf: [tileLayoutSchema],
```
adorack, line 149: Same comment as above.

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
```diff
@@ -61,42 +81,116 @@ export const gaugeChartSchema = {
...
+  required: ['position', 'size'],
+  additionalProperties: false,
+} as const satisfies JSONSchema
+
+export type TileLayout = FromSchema<typeof tileLayoutSchema>
+
+export const tileConfigSchema = {
+  type: 'object',
+  properties: {
+    id: {
```
adorack, line 142: We had talked about how we should autogenerate the ID rather than requiring it, but I forgot to post my review comment about that earlier -- sorry about that.  Since autogenerating the ID is probably a more substantial change, I'm fine doing it in a followup -- would you mind creating a ticket?

adorack (changes requested):

packages/analytics/dashboard-renderer/src/components/AnalyticsChartRenderer.vue
```diff
@@ -24,6 +26,7 @@ const props = defineProps<RendererProps<BarChartOptions>>()
 const chartTypeLookup = {
```
adorack, line 26: I can't comment directly on the line (yay Github), but I think on line 24 the props should be `RendererProps<BarChartOptions | TimeseriesChartOptions>`.  If this causes any trouble, then don't worry about fixing it, but please let me know so I can try to fix it later.

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
filipgutica: This doesn't work, FromSchema fails, results in `TileConfig` being type `never` 

image

filipgutica: never mind

PR #991, opened 2023-12-08
feat(update-analytics-types): Update types for null-positions, and grouped display

hfukada:
- allow for null positions
- flag in display blob to denote the "everything in this label is the /rest/ of the grouped data" e.g. for consumers. 50(consumer positions)+1(other position),


filipgutica:
Is there a mock up or visual component regarding how this "other" grouping is supposed to show up/look like in the chart? 
I'm not sure if the changes in this PR will actually result in a dataset that shows up as "other" if that's the intent here?

packages/analytics/analytics-chart/src/composables/useExploreResultToDatasets.spec.ts
```diff
@@ -26,6 +26,46 @@ describe('useVitalsExploreDatasets', () => {
...
+    }))
+
+    const result = useExploreResultToDatasets({ fill: true }, exploreResult)
+
+    expect(result.value).toEqual({
+      datasets: [
+        {
+          backgroundColor: '#a86cd5',
+          data: [
+            null,
```
filipgutica, line 58: I think ChartJS will ignore this.

filipgutica (changes requested):

packages/analytics/analytics-chart/src/composables/useExploreResultToDatasets.spec.ts
```diff
@@ -26,6 +26,46 @@ describe('useVitalsExploreDatasets', () => {
...
+        timestamp: '2022-01-01T01:01:02Z',
+        event: {
+          dimension: null,
+        },
+      }],
+      meta: {
+        start: 1640998862,
+        end: 1640998870,
+        granularity: 5000,
+        dimensions: { dimension: ['dimension'] },
```
filipgutica, line 42: The array maps to the actual values of that particular dimension that appears in the records so this should be 
`dimensions: { dimension: [null] },`

If there was another record with event 
```
       event: {
          dimension: 'service1',
        }
``` 
For example, then the meta would be 
`dimensions: { dimension: [null, 'service1'] },`

filipgutica (approved):

PR #1000, opened 2023-12-12
feat: improve chart debugging experience

filipgutica:
# Summary

Add import chart data section to the sandbox

#### Resources


mihai-peteu (approved):

packages/analytics/analytics-chart/src/components/DebugModal.vue
```diff
@@ -0,0 +1,139 @@
...
+    hash: '#chart-js-data',
+    title: 'ChartJS Data',
+  },
+  {
+    hash: '#chart-options',
+    title: 'Chart Options',
+  },
+]
+
+const secretCode = ['ArrowUp', 'ArrowUp', 'ArrowDown', 'ArrowDown', 'ArrowLeft', 'ArrowRight', 'ArrowLeft', 'ArrowRight', 'KeyB', 'KeyA']
```
adamdehaven, line 97: What's the reasoning for shipping a `DebugModal` along with the component? Is this something we want to expose to customers? 

---

Separately, instead of keybindings, could we just look for a certain localStorage key? 

e.g. `localStorage.getItem(analytics-chart-debug, 'true')`?

packages/analytics/analytics-chart/src/components/DebugModal.vue
```diff
@@ -0,0 +1,139 @@
...
+  }, 20000)
+}
+
+onMounted(() => {
+  window.addEventListener('keydown', checkKonamiCode)
+})
+
+onUnmounted(() => {
+  window.removeEventListener('keydown', checkKonamiCode)
+})
```
adamdehaven, line 122: issue: You're binding to every keyboard event on the window -- this isn't ideal.

These should at least be scoped to if the component has focus, etc.

adamdehaven (changes requested):

packages/analytics/analytics-chart/src/components/DebugModal.vue
filipgutica: Its just a fun way to get the chart data out of a chart for a developer or power user. I thought using the code would be a fun way to hide it from customers.

packages/analytics/analytics-chart/src/components/DebugModal.vue
filipgutica: Yeah, true, I'll update that

packages/analytics/analytics-chart/src/components/DebugModal.vue
adamdehaven: I'm not sure it's ideal to ship this into something that can be exposed in Production. Is that ideal? 

IMO we shouldn't be tying more keybindings; if I was working on another component in the Konnect UI and this conflicted, I'd have no way to know where it was coming from since it's not documented anywhere.

adamdehaven: @mihai-peteu why did you mark this as resolved?

ValeryG: Feels like easter egg to me... I know fun stuff is fun, but...  
- Is this functionality something that is requested by PM?  
- Is there an agreement to document it and make it a part of product offering? 
- How often and by whom (other that author) will it be used? 

- If this is part of the functionality exposed to customers and documented - it should be easier accessible than trough the 10 key sequence, if this is for devs only ... hmm, I am confused as why it is needed .

mihai-peteu: It's just a useful debugging tool, something that a power user or a developer would appreciate

adamdehaven: > something that a power user or a developer would appreciate

How would they ever know about it...? I'm not sure this belongs in Konnect?

adamdehaven: Also, what happens when there is more than one chart on the page?

ValeryG: want to repeat my question: is this work done for fun or is it something product / pm requested?

adamdehaven: This seems like a component that could live and only be used in the [published sandbox](https://kongstrong.slack.com/archives/C01RF6STU2F/p1700171044487639) - wasn't that the point of us publishing the sandbox to Amplify?

filipgutica: Just as an example. And this is something I've done many times to debug weird issues and edge cases that might come upt. 

But I imagine a use case where there's a weird issue with a particular chart in Konnect, you can use this to easily extract the chart data, then take that to the sandbox, past it in, in order to render that particular chart you saw in Konnect and start debugging. 

I can remove the debugger stuff for now until we run it by Christian first?

adamdehaven: > But I imagine a use case where there's a weird issue with a particular chart in Konnect, you can use this to easily extract the chart data, then take that to the sandbox, past it in, in order to render that particular chart you saw in Konnect and start debugging.

If there's an issue with a chart for a customer, wouldn't this data be sourced from Datadog, or by impersonating the user and inspecting what is going on with the request/response and how the UI is behaving?

IMO debugging tools have no place in the Production build/deployment.

packages/analytics/analytics-chart/sandbox/utils/utils.ts
```diff
@@ -14,3 +16,25 @@ export function rand(min: number, max: number): number {
...
+export function isValidJson(str: string): boolean {
+  if (!str) return true
+
+  try {
+    JSON.parse(str)
+  } catch (e) {
+    return false
+  }
+
+  return true
```
adamdehaven, line 29: nit: Throw this bad boy in the try block

packages/analytics/analytics-chart/sandbox/utils/utils.ts
```diff
@@ -14,3 +16,25 @@ export function rand(min: number, max: number): number {
...
+}
+
+export function parseJson(exploreResultText: AnalyticsExploreV2Result) {
+  try {
+    const result = JSON.parse(exploreResultText as any)
+
+    return result as AnalyticsExploreV2Result
```
adamdehaven, line 36: not sure the typings here make sense? 🤔

MBasani22 (approved):

PR #1007, opened 2023-12-14
feat: expose a custom sort function for datasets as they appear in tooltip and legend

filipgutica:
# Summary

Needed for https://konghq.atlassian.net/browse/MA-2390 

- `other` should appear at the end of the list regardless of raw value

#### Resources


mihai-peteu (approved):
We might want to add a `ChartTooltip.cy.ts` in the near future, created a ticket for that

https://konghq.atlassian.net/browse/MA-2415

PR #967, opened 2023-11-30
feat: add ICU datetime date format

johncowen:
# Summary

Adds an ICU message format `datetime` format exactly the same as `formatIsoDate` for usage from within ICU files, making it easier to use a standard format within longer form texts such as:

```
I want a present {value, select, 
  0 { for Christmas! }
  undefined { for Christmas! }
  other { on { value, date, datetime } } 
}
```

## PR Checklist

* [x] **Naming & Structure:** the files and package structure use the conventions outlined in the [Creating a Package docs](https://github.com/Kong/public-ui-components/blob/main/docs/creating-a-package.md).
* [x] **Tests pass:** check the output of all package unit and/or component tests.
  * [ ] If this PR is the result of a bug, test coverage was added accordingly.
* [x] **Functional:** all changes do not break existing APIs, but if so, a BREAKING CHANGE commit is in place to bump the major version.
* [x] **Conventional Commits** all commits follow the conventional commit standards [outlined in the main README](https://github.com/Kong/public-ui-components#committing-changes).
* [ ] **Docs:** includes a technically accurate README, and the docs have been updated accordingly based on the changes in this PR.

PR #1012, opened 2023-12-18
feat: add ICU datetime date format

johncowen:
# Summary

Adds an ICU message format `datetime` format exactly the same as `formatIsoDate` for usage from within ICU files, making it easier to use a standard format within longer form texts such as:

```
I want a present {value, select, 
  0 { for Christmas! }
  undefined { for Christmas! }
  other { on { value, date, datetime } } 
}
```

## PR Checklist

Replaces https://github.com/Kong/public-ui-components/pull/967 along with additional README edits.

### DIFF
--- a/packages/core/i18n/README.md
+++ b/packages/core/i18n/README.md
@@ -403,6 +403,8 @@ This plugin includes an additional `datetime` formatter (examples above) which u
 
 - `datetime` format the date as date and time using the same format as `formatIsoDate` e.g. `Sale begins {start, date, datetime}`
 
+In the example `Sale begins {start, date, datetime}`, `start` is the value passed into `t` i.e. `t('key.of.string', {start: new Date().toString()})`, `date` is the name of the ICU format we are using, and `datetime` is to tell FormatJS to use the format we have defined as a standard/common format for formatting 'datetimes'.
+
 
 Unit tests for I18n wrapper in kong-ui/core also has view examples as a references for those.
 [FormatDate](https://github.com/Kong/shared-ui-components/blob/4a1f99d5cee2d4409f4370a9bce2377450a6429d/packages/core/src/useI18n/i18n.spec.ts#L81)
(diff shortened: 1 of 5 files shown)


packages/core/i18n/README.md
```diff
@@ -204,6 +209,11 @@ With pluralization:
   0 versions available
   1 version available
   11 version available
+
+With datetime formatting:
+  Nov 30, 2023, 11:40 AM0
```
adamdehaven, line 214: Typo here?
```suggestion
  Nov 30, 2023, 11:40 AM
```

johncowen: Whoops! Thanks!

jakubdyszkiewicz (dismissed):

packages/core/i18n/README.md
```diff
@@ -165,6 +165,7 @@ const i18n = createI18n<typeof english>(props.locale || 'en-us', english)
   "test": {
     "withParams": "Good {dayPart}, My name is {name}!",
     "withPluralization": "{count} {count, plural, =0 {versions} =1 {version} other {versions}} available"
+    "withDateTimeFormatting": "{value, date, datetime}",
```
ValeryG, line 168: Looks good, the last thing is: would be great if you can add a short description on what each of those params in {} is: like 
`value`: - actual value of the date to be present
`date`: flag to indicate that value needs to be treated as a date 
`datetime` - ...???

ValeryG (changes requested):

packages/core/i18n/README.md
ValeryG: > "withDateTimeFormatting": "{value, date, datetime}",

reading this line in en.json I am still not sure what each of those means. can you add a clarification there.  all the examples that use this key, only pass value, so what are 2-nd and 3-d arguments for?

ValeryG (dismissed):
just one README comment and let's merge this thing ...

packages/core/i18n/README.md
johncowen: I added https://github.com/Kong/public-ui-components/pull/1012/commits/025f0b29688223ea9c9308a65390af683eca14f2 down in the `## Formatting numbers, dates and times` section, lmk if that helps.

johncowen:
Added clarification in https://github.com/Kong/public-ui-components/pull/1012/commits/025f0b29688223ea9c9308a65390af683eca14f2 (see also inline convo above) lmk

ValeryG (approved):

PR #1017, opened 2023-12-19
fix(plugins): card component is not KCard [khcp-9970]

kaiarrowood:
# Summary

Switch the plugin card component away from using `KCard` since design does not match.
Addresses [KHCP-9970](https://konghq.atlassian.net/browse/KHCP-9970).

#### Resources

[KHCP-9970]: https://konghq.atlassian.net/browse/KHCP-9970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-plugins/sandbox/pages/PluginSelectPage.vue
+++ b/packages/entities/entities-plugins/sandbox/pages/PluginSelectPage.vue
@@ -3,6 +3,7 @@
     <h2>Konnect API</h2>
     <PluginSelect
       :config="konnectConfig"
+      :disabled-plugins="{ 'acl': 'ACL is not supported for this entity type'}"
       @delete-custom:success="handleDeleteSuccess"
     />
 
--- a/packages/entities/entities-plugins/src/components/PluginIcon.vue
+++ b/packages/entities/entities-plugins/src/components/PluginIcon.vue
@@ -10,7 +10,7 @@
 </template>
 
 <script setup lang="ts">
-import { ref, computed } from 'vue'
+import { ref } from 'vue'
 import composables from '../composables'
 
 const props = defineProps({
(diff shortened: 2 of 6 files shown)


packages/entities/entities-plugins/sandbox/pages/PluginSelectPage.vue
```diff
@@ -3,6 +3,7 @@
     <h2>Konnect API</h2>
     <PluginSelect
       :config="konnectConfig"
+      :disabled-plugins="{ 'acl': 'ACL is not supported for this entity type'}"
```
adamdehaven, line 6: question: seems unrelated?

packages/entities/entities-plugins/src/components/PluginIcon.vue
```diff
@@ -31,11 +30,27 @@ defineProps({
   },
 })
 
+const defaultIcon = ref('')
+const src = ref('')
```
adamdehaven, line 34: suggestion: can we rename this different than the native attribute? `imgSrc` or something?

packages/entities/entities-plugins/src/components/PluginIcon.vue
```diff
@@ -31,11 +30,27 @@ defineProps({
...
 const img = ref<HTMLImageElement>()
```
adamdehaven, line 35: img ref should be initialized as `null`

packages/entities/entities-plugins/src/components/PluginIcon.vue
```diff
@@ -31,11 +30,27 @@ defineProps({
...
+    defaultIcon.value = new URL('../assets/images/plugin-icons/missing.png', import.meta.url).href
+  } catch (err) {
+    defaultIcon.value = ''
+  }
+
+  try {
+    src.value = composables.getPluginIconURL(props.name)
+  } catch (err) {
+    src.value = defaultIcon.value
+  }
```
adamdehaven, line 54: question: why isn't the try/catch in the composable where the import is actually happening?

packages/entities/entities-plugins/src/components/select/PluginSelectCard.vue
```diff
@@ -3,66 +3,78 @@
     :label="plugin.disabledMessage"
     position-fixed
   >
-    <KCard
-      class="plugin-card-content"
-      :class="{ disabled: isDisabled }"
+    <div
```
adamdehaven, line 6: question: why are we using a `div` as a `role="button"`? Can this be an `a` tag instead?

packages/entities/entities-plugins/sandbox/pages/PluginSelectPage.vue
kaiarrowood: Using it to make sure I didn't break display of disabled plugins in the sandbox

packages/entities/entities-plugins/src/components/select/PluginSelectCard.vue
kaiarrowood: I can switch it, @portikM had recommended `div` with `role="button"` in the associated ticket.

adamdehaven: I'd recommend an `a` anchor tag since they are natively focusable and clickable. Leave the `role="button"` unless we're actually navigating (e.g. if we're just opening a modal, or "selecting" something)

PR #1023, opened 2023-12-20
fix(plugins): icon rendering

kaiarrowood:
# Summary
Attempt to fix issue where we attempt to render a non-existing icon and it fills up the console with warnings.

#### Resources

### DIFF
--- a/packages/entities/entities-plugins/src/composables/usePluginMeta.ts
+++ b/packages/entities/entities-plugins/src/composables/usePluginMeta.ts
@@ -18,7 +18,22 @@ import KeyAuthEncSchema from './plugin-schemas/KeyAuthEnc'
 import keyEncCredentialSchema from './plugin-schemas/credentials/mockedKeyEncAuthSchema.json'
 
 export const getPluginIconURL = (pluginName: string) => {
-  return new URL(`../assets/images/plugin-icons/${pluginName}.png`, import.meta.url).href
+  let defaultIcon = ''
+  let imgSrc = ''
+
+  try {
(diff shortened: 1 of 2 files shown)


packages/entities/entities-plugins/src/composables/usePluginMeta.ts
```diff
@@ -18,7 +18,22 @@ import KeyAuthEncSchema from './plugin-schemas/KeyAuthEnc'
...
+    defaultIcon = ''
+  }
+
+  try {
+    imgSrc = new URL(`../assets/images/plugin-icons/${pluginName}.png`, import.meta.url).href
+  } catch (err) {
+    imgSrc = defaultIcon
+  }
+
+  return imgSrc
```
adamdehaven, line 36: The issue here is that even a string of `undefined` will result in an invalid asset request.

You may want to try refactoring to something like this:
```suggestion
  let imgSrc = ''
  
  // If no `pluginName`, exit early to prevent `undefined` in the asset path
  if (!pluginName) {
    return imgSrc
  }
  
  let defaultIcon = ''
  
  try {
    defaultIcon = new URL('../assets/images/plugin-icons/missing.png', import.meta.url).href
    
    if (defaultIcon.includes('undefined')) {
      throw new Error('invalid path')
    }
  } catch (err) {
    defaultIcon = ''
  }

  try {
    imgSrc = new URL(`../assets/images/plugin-icons/${pluginName}.png`, import.meta.url).href
    
    if (imgSrc.includes('undefined')) {
      throw new Error('invalid path')
    }
  } catch (err) {
    imgSrc = defaultIcon
  }

  return imgSrc
```

PR #1028, opened 2023-12-21
fix(swagger-ui-web-component): operationId (#987)

andrewwylde:
### Note: this is to backport to the 0.x.x version of `swagger-ui-web-component` and eventually `spec-renderer`.

Fixes an issue where the escaping pattern is not matched with how swagger-ui creates the id


adamdehaven (approved):
🤞🏼

PR #1029, opened 2023-12-21
feat(entities-shared): support pagination for new filtering [KHCP-10099]

LukeSwierlik:
# Summary
- Feat - Adding support for pagination when using the new version of filtering.
- Task in Jira: https://konghq.atlassian.net/browse/KHCP-10099

Screenshot 2023-12-21 at 21 10 23
Screenshot 2023-12-21 at 21 10 31

#### Resources


sumimakito (approved):
These changes look good to me and it should work fine with Kong Manager as we are not using the `filter[name]'` parameter. It would be better if we parse and reconstruct the URL with libraries in my personal opinion, but please feel free to merge the PR. 👍

PR #1040, opened 2023-12-28
fix(entities-gateway-services): fix broken radio label

nekolab:
Nested label will break the click reactivity. Since the radio font weight is default to 600 now, there is no need to wrap with nested `<label />` and set the `font-weight` explicitly.

# Summary

#### Resources


Leopoldthecoder (approved):
LGTM

PR #1046, opened 2024-01-03
fix(entities-plugins): remove click handler from submit button

Leopoldthecoder:
# Summary

The submit button already has `saveFormData` as its click handler: https://github.com/Kong/public-ui-components/blob/76d87a0a2f176b0687f90b4511f210db2115feee/packages/entities/entities-plugins/src/components/PluginForm.vue#L83
So the form doesn't need a submit handler. Otherwise, in some circumstances, `saveFormData` will run twice when the submit button is clicked, e.g. https://github.com/Kong/kong-test-automation/actions/runs/7393902026/job/20114469588.

#### Resources


nekolab (approved):
LGTM

PR #1072, opened 2024-01-12
fix(analytics-chart): sizing and overflow issues plus vertical scroll

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-2508

implement vertical scroll plus get rid of width and height props. Host app will have to wrap in a container that sets dimensions accordingly

BREAKING CHANGE:
- No more hight and width props. Host app must wrap in a container that explicitely sets width and height.
- There is vertical scroll now in bar charts verify in host apps where applicable.

#### Resources


adorack (changes requested):
I'd like to discuss this further on Tuesday before we implement it.

packages/analytics/analytics-chart/src/components/chart-types/StackedBarChart.vue
```diff
@@ -359,18 +360,22 @@ const chartWidth = computed(() => {
 })
 
 const chartHeight = computed(() => {
-  let chartHeight = parseInt(props.height, 10)
-  if (numLabels.value && isHorizontal.value) {
+  let value: number | string = DEFAULT_CHART_HEIGHT
```
adamdehaven, line 363: suggestion: can you add a return type to the computed function, and add some comments as to what is going on within this computed variable?

packages/analytics/analytics-chart/src/components/chart-types/StackedBarChart.vue
```diff
@@ -340,7 +341,7 @@ const numLabels = computed(() => {
 })
 
 const chartWidth = computed(() => {
-  let value: number | string = MIN_CHART_WIDTH
+  let value: number | string = DEFAULT_CHART_WIDTH
```
adamdehaven, line 344: question: how can a const be a number or a string?

packages/analytics/analytics-chart/src/components/chart-types/StackedBarChart.vue
```diff
@@ -359,18 +360,22 @@ const chartWidth = computed(() => {
 })
 
 const chartHeight = computed(() => {
-  let chartHeight = parseInt(props.height, 10)
-  if (numLabels.value && isHorizontal.value) {
+  let value: number | string = DEFAULT_CHART_HEIGHT
+
+  if (canvas.value) {
+    value = canvas.value.width
```
adamdehaven, line 366: suggestion: we already have `{variable}.value` everywhere in Vue; can you rename this `value` variable as something else?

packages/analytics/analytics-chart/src/components/chart-types/StackedBarChart.vue
```diff
@@ -492,14 +493,14 @@ const axisDimensions = computed(() => {
 })
 
 const onScrolling = (event: Event) => {
-  if (axisDimensions.value && chartInstance.value) {
+  const target: HTMLElement = event.target as HTMLElement
```
adamdehaven, line 496: nit: you don't need the type if you're going to cast the `event.target`

packages/analytics/analytics-chart/src/components/chart-types/StackedBarChart.vue
```diff
@@ -563,6 +565,7 @@ const handleChartClick = () => {
   }
 }
 
+const redrawKey = computed(() => `${props.orientation}`)
```
adamdehaven, line 568: nit: add a return type to the computed

packages/analytics/analytics-chart/src/components/chart-types/StackedBarChart.vue
```diff
@@ -571,7 +574,7 @@ const handleChartClick = () => {
 
 .chart-container {
   -ms-overflow-style: thin;  /* IE and Edge */
-  overflow-y: hidden;
+  overflow-y: scroll;
```
adamdehaven, line 577: question: should this be `auto` instead of scroll?

packages/analytics/dashboard-renderer/src/components/BarChartRenderer.vue
```diff
@@ -4,15 +4,17 @@
...
-    <AnalyticsChart
-      :chart-data="data"
-      :chart-options="options"
-      chart-title=""
-      :height="`${height}px`"
-      legend-position="bottom"
-      :show-annotations="chartOptions.showAnnotations"
-      tooltip-title=""
-    />
+    <div :style="{ height: `${height}px`}">
```
adamdehaven, line 7: Instead of adding inline styles, you can use `v-bind` down in the style block https://vuejs.org/api/sfc-css-features.html#v-bind-in-css

packages/analytics/dashboard-renderer/src/components/TimeseriesChartRenderer.vue
```diff
@@ -4,14 +4,16 @@
...
   >
-    <AnalyticsChart
-      :chart-data="data"
-      :chart-options="options"
-      chart-title=""
-      :height="`${height}px`"
-      legend-position="bottom"
-      tooltip-title=""
-    />
+    <div :style="{ height: `${height}px`}">
```
adamdehaven, line 7: same comment here on inline styles

packages/analytics/analytics-chart/src/components/chart-types/StackedBarChart.vue
filipgutica: done

packages/analytics/analytics-chart/src/components/chart-types/StackedBarChart.vue
```diff
@@ -139,24 +132,31 @@ const props = defineProps({
...
 const MAX_BARS_VERTICAL = 30
+const MAX_BARS_HORIZONTAL = 35
 const MIN_BAR_WIDTH = 40
 const SCROLL_MIN = 0
 const SCROLL_MAX = 10
 const AXIS_BOTTOM_OFFSET = 10
 const AXIS_RIGHT_PADDING = 1
-const BAR_MARGIN = 6
+
+const chartContainerRef = ref<HTMLDivElement>()
```
adamdehaven, line 151: issue: template ref needs to be initialized as null 
```suggestion
const chartContainerRef = ref<HTMLDivElement | null>(null)
```

packages/analytics/analytics-chart/src/components/chart-types/StackedBarChart.vue
```diff
@@ -139,24 +132,31 @@ const props = defineProps({
...
+const initialContainerHeight = ref(0)
```
adamdehaven, line 152: nit
```suggestion
const initialContainerHeight = ref<number>(0)
```

packages/analytics/analytics-chart/src/components/chart-types/StackedBarChart.vue
```diff
@@ -563,6 +553,7 @@ const handleChartClick = () => {
   }
 }
 
+const redrawKey = computed<string>(() => `${props.orientation}`)
```
adamdehaven: nit: simplify
```suggestion
const redrawKey = computed<string>(() => props.orientation)
```

packages/analytics/dashboard-renderer/src/components/BarChartRenderer.vue
```diff
@@ -4,15 +4,16 @@
...
-      tooltip-title=""
-    />
+    <div class="analytics-chart">
+      <AnalyticsChart
+        :chart-data="data"
+        :chart-options="options"
+        chart-title=""
+        legend-position="bottom"
+        :show-annotations="chartOptions.showAnnotations"
+        tooltip-title=""
```
adamdehaven, line 14: question: what is the purpose of passing in the title attributes as empty strings? I would imagine that's the default?

packages/analytics/dashboard-renderer/src/components/TimeseriesChartRenderer.vue
```diff
@@ -36,3 +38,9 @@ const options = computed<AnalyticsChartOptions>(() => ({
   chartDatasetColors: props.chartOptions.chartDatasetColors,
 }))
 </script>
+
+<style scoped lang="scss">
+.analytics-chart {
+  height: v-bind('`${height}px`')
```
adamdehaven, line 44: question: did you verify this is outputting the desired value in the browser dev tools? (for all the computed CSS vals)

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
```diff
@@ -368,11 +342,11 @@ provide('legendPosition', toRef(props, 'legendPosition'))
...
 .analytics-chart-shell {
-  height: v-bind('heightRef');
-  width: v-bind('props.width');
+  height: 100%;
+  width: 100%;
   .analytics-chart-parent {
-    height: 100%;
-    width: 100%;
+    height: inherit;
+    width: inherit;
```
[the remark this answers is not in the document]
filipgutica: If I get rid of inherit dimensions here, the dimensions are messed up

image

packages/analytics/dashboard-renderer/src/components/TimeseriesChartRenderer.vue
filipgutica: yes its always coming in as a number, the dashboard renderer enforces setting a width and a height for each tile

filipgutica: in this case height is all we need cause width is always gonna be 100% by default

packages/analytics/analytics-chart/src/components/chart-types/StackedBarChart.vue
```diff
@@ -339,38 +332,30 @@ const numLabels = computed(() => {
...
 })
 
 const chartHeight = computed(() => {
-  let chartHeight = parseInt(props.height, 10)
-  if (numLabels.value && isHorizontal.value) {
+  if (canvas.value && props.chartData?.labels && props.chartData?.labels.length > MAX_BARS_HORIZONTAL && isHorizontal.value) {
+    const numLabels = props.chartData.labels.length
+
+    const baseHeight = canvas.value.offsetHeight
+    const preferredChartHeight = baseHeight + ((numLabels - MAX_BARS_HORIZONTAL) * MIN_BAR_WIDTH)
```
adorack, line 353: Why is this calculation changing?  Is it possible to implement this without changing the calculation?

In the past, we ran into issues with trying to do `baseHeight + <dynamic height>` because it's harder to calculate the dynamic part of the height -- for example, here the overall height may end up negative if there is not much canvas and only 1 bar.  That's why the old calculation relied on `Math.max`.  If we change it, we're going to need to do a lot more testing to make sure it behaves properly with varying numbers of bars.

packages/analytics/dashboard-renderer/src/components/TimeseriesChartRenderer.vue
```diff
@@ -4,14 +4,14 @@
...
-      chart-title=""
-      :height="`${height}px`"
-      legend-position="bottom"
-      tooltip-title=""
-    />
+    <div class="analytics-chart">
+      <AnalyticsChart
+        :chart-data="data"
+        :chart-options="options"
+        :height="`${height}px`"
```
adorack, line 11: This property should go away.

adorack (changes requested):

packages/analytics/analytics-chart/src/components/chart-types/StackedBarChart.vue
filipgutica: It can't be negative cause this code only gets execuded if numLabels is greater than `MAX_BARS_HORIZONTAL`

filipgutica: Also, just making it more consistent with the width calculation

filipgutica: nvm, moving to the old calculation for both the width and height computed props

packages/analytics/analytics-chart/src/components/chart-types/StackedBarChart.vue
```diff
@@ -402,7 +414,7 @@ const options = computed<ChartOptions>(() => {
 
 const chartInstance = composables.useChartJSCommon(
   'bar',
-  canvas,
+  canvas as Ref<HTMLCanvasElement>,
```
adorack, line 417: Is this cast because `canvas.value` can be null?  I think either the signature of `useChartJSCommon` should be adjusted (to reflect that it can handle nulls), or `chartInstance` should only get set if it's non-null.

adorack (approved):

PR #1080, opened 2024-01-16
fix(*): fix preprocessing form fields for gateway service form [khcp-10152]

mptap:
# Summary
https://konghq.atlassian.net/browse/KHCP-10152
Verified unique name constraint applies as expected in Gateway Service Form
Screenshot 2024-01-17 at 12 30 44 PM


packages/entities/entities-gateway-services/src/components/GatewayServiceForm.vue
```diff
@@ -714,7 +714,7 @@ const saveTlsVerify = (gatewayService: Record<string, any>) => {
   delete gatewayService.tls_verify_value
 }
 
-const getPayload = computed((): Record<string, any> => {
+const getPayload = (): Record<string, any> => {
```
adamdehaven, line 717: This function is `computed` in every other instance in the repo - why would it be changed only here?

Also, it is exposed via `defineExpose` in this component - does the usage throughout the repo where it is used need to be updated?

adamdehaven (dismissed):
blocking merge for discussion

packages/entities/entities-gateway-services/src/components/GatewayServiceForm.vue
mptap: It was a static object earlier so I'm basically reverting my change.

adamdehaven: Only in this one instance?

PR #1088, opened 2024-01-18
fix(gatewayServicesForm): not use computed in gateway service form [khcp-10152]

mptap:
# Summary
https://konghq.atlassian.net/browse/KHCP-10152
Fixes bug `GatewayServiceForm` where json/yaml blocks are empty in https://github.com/Kong/public-ui-components/pull/1080

Verified unique name constraint is still enforced
Screenshot 2024-01-18 at 10 22 51 AM

Deploy Preview https://pr-7419--khcp-ui.cloud-preview.konghq.tech/

### DIFF
--- a/packages/entities/entities-gateway-services/src/components/GatewayServiceForm.cy.ts
+++ b/packages/entities/entities-gateway-services/src/components/GatewayServiceForm.cy.ts
@@ -297,6 +297,25 @@ describe('<GatewayServiceForm />', { viewportHeight: 800, viewportWidth: 700 }, 
       cy.getTestId('gateway-service-host-input').should('have.value', gatewayService1.host)
     })
 
+    it('should correctly pass getPayload as props to json/yaml code blocks', () => {
+      interceptKonnect()
+
+      cy.mount(GatewayServiceForm, {
+        props: {
+          config: baseConfigKonnect,
+          gatewayServiceId: gatewayService1.id,
+          isEditing: true,
+        },
+      })
+
+      cy.wait('@getGatewayService')
+      cy.get('.kong-ui-entities-gateway-service-form').should('be.visible')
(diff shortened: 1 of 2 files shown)


jillztom:
Why are we going back and forth on this? Is there a concern when using computed or can we leave it as it is? I'm trying to understand the context here.

mptap:
> Why are we going back and forth on this? Is there a concern when using computed or can we leave it as it is? I'm trying to understand the context here.

@jillztom I demoed this in today's standup PL as using `computed` causes a bug where the json/yaml blocks are empty only in `GatewayServiceForm` (attached screenshot below). Only for `GatewayServiceForm` I want to use a regular object so everything works as expected in the form and in the json/yaml blocks. I can demo on Zoom if that's easier.

Screenshot 2024-01-18 at 10 25 34 AM

ValeryG (changes requested):
this looks like a fix - if fix changes behavior - we might need some sort of test coverage to: - make sure new behavior works, but most impotently. to make sure it is not broken in the future. any change to add tests for this change?

jillztom:
> > Why are we going back and forth on this? Is there a concern when using computed or can we leave it as it is? I'm trying to understand the context here.
> 
> @jillztom I demoed this in today's standup PL as using `computed` causes a bug where the json/yaml blocks are empty only in `GatewayServiceForm` (attached screenshot below). Only for `GatewayServiceForm` I want to use a regular object so everything works as expected in the form and in the json/yaml blocks. I can demo on Zoom if that's easier.
> 
> Screenshot 2024-01-18 at 10 25 34 AM

The issue is, you are passing `:form-fields="getPayload.value"` instead of `:form-fields="getPayload"`. Remember, it's a computed property.

jillztom (approved):

PR #1089, opened 2024-01-18
fix(analytics-chart): style tweaks

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-2457

Issue: Legend items only rendering with one item per row when there's definitely more room/ 
image

Fix: style tweaks, plus run the formatGrid() function on mounted and on resize (behind a debounce for performance) in order to consistently re-arrange the legend grid for the available space. 
image

#### Resources

### DIFF
--- a/packages/analytics/analytics-chart/src/components/chart-plugins/ChartLegend.vue
+++ b/packages/analytics/analytics-chart/src/components/chart-plugins/ChartLegend.vue
@@ -169,6 +169,11 @@ onMounted(() => {
   if (legendContainerRef.value) {
     resizeObserver.observe(legendContainerRef.value)
   }
+
+  window.requestAnimationFrame(() => {
+    checkForWrap()
+    formatGrid()
+  })
 })
 
 onBeforeUnmount(() => {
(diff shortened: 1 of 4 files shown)


packages/analytics/analytics-chart/src/components/chart-plugins/ChartLegend.vue
```diff
@@ -140,11 +141,26 @@ watch(() => props.items, () => {
...
+const oldHeight = ref(0)
+
+const resizeObserver = new ResizeObserver(debounce((entries: ResizeObserverEntry[]) => {
+
+  const entry = entries[0]
+  const newWidth = entry.contentRect.width
+  const newHeight = entry.contentRect.height
+
+  // Check if the resize is only in the X direction
+  if (newWidth !== oldWidth.value && newHeight === oldHeight.value) {
```
mihai-peteu, line 154: somewhat minor, but if don't need to `checkForWrap()` or `formatGrid()` on the very first page load, then we could exclude `oldWidth.value === 0`

filipgutica: we do want on the first page load :)

mihai-peteu (approved):

PR #1092, opened 2024-01-20
fix(entities-plugins): fix DataDog plugin tags array conversion [khcp-10331]

mptap:
# Summary
https://konghq.atlassian.net/browse/KHCP-10331
The issue is actually in the nested tags under `metrics` specific to Datadog plugin, and NOT the root level tags which are already converted to/from array during load and PUT/POST [here](https://github.com/Kong/public-ui-components/blob/7d29e61bec17dc450d0c39ee2140a9b05e5fa64e/packages/entities/entities-plugins/src/components/PluginEntityForm.vue#L277-L311).


kaiarrowood (approved):
LGTM! 🚀

PR #1093, opened 2024-01-22
feat(entities-plugins): add a prop to disable consumer group scope

Leopoldthecoder:
# Summary

Kong Manager OSS does not have consumer groups. So in its plugin forms, the consumer group scope should be hidden.

#### Resources


sumimakito (approved):
LGTM 👏

PR #1119, opened 2024-01-29
feat: explore csv standalone

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-2534

image

#### Resources


packages/analytics/analytics-chart/src/components/CsvExportModal.vue
```diff
@@ -97,7 +97,6 @@ const { i18n } = composables.useI18n()
 const props = withDefaults(defineProps<{
   filename: string,
   modalDescription?: string,
-  selectedRange: string,
```
filipgutica, line 100: This prop is not necessary since this can be computed from the chart data using the composable

adorack: Nice!

It looks like you should update the corresponding component test to stop passing the `selectedRange` prop.  Otherwise, code LGTM; checking the sandbox.

adorack (changes requested):

packages/analytics/analytics-chart/src/components/CsvExportModal.vue
filipgutica: done

PR #1124, opened 2024-01-30
feat(analytics-chart): things we need from konnect-vitals [MA-2557]

mihai-peteu:
# Summary

https://konghq.atlassian.net/browse/MA-2557

#### Resources

### DIFF
--- a/packages/analytics/analytics-chart/src/utils/colors.ts
+++ b/packages/analytics/analytics-chart/src/utils/colors.ts
@@ -1,3 +1,23 @@
+import type { AnalyticsChartColors } from '../../src/types'
+
+// Collection of colors matching existing Kong Manager Status Code chart
+const kongManangerColorPalette = {
+  '1XXCount|1XX|1[0-9][0-9]': { solid: '#0072E5', light: '#4DA6FF', description: 'Informational' },
+  'SuccessfulRequests|2XXCount|2XX|2[0-9][0-9]|SUCCESS': { solid: '#0BB652', light: '#6FCC83', description: 'Success' },
+  '3XXCount|3XX|3[0-9][0-9]': { solid: '#FDC53B', light: '#FFD982', description: 'Redirection' },
+  '4XXCount|4XX|4[0-9][0-9]': { solid: '#FE9439', light: '#FFBA81', description: 'Client Errors' },
+  '5XXCount|5XX|5[0-9][0-9]': { solid: '#FF4545', light: '#FF8484', description: 'Server Errors' },
+  'errorCount|FailedRequests|FAILURE': { solid: '#FF4545', light: '#FF8484', description: 'Server Errors' },
+  nonStandard: { solid: '#6f7787', light: '#e7e7ec' }, // grey-500, grey-300
+  standard: { solid: '#1155cb', light: '#8ab3fa' }, // blue-500, blue-300,
+  p99: { solid: '#1356cb', light: '#1356cb' },
+  p95: { solid: '#1fbecd', light: '#1fbecd' },
+  p50: { solid: '#1df97d', light: '#1df97d' },
+  LatencyP99: { solid: '#1356cb', light: '#1356cb' },
+  LatencyP95: { solid: '#1fbecd', light: '#1fbecd' },
+  LatencyP50: { solid: '#1df97d', light: '#1df97d' },
+}
+
 export const datavisPalette = [
   '#a86cd5',
   '#6a86d2',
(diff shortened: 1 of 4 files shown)


mihai-peteu:
@filipgutica I think we're exporting these two files twice (once from `src/utils/index`, once from the root `src/index`)

```
export * from './constants'
export * from './colors'
```

packages/analytics/analytics-chart/src/utils/constants.ts
```diff
@@ -40,4 +42,8 @@ export function formatTime(ts: number, options: TimeFormatOptions = {}) {
   }
 }
 
+export function formatTimeRange(start: number, end: number) {
+  return `${formatTime(start)} - ${formatTime(end, { includeTZ: true })}`
+}
+
```
filipgutica, line 48: Actually I think this belongs more in `analytics-utilities` package

filipgutica (changes requested):
actually one comment

packages/analytics/analytics-chart/src/utils/constants.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { describe, it, expect } from 'vitest'
```
filipgutica, line 1: This test can go to analytics-utilities along with the `formatTime()` function

filipgutica (changes requested):
a few comments

packages/analytics/analytics-chart/src/utils/constants.ts
```diff
@@ -5,6 +5,8 @@ export const DECIMAL_DISPLAY = 2
 export const FONT_SIZE_SMALL = 10
 export const FONT_SIZE_REGULAR = 12
 
+export const CHART_REFRESH_INTERVAL_MS = 30 * 1000
```
filipgutica, line 8: I think this should either go to `analytics-utilities` or remain in whatever host app is using `analytics-chart`

packages/analytics/analytics-chart/src/utils/colors.ts
```diff
@@ -51,3 +71,50 @@ export const darkenColor = (hex: string, amt: number): string => {
...
+export const apiRequestStatusCodeColors = {
+  100: { background: '#f0f5fd', text: '#10599e' },
+  200: { background: '#e5f7f4', text: '#1b6955' },
+  300: { background: '#fff1d5', text: '#a06027' },
+  400: { background: '#fee3b6', text: '#944d19' },
+  500: { background: '#fedada', text: '#852021' },
+}
+
+export const statusCodeBadgeBackgroundColor = (statusCode: string) => {
+  // @ts-ignore TODO
```
adorack, line 84: We shouldn't ignore the type mismatch; the signature of this function currently doesn't make sense.

packages/analytics/analytics-chart/src/utils/colors.ts
```diff
@@ -51,3 +71,50 @@ export const darkenColor = (hex: string, amt: number): string => {
...
+  return apiRequestStatusCodeColors[Math.floor(statusCode / 100) * 100]?.background || '#fafafa'
+}
+
+// TODO: Is this unused ?
```
adorack, line 88: Weird, it looks like it is.  I think it's better to remove it than have it commented out.

adorack (changes requested):

packages/analytics/analytics-chart/src/utils/constants.ts
mihai-peteu: but `formatTime` is in `analytics-chart`, and it's imported from there in a few other places already.

And `formatTimeRange` requires `formatTime`, so I would have to duplicate it in `analytics-utilities`.  I will do so, to avoid breaking anything, leaving a comment in place in `analytics-chart` that we should eventually settle on `-utilities` as the source of truth.

packages/analytics/analytics-chart/src/utils/colors.ts
```diff
@@ -81,16 +89,10 @@ export const apiRequestStatusCodeColors = {
 }
 
 export const statusCodeBadgeBackgroundColor = (statusCode: string) => {
-  // @ts-ignore TODO
-  return apiRequestStatusCodeColors[Math.floor(statusCode / 100) * 100]?.background || '#fafafa'
-}
-
-// TODO: Is this unused ?
+  const keyIndex = Math.floor(statusCode as any as number / 100) * 100
```
mihai-peteu, line 92: @adorack I've typed the status code colors at the top of the file now. 

Even though I have the key as `statusCode: number`, I encounter trouble on the line above, which is why i've added `as any as number` :/

adorack: You shouldn't need this cast -- why is the function attempting to divide `statusCode` by 100 if `statusCode` is actually a string?  Either the function takes a string and needs to `parseInt` it into a number, or it takes a number and doesn't need the cast...

adorack (changes requested):

packages/analytics/analytics-chart/src/utils/colors.ts
```diff
@@ -51,3 +79,44 @@ export const darkenColor = (hex: string, amt: number): string => {
...
+export const apiRequestStatusCodeColors: StatusCodeColors = {
+  100: { background: '#f0f5fd', text: '#10599e' },
+  200: { background: '#e5f7f4', text: '#1b6955' },
+  300: { background: '#fff1d5', text: '#a06027' },
+  400: { background: '#fee3b6', text: '#944d19' },
+  500: { background: '#fedada', text: '#852021' },
+}
+
+export const statusCodeBadgeBackgroundColor = (statusCode: string) => {
+  const keyIndex = Math.floor(parseInt(statusCode) / 100) * 100
```
adorack, line 92: Minor: any time you use `parseInt`, you should explicitly specify the base: `parseInt(num, 10)`.  Otherwise Weird Things can happen.

adorack (changes requested):

packages/analytics/analytics-chart/src/utils/colors.ts
mihai-peteu: Good call. Updated.

filipgutica (approved):

adorack (approved):

PR #1153, opened 2024-02-12
feat(plugins): allow creating dup plugins [khcp-10683]

kaiarrowood:
# Summary

Remove the logic that prevents users from adding duplicate plugins of the same type to the same entity. We need to support creating multiple plugins of the same type for `consumers`/`consumer_groups` to allow scoping to multiple entities simultaneously (ie. I can create a plugin scoped to a specific `consumer_group`/`route` and have a different configuration for the same plugin when it's scoped to a specific `consumer_group`/`gateway-service` correctly).
Addresses [KHCP-10683](https://konghq.atlassian.net/browse/KHCP-10683).

#### Resources

[KHCP-10683]: https://konghq.atlassian.net/browse/KHCP-10683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-plugins/src/components/PluginSelect.vue
+++ b/packages/entities/entities-plugins/src/components/PluginSelect.vue
@@ -344,14 +344,6 @@ const buildPluginList = (): PluginCardList => {
         plugin.disabledMessage = props.disabledPlugins[pluginId]
       }
 
-      if (existingEntityPlugins.value.includes(pluginId)) {
-        plugin.exists = true
-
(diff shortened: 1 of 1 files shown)


mptap (approved):
Code looks fine 👍 
Tested functionally in local sandbox 🚢 
Screenshot 2024-02-12 at 2 28 04 PM

PR #1167, opened 2024-02-19
feat(entity-link): allow links to open in same tab [MA-2617]

mihai-peteu:
# Summary

https://konghq.atlassian.net/browse/MA-2617

Screenshot 2024-02-19 at 2 05 35 PM

_______
And the PR to update usage in the consuming app (Gateway Manager MFE):
https://github.com/Kong/konnect-ui-apps/pull/2455/files


packages/entities/entities-shared/src/components/entity-link/EntityLink.vue
```diff
@@ -30,7 +31,7 @@
       </KTooltip>
     </KExternalLink>
     <KTooltip
-      v-if="entityUuid"
+      v-if="entityUuid && external"
```
adamdehaven, line 34: question: why is the tooltip only needed if an external link?

mihai-peteu: We _should_ probably add a second prop that controls this separate from the `external` boolean. 

I'll check w/ UX on whether they see this as a potential use case (eg: we hide the external link icon, but then keep the "copy uuid or name" tooltip)

mihai-peteu: @adamdehaven How about I default the Copy ID tooltip to `true`, and add a second `allowCopy` prop?  That way we can control the two icons independent of each other.

adamdehaven: TBH I'm not completely familiar with how this component is used specifically in all of the shared components. We'd need to check the usage there first

mihai-peteu:
@adamdehaven @portikM I cheated and added some internal padding on the `ExternalLinkIcon`, so that it ends up being the same size as the `Copy` icon:

Screenshot 2024-02-19 at 3 17 28 PM

I made the assumption (and correct me if I'm wrong) that this was just a tad too big:
Screenshot 2024-02-19 at 3 12 28 PM

packages/entities/entities-shared/src/components/entity-link/EntityLink.vue
mihai-peteu: Currently, it's only used in OverviewOnboardingCards' TopNTables in Gateway Manager:
https://github.com/search?q=repo%3AKong%2Fkonnect-ui-apps%20EntityLink&type=code

And used by API Requests slideout in Konnect:
https://github.com/search?q=repo%3AKong%2Fkhcp-ui%20%3CEntityLink&type=code

So it's safe to make this changes, adoption is not quite yet widespread.

packages/entities/entities-shared/src/components/entity-link/EntityLink.vue
```diff
@@ -15,8 +15,9 @@
     <KExternalLink
       v-else-if="entityUuid"
       class="entity-link"
+      :hide-icon="newWindow ? false : true"
       :href="externalLink"
-      target="_blank"
+      :target="newWindow ? '_blank' : '_self'"
```
mihai-peteu, line 20: changed the window=blank prop to `newWindow`, since `external` was too close to `externalLink`

packages/entities/entities-shared/docs/entity-link.md
```diff
@@ -0,0 +1,66 @@
...
+- required: `false`
+- default: `''`
+
+The absolute URL to be used as the `href` attribute.
+
+#### `newWindow`
+
+- type: `Boolean`
+- required: `false`
+- default: `true`
```
kaiarrowood, line 44: question: is there any way we could default the boolean props to `false` so usage will always be ex. `new-window` instead of always `:new-window="false"`

mihai-peteu: Good point. there are literally only two places to update in consuming apps, so this is preferable :)

MBasani22 (approved):

PR #1168, opened 2024-02-20
DNM - feat(swagger): update to 5.1.0 swagger-ui

davidma415:
# Summary
A spike PR for updating to newest swagger-ui that supports `openapi: 3.1.0`.

Screenshot 2024-02-20 at 11 31 35 AM

I think the breaking changes relate to JSON schemas being used now.

#### Resources

### DIFF
--- a/packages/portal/swagger-ui-web-component/package.json
+++ b/packages/portal/swagger-ui-web-component/package.json
@@ -17,7 +17,7 @@
     "react": "17.0.2",
     "react-dom": "17.0.2",
     "swagger-client": "^3.25.0",
-    "swagger-ui": "^4.19.1"
+    "swagger-ui": "5.1.0"
   },
   "repository": {
     "type": "git",
--- a/packages/portal/spec-renderer/sandbox/test.yaml
+++ b/packages/portal/spec-renderer/sandbox/test.yaml
@@ -1,781 +1,178 @@
-swagger: '2.0'
-info:
-  description: 'This is a sample server Petstore server.  You can find out more about Swagger at [http://swagger.io](http://swagger.io) or on [irc.freenode.net, #swagger](http://swagger.io/irc/).  For this sample, you can use the api key `special-key` to test the authorization filters.'
-  version: 1.0.0
-  title: Swagger Petstore v3
-  termsOfService: 'http://swagger.io/terms/'
-  contact:
-    email: apiteam@swagger.io
-  license:
-    name: Apache 2.0
-    url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
-host: petstore.swagger.io
-basePath: /v3
-tags:
-  - name: pet
-    description: Everything about your Pets
-    externalDocs:
-      description: Find out more
-      url: 'http://swagger.io'
-  - name: pet too
-    description: Access to Petstore orders
-  - name: store
-    description: Access to Petstore orders
-  - name: user
-    description: Operations about user
-    externalDocs:
-      description: Find out more about our store
-      url: 'http://swagger.io'
-schemes:
-  - https
-  - http
-paths:
-  /pets:
-    post:
-      tags:
-        - pet too
-      summary: Add a new pet to the store with a space in the tag 
-      description: ''
-      operationId: addPet
-      consumes:
-        - application/json
-        - application/xml
-      produces:
-        - application/xml
-        - application/json
-      parameters:
-        - in: body
-          name: body
-          description: Pet object that needs to be added to the store
-          required: true
-          schema:
-            $ref: '#/definitions/Pet'
-      responses:
-        '405':
-          description: Invalid input
-      security:
-        - petstore_auth:
-            - 'write:pets'
-            - 'read:pets'
-    get:
-      tags:
-        - pet too
-      summary: Add a new pet to the store with a space in the tag and some characters
-      description: ''
-      operationId: add_a-Pet_now
-      consumes:
-        - application/json
-        - application/xml
-      produces:
-        - application/xml
-        - application/json
-      parameters:
-        - in: body
-          name: body
-          description: Pet object that needs to be added to the store
-          required: true
-          schema:
-            $ref: '#/definitions/Pet'
-      responses:
-        '405':
-          description: Invalid input
-      security:
-        - petstore_auth:
-            - 'write:pets'
-            - 'read:pets'
-    put:
-      tags:
-        - dog-go
-      summary: Dog-go example
-      description: ''
-      operationId: petHead
-      consumes:
-        - application/json
-        - application/xml
-      produces:
-        - application/xml
-        - application/json
-      parameters:
-        - in: body
-          name: body
-          description: Pet object that needs to be added to the store
-          required: true
-          schema:
-            $ref: '#/definitions/Pet'
-      responses:
-        '405':
-          description: Invalid input
-      security:
-        - petstore_auth:
-            - 'write:pets'
-            - 'read:pets'
-  /pet:
-    post:
-      tags:
-        - pet
-      summary: Add a new pet to the store
-      description: ''
-      operationId: addPet
-      consumes:
-        - application/json
-        - application/xml
-      produces:
-        - application/xml
-        - application/json
-      parameters:
-        - in: body
-          name: body
-          description: Pet object that needs to be added to the store
-          required: true
-          schema:
-            $ref: '#/definitions/Pet'
-      responses:
-        '405':
-          description: Invalid input
-      security:
-        - petstore_auth:
-            - 'write:pets'
-            - 'read:pets'
-    put:
-      tags:
-        - pet
-      summary: Update an existing pet
-      description: ''
-      operationId: updatePet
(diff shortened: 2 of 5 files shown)


mptap:
**got swagger UI styles working as expected without local import**
Screenshot 2024-03-01 at 11 27 35 AM

packages/portal/swagger-ui-web-component/package.json
```diff
@@ -17,7 +17,7 @@
     "react": "17.0.2",
     "react-dom": "17.0.2",
     "swagger-client": "^3.25.1",
-    "swagger-ui": "^4.19.1"
+    "swagger-ui": "5.1.0"
```
adamdehaven, line 20: question: is there a reason you're pinning the version?

davidma415:
If we just use swagger's styling
Screenshot 2024-03-04 at 7 11 23 AM

vs. this is how it is currently
Screenshot 2024-03-04 at 7 11 30 AM

I think we'd need design feedback on this, or atleast we should remove the background-color on the operation at the very least so it doesn't look too different than what it is now.

packages/portal/swagger-ui-web-component/public/petstore_v3.json
```diff
@@ -78,12 +78,12 @@
...
             "description": "Successful operation",
             "content": {
-              "application/xml": {
+              "application/json": {
                 "schema": {
                   "$ref": "#/components/schemas/Pet"
                 }
               },
-              "application/json": {
+              "application/xml": {
```
adamdehaven, line 86: These are being switched, is that intentional? same for below

packages/portal/swagger-ui-web-component/src/element.js
```diff
@@ -1,4 +1,5 @@
 import SwaggerUI from 'swagger-ui'
+import swaggerUIStyles from 'swagger-ui/dist/swagger-ui.css'
```
adamdehaven, line 2: Doesn't importing the styles defeat the purpose? @davidma415 check me here

davidma415: Unsure what the best approach is, but yeah I don't think importing the styles fits our use case. Especially in the case that we currently allow theming (e.g. in the dev portal) and it would also affect those colors.

Screenshot 2024-03-04 at 7 51 39 AM

packages/portal/swagger-ui-web-component/public/petstore_v3.json
mptap: @kaiarrowood check me here, I believe they're just being re-ordered alphabetically (`json` before `xml`) and there's no net change

packages/portal/spec-renderer/src/components/SpecDetails.cy.ts
```diff
@@ -42,6 +42,19 @@ describe('<SpecDetails />', () => {
     cy.getTestId('hide-essentials-styles').should('exist')
   })
 
+  // TODO: evaluate if we need to test this
```
mptap, line 45: @kaiarrowood I don't think we need this test to check for styles anymore, shall I remove it?

packages/portal/swagger-ui-web-component/public/petstore_v3.json
adamdehaven: that looks like the change, just making sure

packages/portal/swagger-ui-web-component/package.json
mptap: Unpinned

packages/portal/spec-renderer/src/components/SpecDetails.cy.ts
mptap: Removed

packages/portal/swagger-ui-web-component/src/element.js
mptap: Per the team standup discussion, we're now overriding the css background on operations to match our original spec renderer.
We still do need to import the swagger-ui stylesheet for basic styles.

mptap: Screenshot 2024-03-04 at 10 31 23 AM

jillztom:
Screenshot 2024-03-04 at 1 54 46 PM

We might need to tweak this CSS as well

packages/portal/swagger-ui-web-component/public/petstore_v3.json
kaiarrowood: This is an example `.json` file, I literally copied the latest example one from swagger.io so I'm guessing that's a change they made to the sample in the last few versions. No need to worry about it.

mptap:
@jillztom  > We might need to tweak this CSS as well
This occurs only in the sandbox, and not in the consuming apps, as they're using a custom style to override the background on hover.

packages/portal/spec-renderer/src/components/operations-list/OperationsListItem.vue
```diff
@@ -94,8 +94,6 @@ const methodName = computed((): string => {
   }
 
   &:hover {
-    background: var(--kong-ui-spec-renderer-operations-list-item-background-hover, $kui-color-background-inverse);
```
kaiarrowood, line 97: We need to add this back for theming purposes. Can we adjust the fallback variable (`$kui-color-background-inverse`) instead?

packages/portal/swagger-ui-web-component/src/element.js
```diff
@@ -162,6 +164,9 @@ export class SwaggerUIElement extends HTMLElement {
       console.warn('For correct positioning, you must enable the sidebar with `has-sidebar="true"` and should only display essentials with `essentials-only="true"`')
     }
 
+    // load swagger UI styles to shadow DOM
+    swaggerUIStyles.use({ target: this.shadowRoot, testId: 'swagger-ui-styles' })
```
kaiarrowood, line 168: We should add a test that this test-id exists in `SpecDetails.cy.ts` (we can add it as the last asserting in `it('renders correctly with props')`

packages/portal/spec-renderer/src/components/operations-list/OperationsListItem.vue
mptap: Gotcha, adjusted fallback variable

packages/portal/swagger-ui-web-component/src/element.js
mptap: Added

packages/portal/swagger-ui-web-component/src/styles/overrides.css
```diff
@@ -6,3 +6,23 @@
...
   padding-top: 0;
 }
+
+.swagger-ui .opblock {
+  background: none !important;
+}
+
+.swagger-ui .json-schema-2020-12-accordion,
+.swagger-ui .json-schema-2020-12-expand-deep-button {
+  background-color: var(--kong-ui-section_colors-body) !important;
```
adamdehaven, line 16: question: why a mix of kebab-case and snake_case?

Also, do we not need to provide a fallback value?

packages/portal/swagger-ui-web-component/src/styles/overrides.css
```diff
@@ -6,3 +6,23 @@
...
+}
+
+.swagger-ui .json-schema-2020-12-accordion__icon,
+.swagger-ui .json-schema-2020-12-accordion__icon--collapsed,
+.swagger-ui .json-schema-2020-12-accordion__icon--expanded {
+  /* stylelint-disable-next-line custom-property-pattern*/
+  color: var(--text_colors-primary) !important;
```
adamdehaven, line 23: question: why a mix of kebab-case and snake_case?

Also, do we not need to provide a fallback value?

packages/portal/swagger-ui-web-component/src/styles/overrides.css
mptap: The var comes from portal, I don't have a say

packages/portal/swagger-ui-web-component/src/styles/overrides.css
mptap: The var comes from portal, I don't have a say

adamdehaven (dismissed):
The section and text colors are not prefixed with `kong-ui-` in the codebase

adamdehaven:
> The section and text colors are not prefixed with `kong-ui-` in the codebase

e.g. you should be using `--text_colors-accent` and `--section_colors-accent` etc.

mptap:
> The section and text colors are not prefixed with `kong-ui-` in the codebase

Yes, I'm using `var(--text_colors-primary)` but linter's unhappy so added a disable for that line.

kaiarrowood (dismissed):
LGTM! 🚀 nice work @mptap !

davidma415:
Looks good to me from the portal POV with respect to theming etc. 👍  great job, thanks for helping out with this!

jillztom (approved):

adamdehaven (approved):

PR #1169, opened 2024-02-20
feat(dashboard-renderer): add metric card renderer [MA-2307]

adorack:
Currently just supports golden signals.  Styling TBD.

image

# Summary

#### Resources

### DIFF
--- a/packages/analytics/dashboard-renderer/src/components/GoldenSignalsRenderer.vue
+++ b/packages/analytics/dashboard-renderer/src/components/GoldenSignalsRenderer.vue
@@ -1,0 +1,23 @@
+<template>
+  <MetricsProviderInternal v-bind="options">
+    <MetricsConsumer />
+  </MetricsProviderInternal>
+</template>
+<script setup lang="ts">
+import type { MetricCardOptions, RendererProps } from '../types'
+import { MetricsProviderInternal, MetricsConsumer } from '@kong-ui-public/analytics-metric-provider'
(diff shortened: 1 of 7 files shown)


adamdehaven (approved):
lgtm

packages/analytics/dashboard-renderer/src/components/GoldenSignalsRenderer.vue
```diff
@@ -0,0 +1,23 @@
...
+
+const props = defineProps<RendererProps<MetricCardOptions>>()
+
+const options = computed<ProviderProps>(() => ({
+  longCardTitles: props.chartOptions.longCardTitles,
+  description: props.chartOptions.description,
+  hasTrendAccess: true,
+  refreshInterval: 0,
+}))
+</script>
```
filipgutica, line 23: We're not going to provide the flexibility yet for the host app to determine/define the metric cards they want to see, just keeping it at the three golden signals for now?

filipgutica (commented):
Just had one comment

packages/analytics/dashboard-renderer/src/components/GoldenSignalsRenderer.vue
adorack: That's correct.  A) There isn't any demand for that yet in any of the potential dashboards we want to create; and B) there's some complexity around which queries get issued, so it's not a feature we can just get for free.

PR #1171, opened 2024-02-21
feat(analytics-utilities): explore v4 translation [MA-2641]

mihai-peteu:
# Summary

https://konghq.atlassian.net/browse/MA-2641

#### Resources


adorack (changes requested):
I don't think we want this translator in public-ui.  Let's talk offline?

PR #1175, opened 2024-02-22
fix(entities-plugins): custom plugins are hidden in select page for KM

Leopoldthecoder:
# Summary

We should still show custom plugins in KM, just not in a separate tab.

#### Resources


sumimakito (approved):
LGTM 👍

PR #1188, opened 2024-02-26
 feat(dashboard-renderer): add TopN table support to dashboard renderer

filipgutica:
# Summary

image

add optional prop to hide the default border on TopNTable

dynamically set the hight of the horizontal legend based on if there needs to be
more than one row or not.

Only need to allocate space for one row if there's only one row.

If there's more than one row, allocate space for 2 rows, the rest overflows.

Set the chart-container hight to 100% - 80px to account for the legend + chart title
when chart and legend is stacked

https://konghq.atlassian.net/browse/MA-2637

#### Resources

### DIFF
--- a/packages/analytics/analytics-chart/sandbox/pages/BarChartDemo.vue
+++ b/packages/analytics/analytics-chart/sandbox/pages/BarChartDemo.vue
@@ -288,7 +288,7 @@ const dimensionItems = [{
 
 // Short labels
 const statusCodeLabels = [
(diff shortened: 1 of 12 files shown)


mihai-peteu:
Can you please remove this line (the inner padding), since we now provide it at the tile level?
https://github.com/Kong/public-ui-components/pull/1188/files#diff-ec373fa52acabacda6e6f187a9635f018b68300e231f5534fd68eb43e04bd48aR316

mihai-peteu (approved):

kaiarrowood (approved):

PR #1193, opened 2024-02-27
feat(dashboard-renderer): expand component test [MA-2629]

mihai-peteu:
# Summary

https://konghq.atlassian.net/browse/MA-2629


adorack (changes requested):
I'd like to review.

packages/analytics/dashboard-renderer/sandbox/mock-data.ts
```diff
@@ -223,3 +239,128 @@ export const timeSeriesExploreResponse: ExploreResultV4 = {
...
+        position: {
+          col: 0,
+          row: 3,
+        },
+        size: {
+          cols: 6,
+          rows: 2,
+        },
+      },
+    } as unknown as TileConfig,
```
adorack, line 364: Sorry that the types are messed up.  I'm working on improving that.

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.cy.ts
```diff
@@ -1,30 +1,75 @@
...
+        return Promise.resolve(nonTimeSeriesResponse)
+      }
+    }
+
+    return {
+      queryFn: cy.spy(queryFn).as('fetcher'),
+    }
+  }
+
+  afterEach(() => {
```
adorack, line 53: Could you add a code comment explaining why this is here?

adorack (changes requested):
Looking into the test order strangeness...

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.cy.ts
mihai-peteu: The first test in the file loads up in a very weird manner, and this injected CSS block was my attempt to style the Cypress component runner in a way that allows both charts to show inside the viewport.  But since the subsequent tests look good (weirdly enough), I should probably just remove it.

Screenshot 2024-03-06 at 12 37 10

adorack (approved):

PR #1197, opened 2024-02-28
fix(metric-cards): determine trend range text from response [MA-2650]

adorack:
- Attempt to move away from `timeframeLength`; use the response coming back from the server.
- Use `withTrend` option to determine whether a trend was queried or not.
- Properly pluralize 'previous day' / 'previous N days'

# Summary

#### Resources

### DIFF
--- a/packages/analytics/analytics-metric-provider/src/locales/en.json
+++ b/packages/analytics/analytics-metric-provider/src/locales/en.json
@@ -15,7 +15,7 @@
     }
   },
   "trendRange": {
-    "custom": "vs previous {numDays} days",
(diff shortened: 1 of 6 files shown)


packages/analytics/analytics-metric-provider/src/locales/en.json
```diff
@@ -15,7 +15,7 @@
     }
   },
   "trendRange": {
-    "custom": "vs previous {numDays} days",
+    "custom": "vs previous {numDays, plural, =1 {day} other {# days}}",
```
mihai-peteu, line 18: 🤯 This is wild

filipgutica (approved):

PR #1206, opened 2024-03-04
fix(analytics): update metric card styles [MA-2658]

adorack:
- Update metric card value font size across all instances
- Override title styling in dashboards and don't restrict card height

# Summary

#### Resources


mihai-peteu:
This is good to go.  

Just curious if you have an opinion on moving this `justify-content` attribute over to MetricsCard.vue, so it doesn't get "lost in translation" when we migrate our pages to the MFEs:

https://github.com/Kong/khcp-ui/pull/7703/files#r1511764394

(or, we can at least leave a comment above that line in Konnect to remind us to move it later)

adorack:
I think that's a good idea -- done.

mihai-peteu (approved):

PR #1210, opened 2024-03-05
fix(*): replace CopyUuid with KCopy [khcp-9465]

kaiarrowood:
# Summary

Replace usage of `CopyUuid` with `KCopy` for [KHCP-9465](https://konghq.atlassian.net/browse/KHCP-9465).

#### Resources


portikM:
Looks good to me code-wise, functionally just one nit/question:
I noticed ids in keys and key sets tables are not truncated while in dev currently they are (screenshot). Do we want to truncate those?
Almost the same applies to SNIs table except in dev they are not truncated currently but I think we should pick one standard and truncate ids in tables.
Screenshot 2024-03-06 at 5 18 08 PM
Screenshot 2024-03-06 at 5 14 38 PM

portikM (approved):
LGTM 🚀

PR #1213, opened 2024-03-06
feat(dashboard-renderer): add support for slottable tiles

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-2663

image

#### Resources

### DIFF
--- a/packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
+++ b/packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
@@ -16,6 +16,7 @@ export enum ChartTypes {
   TimeseriesLine = 'timeseries_line',
   GoldenSignals = 'golden_signals',
   TopN = 'top_n',
+  Slottable = 'slottable',
 }
 
 // Common definition for many ChartJS tiles.
@@ -36,6 +37,23 @@ const chartDatasetColorsSchema = {
     type: 'string',
   },
 } as const satisfies JSONSchema
+
+export const slottableSchema = {
+  type: 'object',
+  properties: {
+    type: {
+      type: 'string',
+      enum: [ChartTypes.Slottable],
+    },
+    id: {
+      type: 'string',
+    },
+  },
+  required: ['type', 'id'],
+  additionalProperties: false,
+} as const satisfies JSONSchema
+
+export type SlottableOptions = FromSchema<typeof slottableSchema>
 
 export const barChartSchema = {
   type: 'object',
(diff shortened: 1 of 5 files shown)


packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
```diff
@@ -390,10 +390,17 @@ export type TileLayout = FromSchema<typeof tileLayoutSchema>
 export const tileConfigSchema = {
   type: 'object',
   properties: {
+    id: {
```
adorack, line 393: I kind of feel like the slottable tile should be a new "chart type" with `id` as a property -- otherwise, you could potentially have a tile that's slottable AND has a chart, and I feel like that doesn't make sense.  Happy to talk about it, though.

adorack (changes requested):

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
filipgutica: I started going down that route, but it too messy for my liking and forced the user of the dashboard to include redundant properties. There's no need for a "chart" and there's no need for a "query" for a slottable tile. 

As a result, a slottable component does not need anything from "definition" at all. I'd rather simplify and cut out the definition altogether for a slottable tile than over complicate further up the chain, just to avoid the one scenario you mention which can be cleared up in the documentation or with a console warn. 

Also, it forces us to add a `slottable` enum to the ChartTypes enum to satisfy the renderer lookup. That's also unneeded since with a slottable tile, the `DashboardTile` component doesn't even get rendered, and the code never even gets there with a slottable tile. 

Let me know if there's any other situations where we might want to pass additional information and context to a slottable tile that might require it to be a new "chart type"
(which I don't think is the case since the host app will be declaring it in its template)

filipgutica: Passing through the `syntheticsDataKey` to the slot might be a good idea

filipgutica: Ok, made a change to try and get the best of both worlds.

Keep the tile definition consistent while avoiding doing the nested slots 😬

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
```diff
@@ -393,7 +411,7 @@ export const tileConfigSchema = {
     definition: tileDefinitionSchema,
     layout: tileLayoutSchema,
   },
-  required: ['definition', 'layout'],
```
adorack, line 396: I think this should still be required.

adorack (changes requested):
I think I'm fine with this.  Thanks for looking into it and finding a way to keep the schema consistent!

One minor comment.

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
filipgutica: ah yes

adorack (approved):

PR #1230, opened 2024-03-12
fix(entities-shared): update useFetchUrlBuilder to allow exact match search from host app. [KHCP-11054]

lukewhchen:
# Summary
Ticket: https://konghq.atlassian.net/browse/KHCP-11054

Update `useFetchUrlBuilder` to allow using exact match search on upstream list from konnect.

[slack](https://kongstrong.slack.com/archives/C03LRB400TC/p1709217149772489?thread_ts=1709216110.973549&cid=C03LRB400TC): 
Upstreams are not being served from Kanalytics/Open Search. We are still hitting Koko for upstream.
In this case, the upstream page should be using the old way of getting by name/id not the filter call

#### Resources


Leopoldthecoder (approved):
LGTM

PR #1239, opened 2024-03-13
feat(useFetcher): allow passing in axios request config

adamdehaven:
# Summary

Add a new, optional `axiosRequestConfig` option to the shared `useFetcher` function to allow passing in an `AxiosRequestConfig` object.

As an example, if you have an endpoint needing to allow `404` status responses, you could do something like this:

```typescript
const { fetcher, fetcherState } = useFetcher(props.config, fetcherBaseUrl.value, undefined, {
  validateStatus: (status: any) => status === 404 || (status >= 200 && status < 300),
})
```


Leopoldthecoder:
Closing in favor of https://github.com/Kong/public-ui-components/pull/1243

PR #1240, opened 2024-03-13
feat(analytics-chart): experimenting with legend styling

mihai-peteu:
# Summary

@filipgutica This is an attempt to simplify the Legend height conundrum, and keep it styled via pure CSS (thus allowing us to do away with the bound CSS variable). 

Using color squares instead of lines, and also left-aligning the legend (roughly in line w/ the Y-Axis labels)
Screenshot 2024-03-19 at 11 36 51

#### Resources


packages/analytics/analytics-chart/src/components/chart-plugins/ChartLegend.vue
```diff
@@ -66,44 +66,12 @@ const legendItemsRef = ref<HTMLElement[]>([])
 const showValues = inject('showLegendValues', true)
 const position = inject('legendPosition', ref(ChartLegendPosition.Right))
 const legendItemsTracker = ref<LegendItem[]>([])
-const legendHeight = ref<string>(KUI_SPACE_80)
-
-// Check if the legend wraps by comparing the top position of each item.
-const doesTheLegendWrap = () => {
```
filipgutica, line 72: I think we still need some of this code, just playing around in the sandbox I see some issues with legend items overlapping 

image

filipgutica (changes requested):

filipgutica (changes requested):
Sometimes still overlapping in some situations and also sometimes not showing the second row

image

image

Also, when it wraps to a new line, should there be more space between rows?
image

packages/analytics/analytics-chart/src/components/chart-plugins/ChartLegend.vue
```diff
@@ -96,7 +95,7 @@ const checkForWrap = () => {
   if (legendContainerRef.value && position.value === ChartLegendPosition.Bottom) {
     if (doesTheLegendWrap()) {
       // Allow for two rows of legend items
-      legendHeight.value = KUI_SPACE_110
```
filipgutica, line 99: May be we should still use 110 here? 🤔

filipgutica (changes requested):
There's still some weirdness with the spacing of the legend items, some legend color tags also don't show up or are getting cut off?

On this branch
image


Vs on dev:
image

mihai-peteu:
Closing in favor of https://github.com/Kong/public-ui-components/pull/1462

But the `legendAsGrid` mixin could prove useful in the future...

PR #1245, opened 2024-03-14
feat(dashboard-renderer): stub out support for EntityLink [MA-2716]

mihai-peteu:
# Summary
https://konghq.atlassian.net/browse/MA-2716

#### Resources


packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
```diff
@@ -133,6 +133,9 @@ export const topNTableSchema = {
     description: {
       type: 'string',
     },
+    entityLink: {
+      type: 'string',
+    },
```
mihai-peteu, line 138: This will probably need to be an array of strings

packages/analytics/dashboard-renderer/src/composables/useExternalLinkCreator.ts
```diff
@@ -0,0 +1,12 @@
...
+export default (routeParams: string[]): string => {
+  // TODO: may have to pass in geo from consuming app
+  const activeGeoCode = 'us'
+
+  const parsedRouteParams = routeParams.map(rp => rp.replace('{geo}', activeGeoCode))
+
+  // TODO: replace `{idTuple0}` and `{idTuple1}` with actual record ID vales
+
+  return useExternalLinkCreator(parsedRouteParams)
+}
```
mihai-peteu, line 12: if this isn't going to be shared amongst other charts in the dashboard, the logic could all be stashed inside `generateEntityUrl` in TopNTableRenderer most likely.

packages/analytics/analytics-chart/sandbox/pages/ChartDemoSimple.vue
```diff
@@ -229,9 +229,9 @@ import {
   SimpleChart,
   TopNTable,
 } from '../../src'
+import { SeededRandom } from '@kong-ui-public/analytics-utilities'
 import type { AnalyticsExploreRecord, DisplayBlob, ExploreResultV4, QueryResponseMeta } from '@kong-ui-public/analytics-utilities'
 import type { AnalyticsChartColors, SimpleChartOptions } from '../../src/types'
-import { SeededRandom } from '../utils/SeedRandom'
```
mihai-peteu, line 234: @filipgutica this is my fault, I forgot to update the import in a previous PR 😬  Without this fix, the SimpleChart demo doesn't load.

packages/analytics/dashboard-renderer/src/composables/useExternalLinkCreator.ts
filipgutica: We don't need use external link creator here 

The host app is supposed to do this, the host app supplies the output of external link creator to the dashboard and the renderer  replaces the appropriate ids

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
filipgutica: just one string. The urls for an entity type have the same format

PR #1250, opened 2024-03-15
refactor: expose swagger UI to enable OAuth2 try-it-out in upstream applications

andrewwylde:
# Summary

This is a Proof-of-Concept to attempt a fix to some issues in an [upstream repository](https://github.com/Kong/konnect-portal) (`konnect-portal`). The issue is that for client-side-only OIDC "Try It Out" authorization, one needs to be able to initialize oauth with certain values available.

## More info!

For SPA-based requests to the `/authorize` endpoint, one needs to invoke the following (where `swaggerUI` is an instance that has already been initialized.
```typescript
swaggerUI.initOAuth({
          usePkceWithAuthorizationCodeGrant: true,
          additionalQueryStringParams: {
            nonce: Math.random().toString(36).substring(7)
          }
        })
```

Rather than initializing that within `spec-renderer` (really within `swagger-ui-web-component`), I think the better move is to allow more flexibility and to expose `#instance` so that folks can call it from the consuming application.

Open to differing opinions, this is just how we got things working for our use case: https://gist.github.com/andrewwylde/d47c94c7629a81a0232ac80ef862e788

### DIFF
--- a/packages/portal/spec-renderer/src/components/SpecDetails.vue
+++ b/packages/portal/spec-renderer/src/components/SpecDetails.vue
@@ -76,6 +76,10 @@ const { i18n } = composables.useI18n()
 
 const swaggerRef = ref<SwaggerUIElement | null>(null)
 
+defineExpose({
+  swaggerInstance: swaggerRef,
+})
+
 const hasRequiredProps = computed((): boolean => {
   return !!(props.document || props.url)
 })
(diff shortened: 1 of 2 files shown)


Mierenga (approved):
👍

adamdehaven:
@andrewwylde can you add your team to the CODEOWNERS for this package? https://github.com/Kong/public-ui-components/blob/57142b2efffa5ef98fe3383489b95b96a7b7101d/CODEOWNERS#L40

nateslo:
> @andrewwylde can you add your team to the CODEOWNERS for this package?
> 
> https://github.com/Kong/public-ui-components/blob/57142b2efffa5ef98fe3383489b95b96a7b7101d/CODEOWNERS#L40

https://github.com/Kong/public-ui-components/pull/1266

PR #1252, opened 2024-03-15
fix(*): proper heading element in widgets [KHCP-11027]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-11027

Update title in widget-like components to come with appropriate heading element

Applied fixed can be found in PRs picking up new versions of published packages:

- konnect-ui-apps: https://github.com/Kong/konnect-ui-apps/pull/2698
- khcp-ui: https://github.com/Kong/khcp-ui/pull/7762

#### Resources


packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -6,12 +6,12 @@
       v-if="title"
       #title
     >
-      <div
+      <h2
```
adamdehaven, line 9: nit: you'll likely want the Analytics team to review this to ensure no surprises for Christian or Helen

portikM: Pushed a commit to simplify these change a bit, but it shouldn't change visually/functionally, just the element. Will tag data team to make sure they are aware of this changes

portikM: Oh looks like Mihai already saw it

mihai-peteu: lol - yea just one quick heads up. If we went the `h2` route, it would turn into `var(--kui-font-size-50, 18px)`

And what we have now across the board for all Gateway Manager overview onboarding card Titles is `16px`. Would `<h3>` give us 16px font size?

Generally speaking, I'm all about using using heading tags to control font sizes, we'd just have to make sure all the other `.section-title`s are changed to match.

portikM: Oh shoot you're right, KCard title is `18px` by default. that explains why we're overriding it so much everywhere. I'll ask design team whether we're fine making KCard title `16px` since it's the component we use for widgets by default.

portikM: And no, we shouldn't use `h3` here because we're trying to standardize that widgets use `h2` as they are usually second level importance headings on the page after the page heading component (in most of our pages at least).

mihai-peteu: I see...

I take back my comment about "using heading tags to control font sizes" – since each chart type (including TopNTable) could end up in an embedded dashboard, we probably should avoid using explicit heading tags inside the dashboard because it is not aware of the consuming app's heading hierarchy. 

Even though `<h2>` for the titles may work 90% of the time, I bet that eventually we would run into an heading structure issue.

portikM: We could add `titleTag` prop here like we did in [EntityBaseTable.vue](https://github.com/Kong/public-ui-components/pull/1252/files#diff-fc2212f00918166c6240f9e29994448b13f8e9e154d57c506943a0467fcb596b). Would that work?

portikM: We decided we'll add `titleTag` prop whenever possible to give a bit more flexibility in those components

mihai-peteu (approved):
Provisionally approving this as far changes to analytics components go :)

adamdehaven (approved):

packages/analytics/metric-cards/src/components/display/MetricsCard.vue
```diff
@@ -183,6 +186,10 @@ const props = defineProps({
     required: false,
     default: () => 'lg',
   },
+  titleTag: {
```
adorack, line 189: Could you explain more about the use case behind making this a property?  I'd expect we'd want all metric cards to have the same title tag...

adorack (changes requested):

packages/analytics/metric-cards/src/components/display/MetricsCard.vue
portikM: That is the case most of the time, however if metrics card is seeded inside some other widget that already has an h2 heading - it should be h3 to not break the hierarchy - that's when this prop comes in handy. Regardless of what tag is used - styling for the title remains consistent though

adorack (approved):

PR #1253, opened 2024-03-15
feat: markdown editor [KHCP-11101]

adamdehaven:
# Summary

Implement the new `@kong/markdown` editor UI

### DIFF
--- a/packages/core/documentation/package.json
+++ b/packages/core/documentation/package.json
@@ -38,6 +38,16 @@
     "test:unit": "cross-env FORCE_COLOR=1 vitest run",
     "test:unit:open": "cross-env FORCE_COLOR=1 vitest --ui"
   },
+  "dependencies": {
+    "@kong-ui-public/entities-shared": "workspace:^",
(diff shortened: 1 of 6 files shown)


packages/core/documentation/package.json
```diff
@@ -60,17 +70,7 @@
...
-    "@kong-ui-public/i18n": "workspace:^",
-    "@kong/kongponents": "9.0.0-alpha.113",
-    "vue": "^3.4.21"
-  },
-  "dependencies": {
-    "@kong-ui-public/document-viewer": "workspace:^",
-    "@kong-ui-public/entities-shared": "workspace:^",
-    "@kong/icons": "^1.8.14"
+    "errorLimit": "30MB",
+    "warningLimit": "25MB"
```
adamdehaven, line 74: This is expected, the new `@kong/markdown` package is essentially adding `mermaid` support as well as enabling many more languages by default. We can shave this down in the future once we determine the langs we want to support out of the box.

jillztom (approved):

PR #1257, opened 2024-03-18
feat(analytics): move to v2 config endpoint [MA-2730]

adorack:
BREAKING CHANGE: new config endpoint; types updated

The following components are updated:

- AnalyticsConfigCheck, and the associated store
- MetricsProviderInternal
- AnalyticsBridge (updates type that implementations must adhere to)
- DashboardRenderer (sandbox and tests)

# Summary

#### Resources


packages/analytics/analytics-config-store/src/components/AnalyticsConfigCheck.vue
```diff
@@ -24,8 +24,8 @@ const analyticsConfigStore = useAnalyticsConfigStore()
 const analyticsConfig: Ref<ConfigStoreState> = analyticsConfigStore.getConfig()
 
 const loading = computed<boolean>(() => !analyticsConfig.value)
-const hasAnalytics = computed<boolean>(() => !!analyticsConfig.value && analyticsConfig.value.analytics)
-const hasPercentiles = computed<boolean>(() => !!analyticsConfig.value && analyticsConfig.value.analytics && analyticsConfig.value.percentiles)
+const hasAnalytics = computed<boolean>(() => !!analyticsConfig.value && !!analyticsConfig.value.analytics)
+const hasPercentiles = computed<boolean>(() => !!analyticsConfig.value && !!analyticsConfig.value.analytics?.percentiles)
```
hfukada, line 28: just a curiosity comment here, how come the first part of this isn't a redundant condition? 
could it not just be `!!analyticsConfig.value.analytics?.percentiles`?

hfukada (dismissed):

packages/analytics/analytics-config-store/src/components/AnalyticsConfigCheck.vue
adorack: That was redundant -- thanks, good catch.  (I did run into a frustrating TS issue in `MetricsProviderInternal` where TS didn't like comparing `undefined` with a number, even though the outcome seemed stable.  But that's not the case here.)

hfukada (approved):

PR #1271, opened 2024-03-20
feat: allow dashbaord renderer to optionally auto refresh queries

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-2748

https://www.loom.com/share/e6093efdbbc6426daea4a1f33d3801e2

#### Resources

### DIFF
--- a/packages/analytics/dashboard-renderer/sandbox/pages/RendererDemo.vue
+++ b/packages/analytics/dashboard-renderer/sandbox/pages/RendererDemo.vue
@@ -43,6 +43,7 @@ const context: DashboardRendererContext = {
     type: 'relative',
     time_range: '24h',
   },
+  refreshInterval: 5000,
 }
 
 const dashboardConfig: DashboardConfig = {
@@ -140,7 +141,6 @@ const dashboardConfig: DashboardConfig = {
         query: {
           dimensions: ['time'],
         },
-        refreshInterval: 5000,
       },
       layout: {
         position: {
--- a/packages/analytics/dashboard-renderer/sandbox/pages/RendererDemo.vue
+++ b/packages/analytics/dashboard-renderer/sandbox/pages/RendererDemo.vue
@@ -43,6 +43,7 @@ const context: DashboardRendererContext = {
     type: 'relative',
     time_range: '24h',
   },
+  refreshInterval: 5000,
 }
 
 const dashboardConfig: DashboardConfig = {
--- a/packages/analytics/dashboard-renderer/sandbox/pages/RendererDemo.vue
+++ b/packages/analytics/dashboard-renderer/sandbox/pages/RendererDemo.vue
@@ -43,7 +43,6 @@ const context: DashboardRendererContext = {
     type: 'relative',
     time_range: '24h',
   },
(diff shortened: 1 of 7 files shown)


packages/analytics/dashboard-renderer/sandbox/pages/RendererDemo.vue
```diff
@@ -43,6 +43,7 @@ const context: DashboardRendererContext = {
     type: 'relative',
     time_range: '24h',
   },
+  refreshInterval: 5000,
```
mihai-peteu, line 46: i know this is just the demo page, but shouldn't we make it a little longer, like 20 seconds or so? :)

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
```diff
@@ -6,7 +6,8 @@ import type { ExploreFilter, ExploreQuery, TimeRangeV4 } from '@kong-ui-public/a
 export interface DashboardRendererContext {
   filters: ExploreFilter[]
   timeSpec: TimeRangeV4
-  tz?: string
+  tz?: string,
+  refreshInterval?: number
```
adorack, line 10: I don't feel like this is a good fit for context.  It seems like it should be part of the definition.  Maybe we can discuss after all the meetings?

adorack (changes requested):

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
filipgutica: I just wanted to put it in a spot that made the least amount of code changes, and not add additional props to the component 

I thought the context is a perfect spot for it, makes sense in my mind as it applies to the whole dashboard.

Or are you thinking we should specify refresh interval per tile?

packages/analytics/dashboard-renderer/sandbox/pages/RendererDemo.vue
filipgutica: It's just the sandbox, I wanna see it quick, not sit around for 20-30 seconds 😄

packages/analytics/dashboard-renderer/src/components/QueryDataProvider.vue
```diff
@@ -96,7 +96,7 @@ const { data: v4Data, error, isValidating } = useSWRV(queryKey, async () => {
     emit('queryComplete')
   }
 }, {
-  refreshInterval: 0,
+  refreshInterval: props.context.refreshInterval ?? 0,
```
adorack, line 99: Let's set the default to 30 seconds, based on the argument that we usually want it in order to showcase real-time data.

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
adorack: (Resolved out of band -- I retract my objection, sorry about that.)

adorack (changes requested):

adorack (approved):

PR #1277, opened 2024-03-21
fix(metric-cards): don't default to a heading tag

mihai-peteu:
# Summary

Back to normal:
Screenshot 2024-03-21 at 07 36 14

#### Resources


packages/analytics/metric-cards/src/components/display/MetricsCard.vue
```diff
@@ -11,7 +11,11 @@
         :color="KUI_COLOR_TEXT_NEUTRAL"
         :size="KUI_ICON_SIZE_30"
       />
-      <component :is="titleTag">
+      <span v-if="!titleTag">{{ title }}</span>
```
adorack, line 14: Instead of the `v-if`, can we just default the `titleTag` property to `span` (pending discussion)?

adorack (changes requested):

adorack (approved):

PR #1291, opened 2024-03-25
fix(dashboard-renderer): ResizeObserver on dashboard tiles

filipgutica:
# Summary

implement resize observer on dashboard tiles to notify analytics-chart of height updates when rows have tiles that are "fitToContent"

#### Resources


adorack (changes requested):
I'd like a bit more time to think about failure states and possibly try it out in the sandbox.

PR #1303, opened 2024-04-01
feat(analytics-config-store): export `configResult` and some minor fix

nekolab:
# Summary

This commit exports `configResult` to prevent some reactivity issues. After that, `getConfig` could be used as a pure action.
    
Besides, we mark `fetchedConfig` to false when `configFn` returns with error, and hoisting the `queryBridge` out of the `getConfig` to prevent some unintentional behavior

#### Resources

### DIFF
--- a/packages/analytics/analytics-config-store/src/stores/analytics-config.ts
+++ b/packages/analytics/analytics-config-store/src/stores/analytics-config.ts
@@ -10,10 +10,9 @@ export type ConfigStoreState = null | AnalyticsConfigV2
 export const useAnalyticsConfigStore = defineStore('analytics-config', () => {
   let fetchedConfig = false
   const configResult = ref<ConfigStoreState>(null)
+  const queryBridge: AnalyticsBridge | undefined = inject(INJECT_QUERY_PROVIDER)
 
   const getConfig = (): Ref<ConfigStoreState> => {
-    const queryBridge: AnalyticsBridge | undefined = inject(INJECT_QUERY_PROVIDER)
-
     if (!queryBridge) {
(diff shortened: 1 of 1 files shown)


packages/analytics/analytics-config-store/src/stores/analytics-config.ts
```diff
@@ -10,10 +10,10 @@ export type ConfigStoreState = null | AnalyticsConfigV2
 export const useAnalyticsConfigStore = defineStore('analytics-config', () => {
   let fetchedConfig = false
   const configResult = ref<ConfigStoreState>(null)
```
adorack, line 12: This variable already acts as the config store's loading state.  If it's `null`, the config is still loading.

I think you're concerned about the "no analytics" case.  If the org doesn't have analytics, the config store contains `{ analytics: null, requests: null }` -- in other words, it's possible to distinguish the two cases.

adorack (changes requested):

packages/analytics/analytics-config-store/src/stores/analytics-config.ts
```diff
@@ -46,5 +45,6 @@ export const useAnalyticsConfigStore = defineStore('analytics-config', () => {
 
   return {
     getConfig,
+    configResult,
```
adorack, line 48: As discussed in Slack, I still don't think that both of these should be exported.  If you feel strongly about not accessing the ref through a function, we could instead query the endpoint when the store is instantiated and get rid of `getConfig` entirely.

adorack (changes requested):

PR #1312, opened 2024-04-07
fix(entities-plugins): add missing name for credentials

Leopoldthecoder:
# Summary

Adding missing `name` prop for credential metadata.

#### Resources


sumimakito (approved):
Look Good To Makito 👍

PR #1314, opened 2024-04-08
fix(plugin-icon): use default icon when plugin icon name is undefined [KHCP-11396]

vaibhavrajsingh2001:
## Summary

- if the `getPluginIconURL` method is not provided any icon name, default to `missing` icon
- move check for empty icon name from `PluginIcon` component to `getPluginIconURL` method
- only compute and use default icon URL in `PluginIcon` component if the URL passed in `src` attribute was invalid and the `img` tag emitted an error

## Resources

- [Slack thread discussion](https://kongstrong.slack.com/archives/C05UE909XU5/p1712172404522979?thread_ts=1712153356.330669&cid=C05UE909XU5)
- [KHCP-11396](https://konghq.atlassian.net/browse/KHCP-11396?atlOrigin=eyJpIjoiOGQxMTNjOWJjNzQzNDQxNDkwNmUyNDA0ZDI4NjU3NDIiLCJwIjoiaiJ9)

## ToDo

- [x] add unit tests for `getPluginIconURL`

[KHCP-11396]: https://konghq.atlassian.net/browse/KHCP-11396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-plugins/src/definitions/metadata.ts
+++ b/packages/entities/entities-plugins/src/definitions/metadata.ts
@@ -767,6 +767,6 @@ export const CREDENTIAL_SCHEMAS: Record<string, any> = {
  * @returns URL for the plugin icon
  */
 export const getPluginIconURL = (name: string) => {
-  const imageName = PLUGIN_METADATA[name]?.imageName || name
+  const imageName = PLUGIN_METADATA[name]?.imageName || name || 'missing' // ? default icon is 'missing'
   return new URL(`../assets/images/plugin-icons/${imageName}.png`, import.meta.url).href
 }
--- a/packages/entities/entities-plugins/src/definitions/metadata.ts
+++ b/packages/entities/entities-plugins/src/definitions/metadata.ts
@@ -767,6 +767,6 @@ export const CREDENTIAL_SCHEMAS: Record<string, any> = {
  * @returns URL for the plugin icon
  */
 export const getPluginIconURL = (name: string) => {
-  const imageName = PLUGIN_METADATA[name]?.imageName || name || 'missing' // ? default icon is 'missing'
+  const imageName = PLUGIN_METADATA[name]?.imageName || name || 'missing' // default icon is 'missing'
   return new URL(`../assets/images/plugin-icons/${imageName}.png`, import.meta.url).href
 }
--- a/packages/entities/entities-plugins/src/definitions/metadata.ts
+++ b/packages/entities/entities-plugins/src/definitions/metadata.ts
@@ -767,10 +767,10 @@ export const CREDENTIAL_SCHEMAS: Record<string, any> = {
  * @returns URL for the plugin icon
  */
 export const getPluginIconURL = (name: string) => {
-  const imageName = PLUGIN_METADATA[name]?.imageName || name // default icon is 'missing'
+  const imageName = PLUGIN_METADATA[name]?.imageName || name || 'missing' // default icon is 'missing'
   const iconURL = new URL(`../assets/images/plugin-icons/${imageName}.png`, import.meta.url).href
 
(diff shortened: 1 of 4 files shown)


packages/entities/entities-plugins/src/definitions/metadata.ts
```diff
@@ -767,6 +767,6 @@ export const CREDENTIAL_SCHEMAS: Record<string, any> = {
  * @returns URL for the plugin icon
  */
 export const getPluginIconURL = (name: string) => {
-  const imageName = PLUGIN_METADATA[name]?.imageName || name
+  const imageName = PLUGIN_METADATA[name]?.imageName || name || 'missing' // default icon is 'missing'
```
adamdehaven, line 770: Sent you this in chat, but I think you could also do something similar to this:

```typescript
export const getPluginIconURL = (name: string) => {
  const imageName = PLUGIN_METADATA[name]?.imageName || name
  const imageHref = new URL(`../assets/images/plugin-icons/${imageName}.png`, import.meta.url).href

  if (!!imageHref && imageHref.includes('undefined')) {
    return new URL('../assets/images/plugin-icons/missing.png', import.meta.url).href
  }

  return imageHref
}
```

vaibhavrajsingh2001: Implemented in this commit:
https://github.com/Kong/public-ui-components/pull/1314/commits/d306d2dc7c2f81761198e668ab7e1197e19eb92e#diff-eca5dd8c68b49c246b1f9431a95fd878e6e3c2a128cd15b5fdbad4b13f404f5c

vaibhavrajsingh2001: Ommited `!imageHref` check, since `new URL()` will always generate some kind of URL for us, so doing an empty string check won't be useful.

packages/entities/entities-plugins/src/definitions/metadata.ts
```diff
@@ -767,6 +767,12 @@ export const CREDENTIAL_SCHEMAS: Record<string, any> = {
  * @returns URL for the plugin icon
  */
 export const getPluginIconURL = (name: string) => {
-  const imageName = PLUGIN_METADATA[name]?.imageName || name
-  return new URL(`../assets/images/plugin-icons/${imageName}.png`, import.meta.url).href
+  const imageName = PLUGIN_METADATA[name]?.imageName || name // default icon is 'missing'
+  const iconURL = new URL(`../assets/images/plugin-icons/${imageName}.png`, import.meta.url).href
+
+  if (iconURL.endsWith('/undefined.png') || iconURL.endsWith('/undefined')) {
```
adamdehaven, line 773: question: why not use a more generic `iconURL.includes('undefined)` instead of checking for two separate `endsWith` conditions?

vaibhavrajsingh2001: We can have an icon that might have the word undefined in it, e.g. `user-undefined`

adamdehaven: How about this instead then?
```suggestion
  if (iconURL.includes('plugin-icons/undefined')) {
```

vaibhavrajsingh2001: Yeah, this should work for now until we allow custom asset directory as an arg to the method.

adamdehaven: I think just checking for `undefined` in the string is sufficient; we aren't going to have icons with `*undefined*` in the name

kaiarrowood (approved):
LGTM! 🚀 

Nice work @vaibhavrajsingh2001

PR #1315, opened 2024-04-08
feat(analytics): add disabled state support to metric cards

nekolab:
Changes include:
## For `@kong-ui-public/metric-cards`
* Add `containerSubtitle` to `<MetricCardContainer />` props and support subtitle in container

## For `@kong-ui-public/analytics-metric-provider`
* Add `containerSubtitle` and `isAnalyticsEnabled` to `<MetricsProvider />` props
* `<MetricsProvider />` now provides `containerSubtitle` and `isAnalyticsEnabled` to child components
* `description`, `containerTitle`, `containerSubtitle` and `isAnalyticsEnabled` are now provided with reactive data structure
* `<MetricsConsumer />` now has its own padding and has a gary background when analytics is disabled
* Remove `trendRange`, `description` and metrics value, change percentable text to 'N/A' in analytics disabled mode

## For `@kong-ui-public/dashboard-renderer`
* Add `chartSubtitle` and `isAnalyticsEnabled` to both metric card schema and `GoldenSignalsRenderer`
* Add `noPadding` to the tile definition schema to remove the padding. This would be useful while setting the tile's background style

Jira: KM-14

### DIFF
--- a/packages/analytics/analytics-metric-provider/sandbox/App.vue
+++ b/packages/analytics/analytics-metric-provider/sandbox/App.vue
@@ -112,6 +112,7 @@ const globalProviderProps = {
   longCardTitles: false,
   description: 'Generic Description',
   containerTitle: 'Analytics',
+  containerSubtitle: 'Subtitle',
 }
 const globalBridge = makeQueryBridge()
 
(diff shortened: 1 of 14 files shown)


adorack (changes requested):
I ran out of time to review today; I'll try to take a look tomorrow.  I'm not sure whether or not this is how we want to incorporate this functionality, but will try to finish thinking that through.

adorack (changes requested):
Thanks for taking a stab at this.  We talked about it a fair bit internally, and we think it's better for us (data team) to handle the changes to the metric cards and provider.  Not all of the requirements for a disabled state are in the Figma for the GM MFE; we'd like to do this in a way that applies outside the GM MFE as well.

PR #1318, opened 2024-04-09
fix(addUniqueHeadingSlugs): Fix duplicate headings

azhangsg:
# Summary
The issue is here

```
  const slugCount = slugMap.get(text)
  slugMap.set(text, (slugCount || 0) + 1)
  const duplicateSuffix = slugCount && slugCount > 1 ? `-${slugCount}` : ''
```

First, we get `slugCount` which can be either undefined or a number.
Then, we set the slugMap correctly.
But, when we set the suffix, we use the original `slugCount` that hasn't been updated yet, so it's either 1 less than it should be or undefined.

Proposed changes:
```
  const slugCount = (slugMap.get(text) || 0) + 1
  slugMap.set(text, slugCount)
  const duplicateSuffix = slugCount > 1 ? `-${slugCount}` : ''
```

Move the math logic to when we initialize `slugCount`. And then we don't need to check if slugCount exists.

### DIFF
--- a/packages/portal/document-viewer/src/utils/addUniqueHeadingSlugs.ts
+++ b/packages/portal/document-viewer/src/utils/addUniqueHeadingSlugs.ts
@@ -26,9 +26,9 @@ export function addSlug(child: HeadingNode, slugMap: Map<string, number>, prefix
   const text = isTextNode(firstChild)
     ? firstChild.text || defaultText
     : defaultText
-  const slugCount = slugMap.get(text)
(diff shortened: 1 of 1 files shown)


adamdehaven:
@azhangsg thanks for the contribution! I'm incorporating your changes in https://github.com/Kong/public-ui-components/pull/1321 as we need to refactor our CI workflow to negate the need for secrets in running the test pipeline.

PR #1322, opened 2024-04-10
feat(vfg): kongponents usage [khcp-11333]

kaiarrowood:
# Summary

First drop of VFG refactor.
Part of [KHCP-11323](https://konghq.atlassian.net/browse/KHCP-11323).

Preview package with passing plugins e2es: 
https://github.com/Kong/konnect-ui-apps/pull/2968

#### Resources

[KHCP-11323]: https://konghq.atlassian.net/browse/KHCP-11323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


adamdehaven:
Commenting before I forget - you should **absolutely** issue a breaking change commit (if you're unsure how to do so, use the `pnpm commit` command after staging your changes.

packages/core/forms/src/index.ts
```diff
@@ -1,11 +1,13 @@
 import type { App } from 'vue'
 import VueFormGenerator from './generator/FormGenerator.vue'
 import * as sharedForms from './forms'
+import attributesDirective from './directives'
 
 // Export Vue plugin as the default
 export default {
   install: (app: App): void => {
     app.component('VueFormGenerator', VueFormGenerator)
+    app.directive('attributes', attributesDirective)
```
adamdehaven, line 10: question: we're making this available in the entire host app; is that desired?

packages/core/forms/src/composables/useAbstractFields.ts
```diff
@@ -0,0 +1,250 @@
...
+  get as objGet,
+  forEach,
+  isFunction,
+  isString,
+  isArray,
+  debounce,
+  uniqueId,
+  uniq as arrayUniq,
+  type DebouncedFunc,
+} from 'lodash'
```
adamdehaven, line 12: suggestion: just add the individual `lodash.*` packages we need instead of the entire dep

packages/core/forms/src/composables/index.ts
```diff
@@ -0,0 +1,6 @@
+import useAbstractFields from './useAbstractFields'
+
+// All composables must be exported as part of the default object for Cypress test stubs
+export default {
+  useAbstractFields,
```
adamdehaven, line 5: suggestion: since this is the only composable, just import it directly where you need it instead of having this barrel file.

packages/core/forms/src/composables/useAbstractFields.ts
sumimakito: Curious: Would we consider utilizing `loadash-es` in the future?

packages/core/forms/src/generator/fields/core/fieldLabel.vue
```diff
@@ -1,17 +1,51 @@
...
+  },
+  schema: {
+    type: Object as PropType<Record<string, any>>,
+    required: true,
+  },
+  vfg: {
+    type: Object,
+    required: true,
+  },
+})
```
sumimakito, line 38: Curious: Would we prefer `withDefaults(defineProps<T>(), ...)` when there is no need for validators?

packages/core/forms/src/index.ts
sumimakito: It might be better if we could prepend prefixes like `forms` and rename them to a more self-explaining name like: `v-forms-bind-attrs`

packages/core/forms/src/directives.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const attributesDirective: ObjectDirective = {
+  beforeMount(el, binding, vnode) {
+    bindFieldTypeAttrs(el, binding, vnode)
+  },
+  updated(el, binding, vnode) {
+    bindFieldTypeAttrs(el, binding, vnode)
+  },
+  // Vue 3 does not have a "componentUpdated" lifecycle hook equivalent
+}
```
sumimakito, line 29: I tried to find external usages of `v-attribtues` throughout `entities-plugins` and `konnect-ui-apps`, but I only found the following usage:

https://github.com/Kong/public-ui-components/blob/4a9f171020980adf3ead078b4a34fff121044768/packages/entities/entities-plugins/src/composables/useSchemas.ts#L577-L586

It reminds me that we also have `inputAttributes` in our VFG schema, and I wonder if they overlap: 🤔

https://github.com/Kong/public-ui-components/blob/4a9f171020980adf3ead078b4a34fff121044768/packages/entities/entities-plugins/src/definitions/schemas/typedefs.ts#L48-L66

sumimakito:
It also appears that we are not using `@kong-ui-public/forms` as a plugin and thus not calling the `install` function in host apps:

https://github.com/Kong/public-ui-components/blob/4a9f171020980adf3ead078b4a34fff121044768/packages/core/forms/src/index.ts#L8-L11

We would then need to update the app initialization steps in host apps as well as the sandbox:

https://github.com/Kong/public-ui-components/blob/4a9f171020980adf3ead078b4a34fff121044768/packages/core/forms/sandbox/index.ts#L13

with

```ts
import Forms '../src'
app.use(Forms)
```

in order to apply the directive defined via `app.directive`.

packages/core/forms/src/generator/fields/core/fieldInput.vue
```diff
@@ -1,205 +1,268 @@
...
-    :files="schema.files"
-    v-attributes="'input'")
-  span.helper(v-if="schema.inputType.toLowerCase() === 'color' || schema.inputType.toLowerCase() === 'range'") {{ value }}
+<template>
+  <div
+    v-attributes="'wrapper'"
+    class="form-field-wrapper"
+  >
+    <!--
+    TODO: all these attributes are needed??
```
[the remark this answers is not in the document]
kaiarrowood: @sumimakito do you know if this is the case? It seems like we have separate components for all of these "special" input types, so I'm confused why we had all those attributes bound previously.

packages/core/forms/src/composables/index.ts
kaiarrowood: We will eventually have a few others as well, I can still change the imports if that's preferred over the barrel file.

packages/core/forms/src/directives.ts
kaiarrowood: I think I might have been wrong about that overlap. I suggest let's try taking them out. It looks like we may not use it anyway.

packages/core/forms/src/generator/fields/core/fieldInput.vue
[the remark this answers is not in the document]
sumimakito: @kaiarrowood It seems that we are still using `<input />` in fields under src/generator/fields/core, I would also prefer specific components like `<KCheckbox />` to `<input type="checkbox" />` if possible. 

However, I also see that we have a few usages of the generic `input` element, for example:

https://github.com/Kong/public-ui-components/blob/b1e4b310f406f043860dcddffdfcec04fd3da4fe/packages/entities/entities-plugins/src/composables/useSchemas.ts#L439-L448

IMO, the original list of props/attrs in src/generator/fields/core/fieldInput.vue is way too many:

https://github.com/Kong/public-ui-components/blob/b1e4b310f406f043860dcddffdfcec04fd3da4fe/packages/core/forms/src/generator/fields/core/fieldInput.vue#L3-L39

If we use specific components, we can then make this `fieldInput.vue` a field for text only (strings and numbers are okay, but not checkboxes, not files, etc.).

I also don't have a clear idea why we have `fieldInput.vue` that accepts many attributes while we also have more-specific field based on the `input` element like `fieldUpload.vue`.

packages/core/forms/src/composables/useAbstractFields.ts
kaiarrowood: @adamdehaven do you have an opinion on this?

adamdehaven: It looks fine if you want to switch

kaiarrowood: @adamdehaven @sumimakito 

Why are we wanting to switch to `-es` out of curiosity? I originally thought the goal of this comment (and switching to `lodash-es`) was to improve bundle size. The [reading](https://lodash.com/per-method-packages) about the per method packages, it sounds like usage isn't recommended and they'll be removed in v5. Checking about recommended import strategies, [research](https://www.blazemeter.com/blog/import-lodash-libraries#lodash-and-lodash-es-import-comparison-table) seems to show `lodash-es` will increase package size. 

The recommended import strategy for `lodash` with only a few functions is one-by-one imports (ex. `import get from 'lodash/get'`), so I've gone ahead and switched to using the `lodash-es` package with one-by-one import strategy and it seems like everything is working as expected.

Is there another reason for using `lodash-es`?

sumimakito: > Checking about recommended import strategies, [research](https://www.blazemeter.com/blog/import-lodash-libraries#lodash-and-lodash-es-import-comparison-table) seems to show lodash-es will increase package size.

@kaiarrowood That was an interesting finding. I tried to build `lodash` + one-by-one imports / `lodash-es` + `import { ... } from 'lodash-es'` / `lodash-es` + one-by-one imports with Vite and use `rollup-plugin-visualizer` to inspect the bundle size, it turned out that the render sizes of these `lodash-es` builds were still several bytes smaller than the `lodash` one.

I also did a deeper look at the build artifacts of the three versions and suspected the several bytes difference would be due to UMD overheads like `Object.defineProperty(exports, "__esModule", ...)` containing in the `lodash` build. 🤔

adamdehaven (approved):
lgtm

Leopoldthecoder (approved):
LGTM

PR #1330, opened 2024-04-12
feat(analytics): improved dashboard time spec handling [MA-2759]

adorack:
BREAKING CHANGE: refactor interface for analytics config store

- Change analytics config store to fetch config when first instantiated and expose refs rather than functions.
- Update dashboard renderer to fill in a default time spec based on the org's retention if one isn't provided.
- Don't issue queries until a time spec is determined.
- Fix missing feature flag functions in tests.
- Fix display of "description" property in metric cards: show in container rather than broadcasting to all cards.
- Fix title and description reactivity in metrics provider.
- Replace special timeframe token in tile descriptions.
- Fill in missing context values in the top-level renderer to save code in the tile renderers.

# Summary

#### Resources


mihai-peteu (approved):

kaiarrowood (approved):
LGTM! 🚀

PR #1332, opened 2024-04-15
fix(EntityFormSection): use aria labels for fieldsets [KHCP-11030]

vaibhavrajsingh2001:
# Summary

We were using `fieldset` element in `EnitityFormSection` component for a form group, but there was no corresponding `legend` element.
However, we can't use the `legend` element for the form title, since it's deeply nested and is not a direct child of the `fieldset` element ([`legend` need to be direct child of `fieldset`](https://www.accessibility-developer-guide.com/examples/forms/grouping-with-fieldset-legend/#legend-must-be-a-direct-child-of-fieldset)). 
Given how the component was structured, it would be a major breaking change to move the title as the direct child of the `fieldset` element. So instead opted to use ARIA roles ([ref](https://www.w3.org/WAI/tutorials/forms/grouping/#associating-related-controls-with-wai-aria)).
We're generating a unique ID for each `EnitityFormSection` component for the ARIA labels, since there can be multiple instance of the component on the same page.

[LevelAccess ticket](https://kong.hub.essentia11y.com/short-link/asIDs7vakV3cxAjf)
[KHCP-11030](https://konghq.atlassian.net/browse/KHCP-11030)

[KHCP-11030]: https://konghq.atlassian.net/browse/KHCP-11030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-shared/src/components/entity-form-section/EntityFormSection.vue
+++ b/packages/entities/entities-shared/src/components/entity-form-section/EntityFormSection.vue
@@ -1,6 +1,6 @@
 <template>
   <div
-    :aria-labelledby="legendID"
+    :aria-labelledby="title ? legendId : undefined"
     class="kong-ui-entity-form-section"
     :class="{ 'has-divider': hasDivider }"
     role="group"
@@ -14,7 +14,7 @@
         <component
           :is="titleTag"
           v-if="title"
-          :id="legendID"
+          :id="legendId"
           class="form-section-title"
         >
           {{ title }}
@@ -79,7 +79,7 @@ defineProps({
 })
 
 const slots = useSlots()
-const legendID = uuidv4()
+const legendId = uuidv4()
 </script>
 
 <style lang="scss" scoped>
--- a/packages/entities/entities-shared/src/components/entity-form-section/EntityFormSection.vue
+++ b/packages/entities/entities-shared/src/components/entity-form-section/EntityFormSection.vue
@@ -1,7 +1,9 @@
 <template>
-  <fieldset
+  <div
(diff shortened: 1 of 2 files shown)


packages/entities/entities-shared/src/components/entity-form-section/EntityFormSection.vue
```diff
@@ -1,7 +1,9 @@
 <template>
-  <fieldset
+  <div
+    :aria-labelledby="legendID"
```
portikM, line 3:
```suggestion
    :aria-labelledby="title ? legendID : undefined"
```

issue: if title is not provided, it's not gonna be rendered

portikM (commented):
Can you add a note under the `title` prop section in `packages/entities/entities-shared/docs/entity-form-section.md` saying that if title is not provided you must set `aria-label` attribute on the component please?

packages/entities/entities-shared/src/components/entity-form-section/EntityFormSection.vue
```diff
@@ -75,6 +79,7 @@ defineProps({
 })
 
 const slots = useSlots()
+const legendID = uuidv4()
```
portikM, line 82:
```suggestion
const legendId = uuidv4()
```

nit: may be a personal preference, but I prefer every (non-first) word in camel-cased const name to start with a title latter and then continue in lower case (`ID -> Id`)

portikM (approved):
LGTM 🚀

PR #1355, opened 2024-04-20
refactor(entities-routes): make expressions and monaco as optionalDependencies

sumimakito:
# Summary

This PR moves `@kong-ui-public/expressions` to `peerDependencies` so that we can provide `@kong-ui-public/expressions` and `monaco-editor` in host apps. This allows us to reduce the bundle size of `@kong-ui-public/entities-routes` (several megabytes smaller) as the Expressions features are not always necessary (e.g., Konnect API Products/Overview).

In host apps with non-type imports from `@kong-ui-public/entities-routes` but without the need to use Expressions features, they will need to mark these peer dependencies as `external` (as https://github.com/Kong/konnect-ui-apps/pull/3018 demonstrated):

```ts
defineConfig({
  build: {
    rollupOptions: {
      external: ['@kong-ui-public/expressions', 'monaco-editor'],
    },
  },
})
```

I've also updated the documentation for that package to note this.

### DIFF
--- a/packages/entities/entities-routes/README.md
+++ b/packages/entities/entities-routes/README.md
@@ -48,7 +48,7 @@ import '@kong-ui-public/entities-routes/dist/style.css'
 
 This package installs `@kong-ui-public/expressions` and `monaco-editor` to support [Kong's Expressions language](https://docs.konghq.com/gateway/latest/reference/expressions-language/)-related features. However, this will add several megabytes to your bundle size.
 
-If you don't need Expressions features, but have imported the components (e.g., `<RouteForm.vue />`) with these features in your host app, add the following configuration to your Vite config to avoid bundling them with your build result:
+If you don't need Expressions features, but have imported the components with support for these features (e.g., `routeFlavor` in `<RouteForm.vue />`) in your host app, add the following configuration to your Vite config to avoid bundling them with your build result:
 
 > As of now, only `<RouteForm.vue />` has Expressions features. You should ensure you are not using Expressions features in your host app before adding this configuration.
 
--- a/packages/entities/entities-routes/README.md
+++ b/packages/entities/entities-routes/README.md
@@ -15,17 +15,17 @@ Route entity components.
 - `@kong/kongponents` must be added as a dependency in the host application, globally available via the Vue Plugin installation, and the package's style imports must be added in the app entry file. [See here for instructions on installing Kongponents](https://kongponents.konghq.com/#globally-install-all-kongponents).
 - `@kong-ui-public/i18n` must be available as a `dependency` in the host application.
 - `axios` must be installed as a dependency in the host application
-- `@kong-ui-public/expressions` and `monaco-editor` must be installed as **`devDependencies`**. If you don't need to support [Kong's Expressions language](https://docs.konghq.com/gateway/latest/reference/expressions-language/)-related features, you will need to add the following configuration to your Vite config to avoid them to be bundled with your build result:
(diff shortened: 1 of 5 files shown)


Leopoldthecoder (dismissed):
LGTM

sumimakito:
Update: It seems that the host app does not need to add `@kong-ui-public/expressions` and `monaco-editor` in Rollup's `external`. I've removed that instruction from the readme doc.

Leopoldthecoder (dismissed):

packages/entities/entities-routes/README.md
```diff
@@ -15,6 +15,17 @@ Route entity components.
...
+
+  ```ts
+  defineConfig({
+    build: {
+      rollupOptions: {
+        external: ['@kong-ui-public/expressions', 'monaco-editor'],
+      },
+    },
+  })
+  ```
```
adamdehaven, line 28: Why would a host app need to add this to the config? By adding to externals here in this repo, the packages will not be bundled, meaning they don't exist in the host app regardless. 

If there are imports from these packages here, they would need to be dependencies in the host app; I'm not sure it makes sense to "skip" these imports by telling the host app not to resolve them?

sumimakito: As `entities-routes` contains components that conditionally import from `@kong-ui-public/expressions` (Expressions features). For example:

https://github.com/Kong/public-ui-components/blob/c7993636c84a3efebdbc3a38ba9e246b8bd37432/packages/entities/entities-routes/src/components/RouteFormExpressionsEditorLoader.vue#L62-L70

… which is used here:

https://github.com/Kong/public-ui-components/blob/c7993636c84a3efebdbc3a38ba9e246b8bd37432/packages/entities/entities-routes/src/components/RouteForm.vue#L338-L362

(Note: the Expressions features are not enabled by default because we want to keep the default behavior unchanged as much as possible)

However, if we import `RouteForm` in the host app but don't want to use the Expressions feature (setting `routeFlavors` without `expressions`) provided by `RouteForm`, `@kong-ui-public/expressions` and `monaco-editor` will still be bundled. For instance, in Konnect UI's `overview` app, we import `RouteForm`, but if we don't pass `routeFlavors` with `expressions` to it, it should not enable/use Expressions features. Specifying `@kong-ui-public/expressions` and `monaco-editor` in the `devDependencies` of the `overview` app still makes them be bundled, and will increase the bundle size by ~4MB:

Screenshot 2024-04-22 at 20 43 32

Therefore, I would suggest the host apps using components (we only have `RouteForm` for now) that may import `@kong-ui-public/expressions` to mark them as `external` in their Vite config.

I hope this explanation helps you understand better.

sumimakito: Good idea. I've moved them to `optionalDependencies` and updated the docs for `entities-routes`. Would it look better this time?

packages/entities/entities-routes/README.md
```diff
@@ -43,6 +44,24 @@ import { RouteList, RouteForm, RouteConfigCard } from '@kong-ui-public/entities-
...
+> As of now, only `<RouteForm.vue />` has Expressions features. You should ensure you are not using Expressions features in your host app before adding this configuration.
+
+```ts
+defineConfig({
+  build: {
+    rollupOptions: {
+      external: ['@kong-ui-public/expressions', 'monaco-editor'],
+    },
+  },
+})
```
adamdehaven, line 62: I still don't think this is needed even with the optional dependencies; the host app isn't going to install them by default.

sumimakito: You are right. I've updated the documentation.

PR #1363, opened 2024-04-22
chore: Add type for combined config and CP config endpoint

alxgrant:
Add new type for endpoint in Kanalytics that returns the combined config and CP config to be used by the King service.


adorack (changes requested):
Hmm.  I don't think we need this type in public-ui, and I think that it probably shouldn't return a list -- moving discussion to a thread...

PR #1370, opened 2024-04-25
fix(entities-routes): add prop to hide traditional columns

sumimakito:
# Summary

Adding a prop to allow hiding traditional columns in the route list.


Leopoldthecoder (approved):
LGTM

PR #1373, opened 2024-04-26
fix(vfg): input model reactivity

kaiarrowood:
# Summary

Watch for programmatic changes to model.

#### Resources


packages/core/forms/src/composables/useAbstractFields.ts
```diff
@@ -75,6 +75,19 @@ export default function useAbstractFields(formData: {
     },
   })
 
+  // watch for programmatic changes to the model
+  watch(() => formData.model, (newVal, oldVal) => {
```
adamdehaven, line 79: question: since `formData.model` is an object, I imagine you should add `deep: true` to the watcher?

Leopoldthecoder:
LGTM. Locally tested and problem solved.

Leopoldthecoder (approved):

PR #1375, opened 2024-04-28
fix(entities-plugins): omit redis cluster/sentinel addresses if empty [KAG-4328]

sumimakito:
# Summary

This PR fixed an issue where plugins with `config.redis.cluster_addresses` and `config.redis.sentinel_addresses` cannot be created or edited if the user tried to add some addresses but removed all addresses before submitting the form.

This bug affects the following four plugins:

- graphql-proxy-cache-advanced
- graphql-rate-limiting-advanced
- proxy-cache-advanced
- rate-limiting-advanced

This fix should be a temporary and straightforward workaround since we should ultimately fix it in the `core/forms` package.

KAG-4328

### DIFF
--- a/packages/entities/entities-plugins/package.json
+++ b/packages/entities/entities-plugins/package.json
@@ -35,6 +35,7 @@
     "@kong-ui-public/i18n": "workspace:^",
     "@kong/design-tokens": "1.12.11",
     "@kong/kongponents": "9.0.0-alpha.146",
+    "@types/lodash-es": "^4.17.12",
     "axios": "^1.6.8",
     "vue": "^3.4.21",
     "vue-router": "^4.3.0"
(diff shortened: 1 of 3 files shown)


Leopoldthecoder:
Note: after 3.7, the correct fix would be to change [this line](https://github.com/Kong/public-ui-components/blob/10efd2db06c0e521a2a86d7b1f7cb8a779e91aac/packages/core/forms/src/generator/fields/advanced/FieldArray.vue#L180) to
```ts
this.value = this.value.filter((_, i) => i !== index)
```

This is because `splice` does not trigger the [setter](https://github.com/Kong/public-ui-components/blob/10efd2db06c0e521a2a86d7b1f7cb8a779e91aac/packages/core/forms/src/generator/fields/abstractField.ts#L66-L75) of computed `value`.

Leopoldthecoder (approved):

PR #1384, opened 2024-05-02
fix(entities-plugins): fix consumer jwt rsa pub key input field [KHCP-11775]

dustinryerson:
# Summary

**Ticket:**
https://konghq.atlassian.net/browse/KHCP-11775

**Changes:**
- Updates the RSA public key input field on the consumer JWT credential form to be a textarea instead of a regular text input field

**Screenshots:**
**Old form**
Screenshot 2024-05-03 at 9 45 42 AM

**New form**
Screenshot 2024-05-03 at 9 46 53 AM

#### Resources

### DIFF
--- a/packages/entities/entities-plugins/src/components/PluginForm.vue
+++ b/packages/entities/entities-plugins/src/components/PluginForm.vue
@@ -685,7 +685,7 @@ const buildFormSchema = (parentKey: string, response: Record<string, any>, initi
         }, {})
     }
 
-    if (treatAsCredential.value && props.config.app === 'kongManager' && credentialSchema) {
+    if (treatAsCredential.value && credentialSchema) {
       for (let i = 0; i < credentialSchema.length; i++) {
         if (credentialSchema[i][field]) {
           initialFormSchema[field] = {


Leopoldthecoder:
Hi @dustinryerson, I feel the right fix for this would be to update [this line](https://github.com/Kong/public-ui-components/blob/e5dcba72ae6bb9025a0e9132dc4f150e2b50aaa5/packages/entities/entities-plugins/src/components/PluginForm.vue#L688) to
```diff
-    if (treatAsCredential.value && props.config.app === 'kongManager' && credentialSchema) {
+    if (treatAsCredential.value && credentialSchema) {
```

Reason: we take two steps to generate schema:
1. Fetch schema from the backend.
  a. Konnect backend does not provide credentials schema, so we have them hardcoded [here].(https://github.com/Kong/public-ui-components/tree/main/packages/entities/entities-plugins/src/definitions/schemas/credentials)
2. Merge the backend schema with some frontend overrides. This is done in [`buildFormSchema`](https://github.com/Kong/public-ui-components/blob/e5dcba72ae6bb9025a0e9132dc4f150e2b50aaa5/packages/entities/entities-plugins/src/components/PluginForm.vue#L465-L761).

Step two should apply to both KM and Konnect, so we can remove the `props.config.app === 'kongManager'` condition.

`packages/entities/entities-plugins/src/definitions/schemas/credentials/mockedJwtSchema.json` needs to stay the same as what the actual backend schema looks like (if there is one), meaning its `"type"` will not be `"textArea"`. And then in `packages/entities/entities-plugins/src/components/PluginForm.vue` we don't have to handle the situation where `scheme.type === 'textArea'`.

So could you revert your changes and verify if removing `props.config.app === 'kongManager'` is good enough? Thanks!

Note: after removing `props.config.app === 'kongManager'` you'll notice that the `rsa_public_key` field is gone. This is because it needs the algorithm to be `ES*` or `RS*` to show as defined [here](https://github.com/Kong/public-ui-components/blob/e5dcba72ae6bb9025a0e9132dc4f150e2b50aaa5/packages/entities/entities-plugins/src/definitions/schemas/JWT.ts#L40-L42)

dustinryerson:
> Hi @dustinryerson, I feel the right fix for this would be to update [this line](https://github.com/Kong/public-ui-components/blob/e5dcba72ae6bb9025a0e9132dc4f150e2b50aaa5/packages/entities/entities-plugins/src/components/PluginForm.vue#L688) to
> 
> ```diff
> -    if (treatAsCredential.value && props.config.app === 'kongManager' && credentialSchema) {
> +    if (treatAsCredential.value && credentialSchema) {
> ```
> 
> Reason: we take two steps to generate schema:
> 
> 1. Fetch schema from the backend.
>    a. Konnect backend does not provide credentials schema, so we have them hardcoded [here](https://github.com/Kong/public-ui-components/tree/main/packages/entities/entities-plugins/src/definitions/schemas/credentials).
> 2. Merge the backend schema with some frontend overrides. This is done in [`buildFormSchema`](https://github.com/Kong/public-ui-components/blob/e5dcba72ae6bb9025a0e9132dc4f150e2b50aaa5/packages/entities/entities-plugins/src/components/PluginForm.vue#L465-L761).
> 
> Step two should apply to both KM and Konnect, so we can remove the `props.config.app === 'kongManager'` condition.
> 
> `packages/entities/entities-plugins/src/definitions/schemas/credentials/mockedJwtSchema.json` needs to stay the same as what the actual backend schema looks like (if there is one), meaning its `"type"` will not be `"textArea"`. And then in `packages/entities/entities-plugins/src/components/PluginForm.vue` we don't have to handle the situation where `scheme.type === 'textArea'`.
> 
> So could you revert your changes and verify if removing `props.config.app === 'kongManager'` is good enough? Thanks!
> 
> Note: after removing `props.config.app === 'kongManager'` you'll notice that the `rsa_public_key` field is gone. This is because it needs the algorithm to be `ES*` or `RS*` to show as defined [here](https://github.com/Kong/public-ui-components/blob/e5dcba72ae6bb9025a0e9132dc4f150e2b50aaa5/packages/entities/entities-plugins/src/definitions/schemas/JWT.ts#L40-L42).

@Leopoldthecoder this was super helpful information, I tested your suggested fix and it worked perfectly. Thank you!

adamdehaven (approved):

PR #1397, opened 2024-05-10
fix(gateway-services): fix gateway service tags input [KHCP-11838]

dustinryerson:
# Summary

**Ticket:**
https://konghq.atlassian.net/browse/KHCP-11838

**Changes**
Fixes the Gateway Service tags input field to be consistent with other tags input fields. Other tag input fields allow input such as `tag1, tag2, tag3`, but the Gateway Service input field currently only allows input without spaces, i.e. `tag1,tag2,tag3`. This PR should fix this.

#### Resources


mptap (approved):
Lgtm!

PR #1400, opened 2024-05-13
fix(*): kongponents alpha phase 13 [KHCP-11882]

portikM:
# Summary

Jira: https://konghq.atlassian.net/browse/KHCP-11882

Phase 13 adoption of components reskinned in alpha version of Kongponents v9 release. Includes these components:
- `KPop`

#### Resources


packages/portal/swagger-ui-web-component/package.json
```diff
@@ -14,8 +14,8 @@
   },
   "dependencies": {
     "@kong/swagger-ui-kong-theme-universal": "^4.3.3",
-    "react": "17.0.2",
-    "react-dom": "17.0.2",
+    "react": "18.3.1",
+    "react-dom": "18.3.1",
```
portikM, line 18: @davidma415 I had to bump these packages to fix build error: https://github.com/Kong/public-ui-components/actions/runs/9099234899/job/25011319267
Not sure if this is ideal

adamdehaven: No I don't think this is ideal; if we bump the major, I believe it would be a major version bump for our package as well.

It was building fine before this PR, what changed that now causes it to fail and require a version bump?

adamdehaven (changes requested):
(blocking merge for discussion on react version bump)

PR #1429, opened 2024-06-03
feat(analytics-utilities): export country name mapping [MA-2937]

mihai-peteu:
# Summary

https://konghq.atlassian.net/browse/MA-2937

#### Resources


packages/analytics/analytics-utilities/src/country-codes.ts
```diff
@@ -0,0 +1,250 @@
+// ISO 3166-1 alpha-2 country codes to 3-digit UN country code mapping
+export const COUNTRY_CODES = {
+  4: { code: 'AF', name: 'Afghanistan' },
```
adamdehaven, line 3: question: I see you sorted the codes by country name instead of the base key (the integer)

Is this because that leading integer (e.g. `4` on this line) is needed to link with your API calls or something?

mihai-peteu: Correct, this comes in handy here:
https://github.com/Kong/khcp-ui/pull/7975/files#diff-5a10c8944faa74aabe61ff3d16237598b88f5f3737579393c56635d9022359a7R410-R412

We may not need this file once the kanalytics api picks up changes in @voigtjr 's Alpha-2 update PR, so I'll pause on this for now.
https://github.com/Kong/flink-jobs/pull/793

mihai-peteu:
Mapping is no longer necessary, closing.

PR #1430, opened 2024-06-04
feat: geo map component

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-2941

image

image

#### Resources


packages/analytics/analytics-geo-map/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "errorLimit": "200KB"
+  },
+  "peerDependencies": {
+    "@kong/kongponents": "9.0.0-alpha.162",
+    "maplibre-gl": "^4.3.2",
+    "vue": "^3.4.27"
+  },
+  "dependencies": {
+    "@kong-ui-public/analytics-utilities": "workspace:^",
+    "@types/geojson": "^7946.0.14"
```
adamdehaven, line 71: nit: types should be a dev dependency

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
```diff
@@ -0,0 +1,244 @@
+<template>
+  <div
+    id="mapContainer"
```
adamdehaven, line 3: issue: you’re using a Vue ref for an ID attribute. 

I would think you’d want to exclude ID in case there could be multiple instances on the same page?

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
```diff
@@ -0,0 +1,244 @@
...
+  },
+  fitToCountry: {
+    type: String,
+    required: false,
+    default: '',
+  },
+  initialZoom: {
+    type: Number,
+    required: false,
+    default: null,
```
adamdehaven, line 40: if the type is a number, the default should be a number

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
```diff
@@ -0,0 +1,244 @@
...
+    validator: (value: number) => value >= 0 && value <= 24,
+  },
+  metricUnit: {
```
adamdehaven, line 43: question: should this prop have a validator or is literally any string acceptable?

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
```diff
@@ -0,0 +1,244 @@
...
+        getColor(metric),
+      ])
+      : ['#ccc', '#ccc']),
+    '#ccc', // default color if no match
+  ] as DataDrivenPropertyValueSpecification<ColorSpecification>,
+  'fill-opacity': 0.7,
+}))
+
+const getColor = (metric: number) => {
+  if (metric > 80) return '#296378'
```
adamdehaven, line 69: would a switch be more readable?

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
```diff
@@ -0,0 +1,244 @@
...
+  const flat: (number | number[])[] = position.flat(1)
+
+  if (Array.isArray(flat[0]) && Array.isArray(flat[0][0])) {
+    return flattenPositions(flat)
+  }
+
+  return flat as number[][]
+}
+
+const goToCountry = (countryCode: string) => {
```
adamdehaven, line 87: I think the countryCode should be typed more specific if only those few values are allowed

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
```diff
@@ -0,0 +1,244 @@
...
+    return
+  } else if (countryCode === 'NO') {
+    map.value?.fitBounds([
+      [4, 57],
+      [32, 71],
+    ])
+    return
+  }
+
+  const found = (props.geoJsonData as FeatureCollection).features.find((f: Feature) => f.properties?.ISO_A2 === countryCode)
```
adamdehaven, line 116: Couple things:

1. Can you type the prop instead of force casting it?
2. Can you type the `found` variable? I think it’s a boolean
3. Should it be computed? If the prop changes after this is initialized it will be static

packages/analytics/analytics-geo-map/src/index.ts
```diff
@@ -0,0 +1,17 @@
+// import type { App } from 'vue'
+import AnalyticsGeoMap from './components/AnalyticsGeoMap.vue'
+
+// Export Vue plugin
```
adamdehaven, line 4: nit: you can delete all this plugin init code

packages/analytics/analytics-geo-map/src/locales/en.json
```diff
@@ -0,0 +1,6 @@
+{
+  "header": {
```
adamdehaven, line 2: Do you not need i18n? If no? You can remove this

packages/analytics/analytics-geo-map/README.md
```diff
@@ -0,0 +1,105 @@
...
+- **type:** `String`
+- **required:** `false`
+- **default:** `''`
+- **description:** Country code to fit the map bounds to. Overrides initial center and zoom.
+
+#### `initialZoom`
+
+- **type:** `Number`
+- **required:** `false`
+- **default:** `null`
```
mihai-peteu, line 69: If not required, we should fall back to a certain value. A value closer to zero maybe?

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
mihai-peteu: Darn, you beat me to it... Maybe a value closer to zero, for a broader map view.

filipgutica: I need a way to tell that a user did not provide an initial zoom. Can I type this as `Number | null` ?

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
mihai-peteu: Whether `id` and/or `data-testid` - we should probably allow the consuming app to be in charge of generating unique ids

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
adamdehaven: I believe so yes

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
adamdehaven: Edit: well it’s actually a string, but it’s confusing because it matches the ref name 😅

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
```diff
@@ -0,0 +1,244 @@
...
+      closeOnClick: false,
+    })
+
+    map.value?.on('mousemove', 'countries-layer', (e) => {
+      const feature = e.features?.[0]
+      if (feature) {
+        const { ISO_A2, ADMIN } = feature.properties
+        const metric = props.countryMetrics[ISO_A2]
+        if (metric !== undefined) {
+          popup.setLngLat(e.lngLat).setHTML(`<strong>${ADMIN}</strong>: ${metric} ${props.metricUnit}`).addTo(map.value as Map)
```
mihai-peteu, line 185: non-blocking, just a thought. Since `setHTML` doesn't sanitize, hoping we can create a custom plugin here like we did for ChartJS.

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
filipgutica: done

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
filipgutica: Changed this completely need a re-review

packages/analytics/analytics-geo-map/README.md
filipgutica: made this Number | null

packages/analytics/analytics-geo-map/src/locales/en.json
filipgutica: added it

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
filipgutica: done

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
filipgutica: any string is not acceptable, it should be a number between 0 and 24 or null if the prop is not provided.

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
filipgutica: done

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
filipgutica: maplibre requires an id. The `ref` was not necessary

packages/analytics/analytics-geo-map/package.json
filipgutica: done

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
filipgutica: > non-blocking, just a thought. Since `setHTML` doesn't sanitize, hoping we can create a custom plugin here like we did for ChartJS.

This is maplibre's `setHTML` fn: https://maplibre.org/maplibre-gl-js/docs/API/classes/Popup/#sethtml 

We can followup on this later. values templated into the html (metric and metricUnit) are well typed and validated, so for simplicity we can just use maplibre's api for displaying a popup until we need something more "custom"

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
adamdehaven: Then the I’d should be required by the host app. If it’s built-in to the component you could end up with multiple on the same page

packages/analytics/analytics-geo-map/src/index.ts
filipgutica: done

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
```diff
@@ -0,0 +1,377 @@
...
+})
+
+</script>
+
+<style lang="scss" scoped>
+.kong-ui-public-analytics-geo-map {
+  height: 100%;
+  position: relative;
+  width: 100%;
+
```
mihai-peteu, line 333: Do you mind adding this to hide the MapLibre attribution?

```
&:deep(.maplibregl-control-container) {
    .maplibregl-ctrl-attrib {
      display: none;
    }
  }
 ```

mihai-peteu (commented):
Non-blocking, just looks cleaner without the attribution

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
filipgutica: done

mihai-peteu (approved):

mihai-peteu (approved):

PR #1433, opened 2024-06-06
fix(entities-plugins): hide ordering for consumer group plugins

Leopoldthecoder:
# Summary

Plugins that are scoped to a consumer group do not support dynamic ordering either.

#### Resources


nekolab (approved):
LGTM

PR #1450, opened 2024-06-12
feat(analytics): add support for basic datasource [MA-2931]

adorack:
BREAKING CHANGE: change signature of `queryFn` in query bridge, and make `datasource` a required property in dashboard definitions.

- Change signature of `queryFn`.
- Add `datasource` parameter to MetricsProvider.
- Add `datasource` property to dashboard query definitions.

# Summary

#### Resources

### DIFF
--- a/packages/analytics/analytics-metric-provider/src/components/MetricsProvider.vue
+++ b/packages/analytics/analytics-metric-provider/src/components/MetricsProvider.vue
@@ -10,7 +10,7 @@ import {
   type AnalyticsBridge,
   type ExploreFilter,
   queryableExploreDimensions,
-  type QueryableExploreDimensions,
+  type QueryableExploreDimensions, type QueryDatasource,
   type Timeframe,
 } from '@kong-ui-public/analytics-utilities'
 import { TimePeriods, TimeframeKeys } from '@kong-ui-public/analytics-utilities'
@@ -19,6 +19,7 @@ import { INJECT_QUERY_PROVIDER, DEFAULT_REFRESH_INTERVAL } from '../constants'
 import { useAnalyticsConfigStore } from '@kong-ui-public/analytics-config-store'
 
 const props = withDefaults(defineProps<{
+  datasource?: QueryDatasource
   maxTimeframe?: TimeframeKeys,
   overrideTimeframe?: Timeframe,
   tz?: string,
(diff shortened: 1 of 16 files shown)


packages/analytics/analytics-metric-provider/src/components/MetricsProvider.vue
```diff
@@ -10,7 +10,7 @@ import {
   type AnalyticsBridge,
   type ExploreFilter,
   queryableExploreDimensions,
-  type QueryableExploreDimensions,
+  type QueryableExploreDimensions, type QueryDatasource,
```
hfukada, line 13: minor lint, new line probably preferred?

packages/analytics/analytics-metric-provider/src/components/MetricsProvider.vue
```diff
@@ -85,17 +88,34 @@ const tz = computed(() => {
...
+  // The feature flag client is guaranteed to be initialized by the time the code gets to this place.
+  return evaluateFeatureFlagFn('MA-2527-analytics-sku-config-endpoint', false)
+})
+
+const resolvedDatasource = computed<QueryDatasource>(() => {
+  if (props.datasource) {
+    return props.datasource
+  }
+
+  return skuFeatureFlag.value ? 'basic' : 'advanced'
```
hfukada, line 101: just commenting out loud: 

If datasource is set, use that datasource. 

if it's not set, if the skuFlag is off, default to advanced. else true -> roll basic.

:+1:

packages/analytics/dashboard-renderer/src/components/QueryDataProvider.vue
```diff
@@ -76,13 +81,31 @@ const { data: v4Data, error, isValidating } = useSWRV(queryKey, async () => {
...
+    } = props.query
+
+    if (!datasource && queryBridge) {
+      // TODO(MA-2987): Remove this.
+      const analyticsSKU = queryBridge.evaluateFeatureFlagFn('MA-2527-analytics-sku-config-endpoint', false)
+
+      datasource = analyticsSKU ? 'basic' : 'advanced'
+    }
+
+    // TODO: similar to other places, consider adding a type guard to ensure the query
```
hfukada, line 96: technically this is enforced at the API level so it's luckily not too bad.

hfukada (approved):

packages/analytics/analytics-metric-provider/src/components/MetricsProvider.vue
adorack: Thanks -- not sure why that isn't caught by the linter.  I'd like to merge now so that I can keep rolling with the others, but I'll fix this when I fix the dashboard types later on.

PR #1475, opened 2024-07-04
feat(entities): preserve one column in entity lists

Leopoldthecoder:
# Summary

Always display one of the columns in entity lists.

KM-289

#### Resources


sumimakito (approved):
Verified entity lists in all affected packages. LGTM

PR #1490, opened 2024-07-10
feat(analytics): ai explore types

hfukada:
- Implement necessary types to build out the new experience

### DIFF
--- a/packages/analytics/analytics-utilities/src/types/analytics-data.ts
+++ b/packages/analytics/analytics-utilities/src/types/analytics-data.ts
@@ -1,4 +1,4 @@
-import type { ExploreAggregations } from './explore-v4'
+import type { ExploreAggregations, AllAggregations } from './explore-v4'
 
 export interface RecordEvent {
   [field: string]: string | number | null
@@ -15,7 +15,7 @@ export interface DimensionMap {
 }
 
 export type MetricUnit = {
-  [metricName in ExploreAggregations]?: string
+  [metricName in AllAggregations]?: string
 }
 
 /**
--- a/packages/analytics/analytics-utilities/src/types/explore-v4.ts
+++ b/packages/analytics/analytics-utilities/src/types/explore-v4.ts
@@ -1,7 +1,7 @@
 // Query types
 import type { MetricUnit, RecordEvent } from './analytics-data'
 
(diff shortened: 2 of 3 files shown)


packages/analytics/analytics-utilities/src/types/explore-v4.ts
```diff
@@ -52,6 +72,15 @@ export const basicExploreAggregations = [
 
 export type BasicExploreAggregations = typeof basicExploreAggregations[number]
 
+export const aiExploreAggregations = [
+  'total_tokens',
+  'request_tokens',
+  'response_tokens',
+  'ai_request_count',
+]
```
adorack, line 80:
```suggestion
] as const
```

(Without this, Typescript can't infer things properly.)

adorack (changes requested):
Could you also please update `query-bridge.ts` to add an `AiDatasourceQuery` type to the union there?

packages/analytics/analytics-utilities/src/types/analytics-data.ts
```diff
@@ -18,6 +18,10 @@ export type MetricUnit = {
   [metricName in ExploreAggregations]?: string
 }
 
+export type AiMetricUnit = {
```
adorack, line 21: I think there should be a type that represents a superset of all possible aggregations.  Something like AllAggregations = ExploreAggregations | AiExploreAggregations.  Then drive MetricUnit off of that (instead of adding AiMetricUnit, and then type metric_names and metric_units off of the AllAggregations type.

Logically, I think that makes the most sense: AnalyticsChart needs to be able to handle all possible aggregations -- not just explore aggregations.

packages/analytics/analytics-utilities/src/types/explore-v4.ts
```diff
@@ -159,7 +196,7 @@ export interface QueryResponseMeta {
   start_ms: number
   end_ms: number
   display: DisplayBlob
-  metric_names?: ExploreAggregations[]
+  metric_names?: ExploreAggregations[] | AiExploreAggregations[]
```
adorack, line 199: I think this should be `AllAggregations[]`.  Otherwise, I think it's looking good...

adorack (approved):

PR #1499, opened 2024-07-16
fix(plugins): responsiveness improvements [KHCP-12616]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-12616

Plugin select grid view improvements

Changes:
* remove `pluginsPerRow` prop - instead rely on grid to display as many plugins per row as it can fit and make everything below the first row collapsible
* plugin select view loading screen improvement (screenshots)

Before
Screenshot 2024-07-17 at 10 38 44 AM

After
Screenshot 2024-07-17 at 10 35 08 AM

#### Resources


Leopoldthecoder:
I didn't look at the code changes closely, but could you mark this as a breaking change since we are removing an existing prop `pluginsPerRow`?

portikM:
> I didn't look at the code changes closely, but could you mark this as a breaking change since we are removing an existing prop `pluginsPerRow`?

Yeah, it's still in draft mode, I gotta fix couple of things before it's gonna be ready for review but that's what I was gonna do tomorrow 🙂

Leopoldthecoder:
Since this is still a draft, maybe you already know this (but I'll post it here anyway): we only invoke `setToggleVisibility` in the `onMounted` hook, so when we resize the screen, the visibility of the `View less` and `View more` buttons won't be updated:

https://github.com/user-attachments/assets/9589b488-716b-4fe8-b168-b05162715454

portikM:
> Since this is still a draft, maybe you already know this (but I'll post it here anyway): we only invoke `setToggleVisibility` in the `onMounted` hook, so when we resize the screen, the visibility of the `View less` and `View more` buttons won't be updated:
> 
>  Kapture.2024-07-17.at.13.33.54.mp4

Fixed now, thanks for pointing out!

packages/entities/entities-plugins/src/components/custom-plugins/PluginCustomGrid.vue
```diff
@@ -26,20 +26,28 @@
...
       <!-- don't display a trigger if all plugins will already be visible -->
       <template
-        v-if="modifiedCustomPlugins.length <= pluginsPerRow"
+        v-if="!showCollapseTrigger"
         #trigger
       >
         &nbsp;
       </template>
+
+      <!-- not actually using KCollapse to hide/show content here, just using it for the trigger, title and container -->
```
adamdehaven, line 39: suggestion: do we really need all of these comments in the templates? These render in the DOM and are visible to customers.

portikM: IMO we're not exposing anything we wouldn't want to expose in this comment

adamdehaven: Right but I’m not sure it’s necessary. No need to expose implementation details in the HTML when you can add comments in the script section if needed

kaiarrowood (approved):
LGTM! 🚀

PR #1503, opened 2024-07-20
feat(*): vault secret picker [KM-307]

sumimakito:
# Summary

### core/forms

- [x] Autofill for fields in VFG
- [x] Remove <del>`showRemoveButton`</del> (It turned out this is no longer used after searching in the codebase)
- [x] Use backend schema for OIDC (verified)

### entities/entities-plugins

- [x] Schema builder adoption
- [x] Autofill adoption
- [x] Tests

### entities/entities-vaults

- [x] Export endpoints

#### useDebouncedFilter

- [x] Reactivity fix
- [x] Support exact matching on the `key` field
- [x] Use cached response when the query is empty after the first fetch

KM-307


packages/entities/entities-plugins/src/components/PluginForm.vue
```diff
@@ -545,6 +555,8 @@ const buildFormSchema = (parentKey: string, response: Record<string, any>, initi
     initialFormSchema[field] = { id: field, model: key } // each field's key will be set as the id
     initialFormSchema[field].type = scheme.type === 'boolean' ? 'checkbox' : 'input'
     initialFormSchema[field].required = scheme.required
+    initialFormSchema[field].values = scheme.values
```
Leopoldthecoder, line 558: Question: what is `values`?

packages/core/forms/src/components/fields/FieldArrayItem.vue
```diff
@@ -1,13 +1,18 @@
...
-  <div class="array-item">
-    <slot />
-    <KButton
-      appearance="tertiary"
-      class="delete"
-      @click="$emit('remove-item')"
-    >
-      <TrashIcon />
-    </KButton>
+  <div class="array-item-wrapper">
```
Leopoldthecoder, line 2: Is this new wrapper necessary? I don't know if this will break any style since the DOM structure is changed.

sumimakito: Good catch. I feel it is redundant after some experiments. I've removed it.

packages/entities/entities-plugins/src/components/PluginForm.vue
sumimakito: Good question. This is to pass the `values` attribute on a `map` field to the field in the generated schema which allows the autofill providers to consume the attributes inside.

For example, with the `http-log` plugin, we have the following schema for the `headers` field:

```json
{
  "headers": {
    "description": "...",
    "type": "map",
    "values": {
      "referenceable": true,
      "type": "string"
    },
    "keys": "..."
  }
}
```

Since we seem to only have `values` for `map` fields now, I've refactored the code and added some comments to ensure this is easy to understand.

Leopoldthecoder (dismissed):

Leopoldthecoder (approved):

PR #1508, opened 2024-07-23
feat(entities-vaults): disable secret key input when editing

Leopoldthecoder:
# Summary

The `key` field in the secret edit form should be disabled.

KM-343

#### Resources


sumimakito (approved):
LGTM and thanks

PR #1509, opened 2024-07-23
fix(entities): create button visibility

Leopoldthecoder:
# Summary

Hide the create button initially and only show it when the table has data.

KM-342

#### Resources


nekolab (approved):
LGTM

PR #1513, opened 2024-07-24
fix(entities-plugins): allow ACL to be scoped to consumers and consumer groups

sumimakito:
# Summary

Allowing the ACL plugin to be scoped to consumers and consumer groups.

Note: We should also update the schema for this plugin at [Kong/kong](https://github.com/Kong/kong/blob/b5716b1155044a79fceaa4b10a8349643c7a8d34/kong/plugins/acl/schema.lua#L7) and [Kong/kong-ee](https://github.com/Kong/kong-ee/blob/c3273ad1f067b8ddc567a3b435a9494e4a528964/kong/plugins/acl/schema.lua#L14-L15).

KM-350


sumimakito:
Closing as this is not a bug as per our discussion on Slack

PR #1525, opened 2024-08-01
fix(vaultform): fix compatibility of STS endpoint url in Konnect

TT1228:
# Summary

Fix compatibility of STS endpoint url field in AWS vault form when saving in Konnect


Leopoldthecoder (approved):
LGTM

PR #1526, opened 2024-08-01
refactor(cypress): better typings for `cy.mount` and `cy.get('@vueWrapper')`

nekolab:
This PR improves the typings for `cy.mount` and `cy.get('@vueWrapper')`, making it more accurate and able to detect the wrong props usage.

As cy.get('@vueWrapper') is widely used across the project, the chainable typing is specificed to reflect the real value inside the wrapper.

Some obvious wrong props are fixed accordingly in this PR

JIRA: KM-366


cypress/support/component.ts
```diff
@@ -1,21 +0,0 @@
...
-
-declare global {
-  namespace Cypress {
-    interface Chainable {
-      mount: typeof mount
-    }
-  }
-}
-
-Cypress.Commands.add('mount', mount)
```
adamdehaven, line 18: question: why was this removed?

packages/entities/entities-certificates/src/components/CACertificateList.cy.ts
```diff
@@ -73,10 +73,10 @@ describe('<CACertificateList />', () => {
         props: {
           cacheIdentifier: `ca-certificate-list-${uuidv4()}`,
           config: baseConfigKonnect,
-          canCreate: () => {},
-          canEdit: () => {},
-          canDelete: () => {},
-          canRetrieve: () => {},
+          canCreate: () => true,
```
adamdehaven, line 76: question: you're changing a ton of defaults in tests - is this desired? I would imagine this would impact the pass/fail

cypress/support/component.ts
nekolab: Same as the file in `shared-ui-components`. It's
- not imported by other modules, so the `Cypress.Commands.add('mount', mount)` is not executed
- the `declare global {` now conflicts with the declarations in `cypress/support/index.d.ts`
- the `Cypress.Commands.add('mount', mount)` (if it's executed), will conflict with `cypress/support/commands.ts:32`

adamdehaven (approved):

packages/entities/entities-certificates/src/components/CACertificateList.cy.ts
nekolab: Oops, I think we should change it to `() => false` here.
Most of the `canXXX`s are declared as `() => boolean | Promise<boolean>`. Before we improved the typing support for`cy.mount`, it could not detect the wrong props type so passing `() => {}` didn't trigger any error.
We should change it here but it seems like changing it to `() => false` is safer than changing it to `() => true`

adamdehaven (approved):

cypress/support/component.ts
adamdehaven: @nekolab I forgot to respond here: Cypress automatically looks for this file (it's the default configuration).

In this repo, since we weren't really adding anything here, there's no harm so far in removing; however, in `konnect-ui-apps` for example, we have an entire component test configuration that differs from the e2e test setup, so the `component.ts` files there must remain.

nekolab: Thanks for the explanation! I think thats also why we have `cy.baseMount` in `konnect-ui-apps`.

PR #1528, opened 2024-08-02
fix(entities-certificates): remove snis if field not displayed

Leopoldthecoder:
# Summary

The request body should include the `snis` key only when the `snis` field is present.

#### Resources


TT1228 (approved):
LGTM and verified it in sandbox, the fix works

PR #1533, opened 2024-08-06
test(vfg): add field tester and convert FormCheckbox [khcp-11331]

kaiarrowood:
# Summary

Fixes:
- Convert `FormCheckbox` to use Kongponents - [KHCP-11331](https://konghq.atlassian.net/browse/KHCP-11331)
- Vue Form Generator testing improvements
  - Add a field tester component to normalize cypress component testing and use it for `FieldInput` and `FieldCheckbox`

image

#### Resources

[KHCP-11331]: https://konghq.atlassian.net/browse/KHCP-11331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/forms/src/index.ts
```diff
@@ -10,7 +11,7 @@ export default {
 }
 
 export { customFields } from './components/fields/exports'
-export { VueFormGenerator, sharedForms }
+export { VueFormGenerator, sharedForms, FieldTester }
```
adamdehaven, line 14: question: why are we exporting the `FieldTester` component? Should it only be used in the tests?

packages/core/forms/src/components/fields/__tests__/FieldTester.vue
```diff
@@ -0,0 +1,149 @@
...
+  updatedFormModel.value[key] = model
+}
+
+// verify programmatic updates to model
+const handleUpdate = (): void => {
+  updatedFormModel.value = props.modifiedModel
+}
+</script>
+
+<style lang="scss" scoped>
```
aanchalm01, line 108: Is there a reason for not using design tokens here ?

packages/core/forms/src/components/fields/__tests__/FieldInput.cy.ts
```diff
@@ -0,0 +1,116 @@
+// Import types for custom commands
+/// <reference types="../../../../../cypress/support" />
```
aanchalm01, line 2: nit: remove comment

packages/core/forms/src/index.ts
kaiarrowood: I want to use it the sandbox (for debug) and the tests, but couldn't figure out another way without ending up with 2 copies.

packages/core/forms/src/components/fields/__tests__/FieldTester.vue
kaiarrowood: Mostly because I was lazy and it's an internal-only tool. I'll go back and clean it up 👍🏻

packages/core/forms/src/components/fields/__tests__/FieldInput.cy.ts
kaiarrowood: I never know when we need these or not

packages/core/forms/src/index.ts
adamdehaven: Just import it in the tests by relative path; you don't want to export it or even import it at the package root.

aanchalm01 (approved):
lgtm!

PR #1534, opened 2024-08-06
fix(app-layout): add aria-current to sidebar links [KHCP-12832]

portikM:
# Summary

Add `aria-current` to sidebar links

#### Resources

### DIFF
--- a/packages/core/app-layout/src/components/sidebar/SidebarItem.vue
+++ b/packages/core/app-layout/src/components/sidebar/SidebarItem.vue
@@ -6,13 +6,12 @@
     <component
       :is="!useAnchorTag ? 'router-link' : 'div'"
       v-slot="slotProps"
-      :aria-current="!useAnchorTag ? item.active : undefined"
+      :aria-current="item.active ? 'page' : undefined"
       :custom="!useAnchorTag ? true : undefined"
       :to="!useAnchorTag && !openInNewWindow ? item.to : undefined"
     >
       <a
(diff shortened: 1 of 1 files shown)


packages/core/app-layout/src/components/sidebar/SidebarItem.vue
```diff
@@ -6,11 +6,13 @@
     <component
       :is="!useAnchorTag ? 'router-link' : 'div'"
       v-slot="slotProps"
+      :aria-current="!useAnchorTag ? item.active : undefined"
```
adamdehaven, line 9: this line is invalid; regardless whether `useAnchorTag` is true or false, the interior `a` tag is always the actual link element. This line would produce multiple `aria-current` attributes in some scenarios

adamdehaven (changes requested):

packages/core/app-layout/src/components/sidebar/SidebarItem.vue
portikM: Ohhh I didn't realize that. In that case, when `useAnchorTag` is `true` I think we should bind `aria-current` ONLY to the interior a element (since the outer element is gonna be a `div`). When `useAnchorTag` is `false`, we should ONLY bind it to the outer element (`router-link`). Do you agree?

adamdehaven: I think it can always go on the external element since that is the element that contains the "active" styles; however, it should be set to `aria-current="page"` instead of `aria-current="true"`

portikM: Agree, pushed the change

adamdehaven (approved):

PR #1541, opened 2024-08-08
refactor(entities-shared): simplify logic in useDebouncedFilter

sumimakito:
# Summary

This pull request simplifies the logic flow in `useDebouncedFilter`

KM-393

### DIFF
--- a/packages/entities/entities-shared/src/composables/useDebouncedFilter.ts
+++ b/packages/entities/entities-shared/src/composables/useDebouncedFilter.ts
@@ -86,15 +86,17 @@ export default function useDebouncedFilter(
     // using this to skip unnecessary fetch fired on focus
     if (previousQuery.value === query) {
       return
-    } else if (query === '') {
-      // use cached results if query is empty
-      results.value = resultsCache.value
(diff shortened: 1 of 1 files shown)


packages/entities/entities-shared/src/composables/useDebouncedFilter.ts
```diff
@@ -105,9 +107,7 @@ export default function useDebouncedFilter(
         if (config.app === 'konnect') { // KoKo only supports exact match
           // If user has typed info in the query field
           let currUrl = url.value + '' // clone
-          if (query) {
-            currUrl += `/${query}`
-          }
+          currUrl += `/${query}`
```
Leopoldthecoder, line 110: Can we combine these 2 lines to
```js
const currUrl = `${url.value}/${query}`
```

sumimakito: You are right. I also moved the line to update `previousQuery` upward.

Leopoldthecoder (approved):
LGTM

PR #1354, opened 2024-04-19
chore(renovate): pin monaco-editor `expressions` and `entities-routes`

sumimakito:
# Summary

This PR tries to pin `monaco-editor` by excluding it for `@kong-ui-public/expressions` and `@kong-ui-public/entities-routes` via `packageRules` in `renovate.json`.

### Why?

`monaco-editor` is currently only used for the Expressions routes support in `@kong-ui-public/expressions` and `@kong-ui-public/entities-routes`. We encountered several issues while consuming the latest `monaco-editor` in the host app during the development. To deliver the Expressions routes support in time, we decided to downgrade it to [`v0.21.3`–which we used previously in Kong Manager](https://github.com/Kong/kong-admin/blob/9bae765787d9ea28cfa48e16bd3c56b65e82a2c7/package.json#L85) to prevent [the Renovate bot from bumping it](https://github.com/Kong/public-ui-components/pull/1352). 

We will try to update it to the latest possible minor release after Kong Gateway 3.7's release.

PR #1542, opened 2024-08-08
chore(deps): revert pinning monaco in core/expressions

sumimakito:
# Summary

> Need thorough tests in host apps

This pull request reverts:

- https://github.com/Kong/public-ui-components/pull/1354

The reason why we decided to pin `monaco-editor` for `core/expressions` was because the latest version by that time was causing errors. However, it seems we cannot reproduce those errors anymore with the latest version as of today. This pull request is to revert the previous change that pinned `monaco-editor`.

KM-390

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -5,13 +5,7 @@
   "ignorePaths": [
     "packages/core/cli/src/__template__/package.json"
   ],
-  "packageRules": [
-    {
-      "description": "Pin monaco-editor for @kong-ui-public/expressions and @kong-ui-public/entities-routes",
-      "matchFileNames": ["packages/core/expressions/package.json", "packages/entities/entities-routes/package.json"],
-      "groupName": "expressions-monaco-editor",
-      "matchPackageNames": ["monaco-editor"],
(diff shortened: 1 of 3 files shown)


renovate.json
```diff
@@ -5,13 +5,7 @@
...
-    {
-      "description": "Pin monaco-editor for @kong-ui-public/expressions and @kong-ui-public/entities-routes",
-      "matchFileNames": ["packages/core/expressions/package.json", "packages/entities/entities-routes/package.json"],
-      "groupName": "expressions-monaco-editor",
-      "matchPackageNames": ["monaco-editor"],
-      "enabled": false
-    }
-  ]
+  "constraints": {
+    "pnpm": "8.6.11"
```
adamdehaven, line 9: issue: this repository uses pnpm v9. I believe this constraint block should now be removed.

adamdehaven (changes requested):

Leopoldthecoder (approved):

adamdehaven (approved):

PR #1547, opened 2024-08-09
fix(analytics): improve dashboard renderer types [MA-2708]

adorack:
Overall goal: improve alignment between the dashboard definitions and the queries that our API accepts.

- Separate queryable and filterable dimensions.
- Strip filters that aren't relevant to the current datasource.
- Remove types and documentation associated with obsolete versions of Explore.
- Separate datasource type definitions into multiple files.
- Remove obsolete helpers.
- Update dashboard renderer sandbox to remove type casts which are no longer needed.
- Separate out advanced and basic schemas for dashboard renderer.

# Summary

#### Resources


filipgutica (changes requested):
I'd like to review

filipgutica (approved):

PR #1551, opened 2024-08-12
fix(expressions): router playground known issues [KM-391]

2eha0:
# Summary
* HTTP headers in requests appeared as [object Object] in the exported JSON
* Support HTTP headers with multiple values (should add context value per header value)

[KM-391](https://konghq.atlassian.net/browse/KM-391?atlOrigin=eyJpIjoiMzc3NTQ1MDAzMWVhNDRjZDllZDFkNDA4OGM4Mjg0MDIiLCJwIjoiaiJ9)

[KM-391]: https://konghq.atlassian.net/browse/KM-391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/expressions/src/utils.ts
+++ b/packages/core/expressions/src/utils.ts
@@ -59,4 +59,14 @@ export const transformCheckRequest = (request: Partial<Request>): string | undef
       return i18n.t('errors.methodShouldCapitalized')
     }
   }
+
+  if (request.headers) {
+    Object.keys(request.headers).forEach(key => {
+      const values = request.headers![key]
+      // transform header values to array
+      if (typeof values === 'string' && values.indexOf(',') !== -1) {
+        request.headers![key] = values.split(',').map(value => value.trim())
+      }
+    })
+  }
 }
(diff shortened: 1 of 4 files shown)


packages/core/expressions/src/utils.ts
```diff
@@ -59,4 +59,14 @@ export const transformCheckRequest = (request: Partial<Request>): string | undef
       return i18n.t('errors.methodShouldCapitalized')
     }
   }
+
+  if (request.headers) {
+    Object.keys(request.headers).forEach(key => {
+      const values = request.headers![key]
+      // transform header values to array
+      if (typeof values === 'string' && values.indexOf(',') !== -1) {
+        request.headers![key] = values.split(',').map(value => value.trim())
```
sumimakito, line 68: Should we filter out empty values from the header's value set? 

e.g. Now: `a,b,c,` → `["a", "b", "c", ""]`

Screenshot 2024-08-15 at 11 27 09

sumimakito (approved):
Did functionality test and LGTM

packages/core/expressions/src/utils.ts
2eha0: I don’t think so. Making it strict would require handling too many cases, which isn’t necessary and meaningful.

PR #1552, opened 2024-08-12
feat: add new filter fields

alxgrant:
# Summary
Include new `field` in filter object and support for new dimensions and metric metrics. 

Examples: 
```
        "filters": [
          {
            "type": ">=",
            "field": "latencies_response_ms",
            "value": 5000
          },
          {
            "type": "!=",
            "field": "latencies_response_ms",
            "value": null
          },
          {
            "type": "in",
            "field": "route",
            "values": ["id"]
          },
          {
            "type": "not_in",
            "dimension": "service",
            "values": ["id"]
          }

        ],
```

#### Resources


packages/analytics/analytics-utilities/src/types/explore/basic.ts
```diff
@@ -22,8 +22,9 @@ export type FilterableBasicExploreDimensions = typeof filterableBasicExploreDime
 
 export interface BasicExploreFilter {
   type: ExploreFilterTypesV2
-  dimension: FilterableBasicExploreDimensions
-  values: (string | number | null)[]
+  dimension?: FilterableBasicExploreDimensions
+  field?: FilterableBasicExploreDimensions
+  values?: (string | number | null)[]
```
alxgrant, line 27: Maybe making this optional is weird, but it's optional in advanced for metric filters when we expect `value` instead

adorack (changes requested):
I'd like to review.

packages/analytics/analytics-utilities/src/types/explore/basic.ts
hfukada: would it makes sense to make a separate Filter type with the exact members and then do the union type as a list for `filters`? cc @adorack 

like. 
```
export interface MetricFilter {
   metric:
   value:
   type: <metricFilterType>
}

AnyFilter = MetricFilter | BasicExploreFilter
```

then in Explore interface have...? :
```
  filters: AnyFilter[],
```

adorack (approved):

PR #1555, opened 2024-08-14
fix(entitites-consumer-groups): filter response transform [KM-410]

TT1228:
# Summary

When apply filter in consumer group list in Konnect, response from backend is not tailored for the table, so this change is made to adapt that format.

KM-410

### DIFF
--- a/packages/entities/entities-shared/src/composables/useFetcher.ts
+++ b/packages/entities/entities-shared/src/composables/useFetcher.ts
@@ -58,11 +58,7 @@ export default function useFetcher(
 
 
       if (data[dataKey]) {
-        if (Array.isArray(data[dataKey])) {
-          tableData = data[dataKey]
-        } else {
-          tableData = [data[dataKey]]
-        }
+        tableData = Array.isArray(data[dataKey]) ? data[dataKey] : [data[dataKey]]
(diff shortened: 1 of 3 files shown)


packages/entities/entities-shared/src/composables/useFetcher.ts
```diff
@@ -54,8 +56,13 @@ export default function useFetcher(
...
 
-      if (data[dataKey] && Array.isArray(data[dataKey])) {
-        tableData = data[dataKey]
+
+      if (data[dataKey]) {
+        if (Array.isArray(data[dataKey])) {
+          tableData = data[dataKey]
+        } else {
+          tableData = [data[dataKey]]
+        }
```
Leopoldthecoder, line 65: Do you think this is better?
```suggestion
        tableData = Array.isArray(data[dataKey]) ? data[dataKey] : [data[dataKey]]
```

Leopoldthecoder:
nit: I suggest changing the PR title to `fix(entitites-consumer-groups): filter response transform [KM-410]`.

Leopoldthecoder (approved):
Manually tested with Konnect draft PR and LGTM.

PR #1566, opened 2024-08-16
feat(analytics-sku): remove feature flag mentions [MA-3136]

mihai-peteu:
# Summary

#### Resources


adorack (changes requested):
I'd like to review.

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.cy.ts
```diff
@@ -508,21 +498,21 @@ describe('<DashboardRenderer />', () => {
...
       global: {
         provide: {
-          [INJECT_QUERY_PROVIDER]: mockQueryProvider({ skuFeatureFlag: true }),
+          [INJECT_QUERY_PROVIDER]: mockQueryProvider({ }),
         },
       },
     }).then(() => {
       // Extra calls may mean we mistakenly issued queries before knowing the timeSpec.
-      cy.get('@fetcher').should('have.callCount', 5)
+      cy.get('@fetcher').should('have.callCount', 2)
```
mihai-peteu, line 506: @adorack if I run this test with `.only`, it indeed intercepts this call five times.

But if i run in sequence, the `fetcher` only gets called twice 🤷

packages/analytics/analytics-metric-provider/src/components/MetricsProvider.vue
```diff
@@ -59,24 +59,21 @@ if (props.dimension && queryableExploreDimensions.findIndex(x => x === props.dim
...
 } else {
   queryFn = queryBridge.queryFn
-  evaluateFeatureFlagFn = queryBridge.evaluateFeatureFlagFn
 }
 
 const analyticsConfigStore = useAnalyticsConfigStore()
 
 // Check if the current org has long enough retention to make a sane trend query.
 // If the feature flag is set, trend access is always true.
-const hasTrendAccess = computed<boolean>(() => skuFeatureFlag.value || analyticsConfigStore.longRetention)
```
adorack, line 79: This should now always return true.  I'm thinking that it probably isn't straightforward to carry this whole thing throughout the metric cards, and we probably don't strictly need to, so it's probably fine to just return computed true.

packages/analytics/analytics-metric-provider/src/components/MetricsProvider.vue
```diff
@@ -113,10 +106,6 @@ const timeframe = computed<Timeframe>(() => {
     return props.overrideTimeframe
   }
 
-  if (skuFeatureFlag.value) {
```
adorack, line 116: Pretend this is always true.  (In this case, the function basically becomes `return props.overrideTimeframe || TimePeriods.get(TimeframeKeys.SEVEN_DAY)!`.)

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.cy.ts
```diff
@@ -99,23 +98,14 @@ describe('<DashboardRenderer />', () => {
         },
       }
 
-      console.log('Resolving config')
-      return Promise.resolve(config)
-    }
-
-    // @ts-ignore: TS doesn't infer things correctly.  NoInfer may help.
-    const evaluateFeatureFlagFn: AnalyticsBridge['evaluateFeatureFlagFn'] = (key) => {
```
adorack, line 107: I think we still want this stub; it should just return true in all circumstances.

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.cy.ts
adorack: I can look at this after the other issues are resolved.

adorack (changes requested):

packages/analytics/analytics-metric-provider/src/components/MetricsProvider.cy.ts
```diff
@@ -284,27 +253,8 @@ describe('<AnalyticsMetricProvider />', () => {
     cy.get('.metricscard').eq(0).find('.metricscard-trend-range').should('contain', 'vs previous 24 hours')
   })
 
-  it('displays "7 days" if the feature flag is set', () => {
```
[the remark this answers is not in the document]
mihai-peteu: renamed it to `displays "7 days" by default`

packages/analytics/analytics-metric-provider/src/components/MetricsProvider.cy.ts
```diff
@@ -454,11 +360,10 @@ describe('<AnalyticsMetricProvider />', () => {
     cy.get('.metricscard-trend-change > div').eq(2).should('have.text', '49.98%')
   })
 
-  it('displays single-entity metrics with no trend', () => {
+  it.skip('displays single-entity metrics with no trend', () => {
```
mihai-peteu, line 363: Fine to drop this one, since we are now hardcoding `hasTrendAccess` to `true`?

mihai-peteu:
@adorack updated.  

- Skipped a test in `MetricsProvider.cy.ts` that can probably be removed (?)
- Skipped a test in `DashboardRenderer.cy.ts` which runs fine on its own, but fails when run in sequence:

`it.only`
Screenshot 2024-08-20 at 16 05 40

vs run as part of full suite:
Screenshot 2024-08-20 at 16 05 59

packages/analytics/analytics-metric-provider/src/components/MetricsProvider.cy.ts
adorack: Yes, I think it's fine to drop this one.

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.cy.ts
```diff
@@ -99,23 +98,19 @@ describe('<DashboardRenderer />', () => {
...
-        return opts?.skuFeatureFlag ?? true
-      }
-
       return true
     }
 
     return {
       queryFn: cy.spy(queryFn).as('fetcher'),
       configFn,
-      evaluateFeatureFlagFn,
```
adorack, line 118: This still needs to be part of the mock bridge.

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.cy.ts
```diff
@@ -480,12 +476,12 @@ describe('<DashboardRenderer />', () => {
...
       global: {
         provide: {
-          [INJECT_QUERY_PROVIDER]: mockQueryProvider({ shortRetention: true, skuFeatureFlag: false }),
+          [INJECT_QUERY_PROVIDER]: mockQueryProvider({ shortRetention: true }),
         },
       },
     }).then(() => {
       // Extra calls may mean we mistakenly issued queries before knowing the timeSpec.
-      cy.get('@fetcher').should('have.callCount', 5)
+      cy.get('@fetcher').should('have.callCount', 2)
```
adorack, line 484: I'll help take a look at this one.

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.cy.ts
```diff
@@ -495,7 +491,7 @@ describe('<DashboardRenderer />', () => {
     })
   })
 
-  it('picks 7 days and basic datasource if FF is enabled', () => {
+  it.skip('picks 7 days and basic datasource', () => {
```
adorack, line 494: Why is this one skipped?

adorack (changes requested):

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.cy.ts
mihai-peteu: No particular reason, other than it was failing before. Passing now :)

adorack (approved):
I pushed a commit and am happy with how it ended up.

filipgutica (approved):

PR #1568, opened 2024-08-19
fix: service delete description correction [KM-425]

TT1228:
# Summary

Correct service delete description

KM-425


Leopoldthecoder (approved):
LGTM

PR #1571, opened 2024-08-20
feat: add more queryable dims [MA-3128]

mihai-peteu:
# Summary

Adding in the new queryable dimensions from https://konghq.atlassian.net/browse/MA-3128

#### Resources


adorack (changes requested):
This shouldn't go into `queryableExplore`, it should go into the `requests.ts` file.

PR #1572, opened 2024-08-21
chore(expressions): remove @kong-ui-public/forms from peerDependencies

2eha0:
# Summary
Since `@kong-ui-public/forms` moved into `dependencies`, should not keep it in `peerDependencies`


nekolab (approved):
LGTM as `"@kong-ui-public/forms"` is already in dependencies

adamdehaven (changes requested):
These changes are being discussed and should go through Core UI

PR #1578, opened 2024-08-22
fix(spec-renderer): fix TypeError xe.tag.toLowerCase is not a function [KHCP-12420]

mptap:
Fixes https://konghq.atlassian.net/browse/KHCP-12420

### DIFF
--- a/packages/portal/spec-renderer/src/components/SpecOperationsList.vue
+++ b/packages/portal/spec-renderer/src/components/SpecOperationsList.vue
@@ -175,7 +175,7 @@ const props = defineProps({
     default: (({ items, query }) => {
       query = query.toLowerCase()
 
-      return items.filter((item) => item.tag && item.tag.toLowerCase().includes(query))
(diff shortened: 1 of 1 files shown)


packages/portal/spec-renderer/src/components/SpecOperationsList.vue
```diff
@@ -175,7 +175,7 @@ const props = defineProps({
     default: (({ items, query }) => {
       query = query.toLowerCase()
 
-      return items.filter((item) => item.tag && item.tag.toLowerCase().includes(query))
+      return items.filter((item) => item.tag && String(item.tag).toLowerCase().includes(query))
```
kaiarrowood, line 178:
```suggestion
      return items.filter((item) => item.tag && String(item.tag || '').toLowerCase().includes(query))
```

question: can `item.tag` ever be `undefined` or `null`, if so, recommend falling back to empty string.

kaiarrowood (approved):
LGTM! 🚀

PR #1585, opened 2024-08-27
feat(ai): ai related new metrics, columns

hfukada:
Addings new metrics for ai explore and columns for api-requests

### DIFF
--- a/packages/analytics/analytics-utilities/src/types/explore/requests.ts
+++ b/packages/analytics/analytics-utilities/src/types/explore/requests.ts
@@ -36,6 +36,7 @@ export const queryableRequestMetrics = [
   'latencies_kong_gateway_ms',
   'response_body_size',
   'request_body_size',
+  'ai_count',
 ] as const
 
 export type QueryableRequestMetrics = typeof queryableRequestMetrics[number]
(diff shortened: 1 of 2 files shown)


packages/analytics/analytics-utilities/src/types/explore/requests.ts
```diff
@@ -36,6 +36,7 @@ export const queryableRequestMetrics = [
   'latencies_kong_gateway_ms',
   'response_body_size',
   'request_body_size',
+  'ai_count',
```
filipgutica, line 39: Shouldn't this go in the "filterable" list if we want to filter on it?

filipgutica (changes requested):

packages/analytics/analytics-utilities/src/types/explore/requests.ts
hfukada: it does doesn't it? 

```
export const filterableRequestMetrics = makeFilterable(queryableRequestMetrics)

export type FilterableRequestMetrics = typeof filterableRequestMetrics[number]

export interface RequestDimensionFilter extends Omit<ExploreFilter, 'dimension'> {
  dimension?: FilterableRequestDimensions
  field?: FilterableRequestDimensions
}

export interface RequestMetricFilter {
  type: MetricFilterTypesV2
  field: FilterableRequestMetrics
  value: number | null
}

export type RequestFilter = RequestDimensionFilter | RequestMetricFilter

```
it ends up in the `RequestMetricFilter` interface

filipgutica: Ohh I see. Nevermind

filipgutica (approved):

PR #1589, opened 2024-08-29
fix: more vite.config cleanups

ValeryG:
# Summary

removing more entities covered in shared

### DIFF
--- a/packages/analytics/analytics-geo-map/vite.config.ts
+++ b/packages/analytics/analytics-geo-map/vite.config.ts
@@ -18,11 +18,11 @@ const config = mergeConfig(sharedViteConfig, defineConfig({
     },
     rollupOptions: {
       // Make sure to externalize deps that shouldn't be bundled into your library
-      external: ['maplibre-gl', '@kong-ui-public/i18n'],
+      external: ['maplibre-gl'],
       output: {
         // Provide global variables to use in the UMD build for externalized deps
         globals: {
-          '@kong-ui-public/i18n': 'kong-ui-public-i18n',
+          'maplibre-gl': 'maplibre-gl',
         },
       },
     },
--- a/packages/entities/entities-routes/vite.config.ts
+++ b/packages/entities/entities-routes/vite.config.ts
@@ -23,6 +23,14 @@ const config = mergeConfig(sharedViteConfig, defineConfig({
         '@kong-ui-public/expressions/dist/style.css',
         'lodash.isequal',
       ],
+      output: {
+        // Provide global variables to use in the UMD build for externalized deps
+        globals: {
+          '@kong-ui-public/expressions': '@kong-ui-public/expressions',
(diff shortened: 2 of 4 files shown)


packages/entities/entities-routes/vite.config.ts
```diff
@@ -23,6 +23,14 @@ const config = mergeConfig(sharedViteConfig, defineConfig({
         '@kong-ui-public/expressions/dist/style.css',
         'lodash.isequal',
       ],
+      output: {
+        // Provide global variables to use in the UMD build for externalized deps
+        globals: {
+          '@kong-ui-public/expressions': '@kong-ui-public/expressions',
+          '@kong-ui-public/expressions/dist/style.css': '@kong-ui-public/expressions/dist/style.css',
+          'lodash.isequal': 'lodash.isequal',
```
adamdehaven, line 31: You don't need `globals` for explicit style paths as they aren't imported via JS.

For the package names (the ones that we own) the format changes for the exposed global like this: 
```suggestion
          '@kong-ui-public/expressions': 'kong-ui-public-expressions',
          'lodash.isequal': 'lodash.isequal',
```

Here's an example of how the exposed global string is generated: https://github.com/Kong/public-ui-components/blob/6c9bb045f8e52a490333a1ab9b0066845aaefff1/packages/entities/entities-certificates/vite.config.ts#L15

packages/analytics/analytics-geo-map/vite.config.ts
```diff
@@ -18,11 +18,11 @@ const config = mergeConfig(sharedViteConfig, defineConfig({
     },
     rollupOptions: {
       // Make sure to externalize deps that shouldn't be bundled into your library
-      external: ['maplibre-gl', '@kong-ui-public/i18n'],
+      external: ['maplibre-gl'],
       output: {
         // Provide global variables to use in the UMD build for externalized deps
         globals: {
-          '@kong-ui-public/i18n': 'kong-ui-public-i18n',
+          'maplibre-gl': 'maplibre-gl',
```
adamdehaven, line 25: is this how maplibre-gl exposes their global in the browser?

adamdehaven: The syntax here is:

```
'actual package name': 'package name variable as exposed on globalThis'
```

So as an example, we expose `@kong/icons` as `KongIcons` in globals (you can see the declaration [here](https://github.com/Kong/icons/blob/dc5fc0bab7eddd5301ca9367e0565b9a479e4592/vite.config.ts#L45)).

So you would need to look up how `maplibre-gl` exposes its global variable in the browser, which I believe is located [here](https://github.com/maplibre/maplibre-gl-js/blob/ba966429637d7ffa2dd7296b8a99cea0bc38e686/rollup.config.csp.ts#L17)

So the correct entry in our repo would look like this
```suggestion
          'maplibre-gl': 'maplibregl',
```

adamdehaven (approved):

PR #1591, opened 2024-09-05
fix(entities): max width of tooltips in entity forms

Leopoldthecoder:
# Summary

- Add `maxWidth` to all label tooltips
- Unify the max width to 400px

#### Resources


nekolab (approved):
LGTM

PR #1605, opened 2024-09-10
fix(entities): toolbar create button size

Leopoldthecoder:
# Summary

This PR uses `useActionOutside` to decide the size of the create entity button in the toolbar, so that:
1. the create entity button size aligns with the learning hub icon *in Gateway Manager*
2. the create entity button size aligns with the column toggle icon *in Kong Manager*

This PR also removes `use-action-outside` on `EntityBaseTable` because `EntityBaseTable` does not have this prop.

| Gateway Manager | Kong Manager |
| - | - |
| image | image |

#### Resources


Justineo (approved):

Justineo:
On a second thought, should we move the `<Teleport>` into `EntityBaseTable` and make the `use-action-outside` prop just work, instead of duplicating `<Teleport>` everywhere?

PR #1627, opened 2024-09-13
feat(entities-vaults): force delete secrets

Leopoldthecoder:
# Summary

When deleting a config store, make sure its secrets are also deleted.

#### Resources

### DIFF
--- a/packages/entities/entities-vaults/src/vaults-endpoints.ts
+++ b/packages/entities/entities-vaults/src/vaults-endpoints.ts
@@ -6,7 +6,7 @@ export default {
   list: {
     konnect: {
       getAll: `${konnectBaseApiUrl}/vaults`,
-      deleteConfigStore: `${konnectConfigStoreApiUrl}/{id}`,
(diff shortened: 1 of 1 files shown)


TT1228 (approved):

nekolab (approved):

packages/entities/entities-vaults/src/vaults-endpoints.ts
```diff
@@ -6,7 +6,7 @@ export default {
   list: {
     konnect: {
       getAll: `${konnectBaseApiUrl}/vaults`,
-      deleteConfigStore: `${konnectConfigStoreApiUrl}/{id}`,
+      deleteConfigStore: `${konnectConfigStoreApiUrl}/{id}?force=true`,
```
Justineo, line 9: Is there a chance we may need to use this API without the `force` param?

Leopoldthecoder: With the current design, no: if a config store is deleted, all its secrets become orphans and cannot be accessed anymore, so it's safe to delete them as well.

PR #1633, opened 2024-09-17
fix(forms): emit model update events for array fields

sumimakito:
# Summary

Non-required string fields were not omitted when empty

KM-514


Leopoldthecoder:
Fixed in https://github.com/Kong/public-ui-components/pull/1634

PR #1652, opened 2024-09-24
fix: update dimension consts

filipgutica:
# Summary

Further categorize dimension based on what their intended behaviour in filters will be. 

Request dimensions should be broken down into 3 categories: 
- Fetchable dimensions. These are dimensions that require a fetch to the backend in order to populate filter values and only from those items can filters be created. User added values are not permitted.
- Searchable local dimensions. These are dimensions for which values are populated by a local list and locally searchable. User added values are permitted.
- Text based dimensions. These require a free form input for which the user will input a filter value, or an empty multiselect which users are to add values to.

Explore dimensions are broken down to 
- Fetchable dimensions. These are dimensions that require a fetch to the backend in order to populate filter values and only from those items can filters be created. User added values are not permitted.
- Searchable local dimensions. These are dimensions for which values are populated by a local list and locally searchable. User added values are permitted.

#### Resources

### DIFF
--- a/packages/analytics/analytics-utilities/src/types/explore/requests.ts
+++ b/packages/analytics/analytics-utilities/src/types/explore/requests.ts
@@ -28,18 +28,78 @@ export const queryableRequestDimensions = [
   'service_port',
   'service_protocol',
   'sse',
+  'status_code',
   'status_code_grouped',
   'time',
   'upstream_uri',
+  'upstream_status_code',
   'upstream_status_code_grouped',
   'websocket',
 ] as const
 
 export type QueryableRequestDimensions = typeof queryableRequestDimensions[number]
 
-export const filterableRequestDimensions = makeFilterable(queryableRequestDimensions)
+// These are dimensions that can be fetched/searched via KSearch/KAnalytics/KHCP
+export const fetchableRequestDimensions = [
+  'api_product',
+  'api_product_version',
+  'application',
+  'consumer',
+  'consumer_group',
+  'control_plane',
+  'control_plane_group',
+  'data_plane_node',
+  'gateway_service',
+  'route',
+] as const satisfies QueryableRequestDimensions[]
 
-export type FilterableRequestDimensions = typeof filterableRequestDimensions[number]
+export type FetchableRequestDimensions = typeof fetchableRequestDimensions[number]
+
+// These are dimensions that can be searched locally (i.e. in-memory) in the frontend
+// Will have a pre-populated list that can be searched from defined in the host app
+export const searchableLocalRequestDimensions = [
+  'auth_type',
+  'country_code',
+  'iso_code',
+  'status_code',
+  'upstream_status_code',
+  'status_code_grouped',
+  'upstream_status_code_grouped',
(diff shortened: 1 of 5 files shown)


packages/analytics/analytics-utilities/src/types/explore/requests.ts
```diff
@@ -28,18 +28,78 @@ export const queryableRequestDimensions = [
...
+  'data_plane_node',
+  'gateway_service',
+  'route',
+] as const satisfies QueryableRequestDimensions[]
+
+export type FetchableRequestDimensions = typeof fetchableRequestDimensions[number]
+
+// These are dimensions that can be searched locally (i.e. in-memory) in the frontend
+// Will have a pre-populated list that can be searched from defined in the host app
+export const searchableLocalRequestDimensions = [
```
alxgrant, line 60: I think we will include `response_source` in this list

packages/analytics/analytics-utilities/src/types/explore/requests.ts
```diff
@@ -28,18 +28,78 @@ export const queryableRequestDimensions = [
   'service_port',
   'service_protocol',
   'sse',
+  'status_code',
```
alxgrant, line 31: What does including `status_code` here help with? We expect `status_code` to be a number value in the API, but now it's being included in ways that expect a string filter value like the `RequestEqualsFilter`.

adorack (changes requested):
I'd like to review (once I get caught up).

packages/analytics/analytics-utilities/src/types/explore/requests.ts
alxgrant: Ah nvm I see you had a comment in slack about this and I see now how the dimensions are used below with the local searchable list.

packages/analytics/analytics-utilities/src/types/explore/requests.ts
filipgutica: Are we sure that it's a searchable one, not a text based one, is it something that will have a pre-populated list to select from?

packages/analytics/analytics-utilities/src/types/explore/requests.ts
filipgutica: It behaves more like a dimension filter in that there's a searchable list of things to select from locally when filtering, in addition to allowing user inputted values. I don't think there's another metric that behaves like this.

I may be wrong, in that case, we should split the metrics into two categories, one for metrics for which we pre-populate a list of things the user can pick from, and ones for which we do not

packages/analytics/analytics-utilities/src/types/explore/requests.ts
alxgrant: Yea, we would like to show "kong" and "upstream". It's in the first sheet here https://docs.google.com/spreadsheets/d/1XQW4sXPlR7_7kUKxhKGTNueeQUZF3Pr4FJnbRz00axE/edit?gid=0#gid=0

PR #1656, opened 2024-09-25
fix(key): fix keySet query expected 404 [KM-545]

TT1228:
# Summary

KM-545

In a key config card, if this key is included in a deleted key set, querying it triggers global 404 in Konnect, we need to avoid this.


Leopoldthecoder (approved):
LGTM

PR #1657, opened 2024-09-25
fix(entities-plugins): set-typed fields

Leopoldthecoder:
# Summary

Treat set-typed fields as array fields. Changes include:
1. When building schema in the plugin form, treat set-typed fields as array fields. Removed additional logic for set-typed fields.
2. Removed frontend schema override for `confluent`, `jwt` and `kafka` plugins. The set-typed fields in these plugins can now be rendered correctly without frontend override.
3. Kept tags in credential forms as plain input boxes.

KM-193

#### Resources


Justineo (approved):
LGTM

PR #1660, opened 2024-09-25
fix(*): ktabledata adoption [KHCP-13280]

portikM:
# Summary

Jira: https://konghq.atlassian.net/browse/KHCP-13280

Changes:
* adopting KTableData in EntityBaseTable component (replacing KTable)
  * related type changes in EntityBaseTable types
* adding `InitialLoad` and `NoRecords` fetcher states in useFetcher composable
* utilizing new fetcher states for hiding toolbar in entities tables components in certain states

#### Resources


packages/entities/entities-gateway-services/src/components/GatewayServiceList.vue
```diff
@@ -390,6 +392,7 @@ const modalContent = computed(() => {
 const toggleEnableStatus = (row: EntityRow) => {
   enablementModalVisible.value = true
   switchEnablementTarget.value = row
+  delete switchEnablementTarget.value.selected
```
Justineo, line 395: I think we need to discuss the API design for this. There may be similar cases across the codebase that we've missed to cover. IMO ideally there should be a `selectedRowKeys: string[]` prop for `<KTableView>` so that we don't have to write into userland row data and then fix potential issues in an ad-hoc manner. And before that we should have a proper row key first as we have discussed.

/cc @adamdehaven @portikM

Leopoldthecoder: A little context: I added this line because there was an issue when toggling the `enabled` status of a gateway service:

Kapture 2024-09-29 at 16 21 29

`selected` is included in the request payload, resulting in a 400 error. This is caused by how the bulk actions functionality is implemented in `KTableView`: [it adds `selected` to the original data](https://github.com/Kong/kongponents/blob/6075bc9e00b7a07fdb0d495b11a9ac5fbf3c573b/src/components/KTableView/KTableView.vue#L275).

We feel instead of deleting the `selected` field every time we see an issue, it may be better to re-think the implementation of bulk actions as @Justineo suggested.

adamdehaven: I agree, this aligns with some of the concerns I brought up on the original PR on reserved keys

portikM: We can change the selected key to something more specific like `bulkActionsSelected` so it doesn't collide with plugins `enabled` status. Would that be sufficient?

adamdehaven: > We can change the selected key to something more specific like `bulkActionsSelected` so it doesn't collide with plugins `enabled` status. Would that be sufficient?

To be honest I think this is the issue with “reserved keys” - the component shouldn’t have a requirement limiting the keys utilized in the `data` object in the first place. Yes, something named this specifically will likely prevent a collision; however, I’m still not convinced this is the desired pattern.

portikM: I pushed a [fix](https://github.com/Kong/kongponents/pull/2405/commits/4a083b8ac23c072b78611f4d21c5577924a64e65) to KTableView, picked it up here and reversed @Leopoldthecoder's changes here. @Justineo @Leopoldthecoder could you confirm the fix solves the issue?

Justineo: I this case it isn't because `selected` collided our existing keys, the problem is it went into our request payload and the backend service doesn't accept unrecognized keys. We'll have to prune the data like the changed line above to make the request work. It's just ... unexpected.

portikM: Oh sorry, I understood it incorrectly. Let me update my fix.

Justineo: I think the proper use case of adding reserved keys is Table's slot props, we can provide things like `index`, `selected` and other internal states to all row-level slots (but with a prefix or naming convention though).

adamdehaven: I still don't understand why data/slot data internal to the table need to be injected/mutated the actual `data` from the host app.

e.g. if my data looks like this:

```ts
[
  { id: 1, title: 'Row 1' },
  { id: 2, title: 'Row 2' },
]
```

It's unexpected that the "bulk select" action would mutate this data: 

```ts
[
  { id: 1, title: 'Row 1' },
  { id: 2, title: 'Row 2', selected: true },
]
```

The selected state for bulk actions (along with the expanded rows, etc.) should be kept separate from the actual `data` object passed in from the host application and just linked to the row via a unique `key` (or something similar).

portikM: I didn't realize that the component returns the modified data object adding the `bulkActionsSelected` property through slot props and emits. I suggest this:
- I'll clean up all emits and slot props to make sure we're stripping out the internal value
- I'll make the key `bulkActionsSelected` even more specific adding `tableId` (random on each instance) in the key. In the end it will look something like this: `5549de9e-2246-48ec-95a1-13b57582d205-bulkActionsSelected: boolean`.

The reason why I don't want to go with the approach of splitting the bulk actions key into a separate object is because it will require a significant refactor and I am still convinced we can make the current implementation work (just need to properly handle some edge cases).

adamdehaven: I think we need to solve this and the key issues in separate PRs, isolated from the KTableData component

Justineo: > I still don't understand why data/slot data internal to the table need to be injected/mutated the actual `data` from the host app.

I wasn't suggesting to mutate data. I just was saying we can add such internal state to slot props but leave the original data unchanged. eg.

```vue
<slot name="foo" v-bind="rowData" :$selected="selected">
```

adamdehaven: > I wasn't suggesting to mutate data. 

Right, but that’s what the component actually does with the reserved keys

Leopoldthecoder: Putting aside the current implementation of `KTableView`, I think the ideal approach is to add a new `rowKey` prop, allowing users to specify which data field should be used as the row identifier. For instance, `rowKey` could be `'id'` for entity lists, since each Gateway entity has a unique ID. Within `KTableView`, instead of mutating the original data, we can use a ref (an array of IDs) to track which rows are expanded. Same can be applied for selected rows.

Additionally, we can introduce new props like `expandedRows` and `selectedRows` to give users control over which rows are expanded/selected externally. This would also cover scenarios where a row needs to be initially expanded/selected.

portikM: This sounds good to me 👆

portikM: Okay so I'll proceed with this steps:
- make sure internal property (`bulkActionsSelected`) is not exposed to the parent component
- make a ticket to modify bulk actions implementation based on @Leopoldthecoder's suggestion above

adamdehaven: Let's pause then on KTableData implementation and create tickets to refactor the underlying KTableView if we're in agreement?

If we can do a "mini component brief" on the changes, we can ensure we get buy-in from everyone in this thread

portikM: > Let's pause then on KTableData implementation

Why do we need to stop on KTableData implementation? After I will fix what I mentioned above KTableData implementation is not blocking us in any way

adamdehaven: 1. There's an existing issue in KTableView with using indexes as keys that presents an issue when a row is deleted
2. The bulk actions key and selected keys need to be refactored into separate props based on this PR discussion

There's no reason to release a new Kongponents component that utilizes KTableView under the hood with these substantial issues/changes in flight

portikM: **Update:** we decided to implement the refactor in KTableView and proceed with merging KTableData and deprecating KTable afterwards to avoid having to test it twice. I will add a sction in the original KTableView component brief discussion the need for new `rowKey` prop. After the refactor is implemented we shouldn't need the fix deleting an unexpected property from the row object.

portikM: The refactor has been implemented, this is ready a final review.

packages/entities/entities-certificates/src/components/CertificateList.vue
```diff
@@ -2,13 +2,13 @@
...
     <EntityBaseTable
       :cache-identifier="cacheIdentifier"
-      disable-pagination-page-jump
       :disable-sorting="disableSorting"
       :empty-state-options="emptyStateOptions"
       enable-entity-actions
       :error-message="errorMessage"
       :fetcher="fetcher"
       :fetcher-cache-key="fetcherCacheKey"
+      :hide-toolbar="hideTableToolbar"
```
Justineo, line 11: question: what criteria should be used to determine when the toolbar should be hidden?

portikM: The toolbar is hidden when fetcher returns no records and there is no search query or filters. Something like a "true" empty state. In that case non of the toolbar items (search, column visibility menu) are actionable.

adamdehaven (approved):

PR #1667, opened 2024-09-26
feat(forms): visibility toggle for password inputs [KM-502]

TT1228:
# Summary

KM-502

Some encrypted fields are required to be able to toggle visibility when edited in forms

### DIFF
--- a/packages/core/forms/src/components/fields/FieldInput.vue
+++ b/packages/core/forms/src/components/fields/FieldInput.vue
@@ -17,11 +17,27 @@
       :placeholder="schema.placeholder"
       :readonly="schema.readonly"
       :required="schema.required"
-      :type="inputType"
+      :type="displayInputType"
       :width="schema.width"
       @blur="onBlur"
       @update:model-value="onInput"
-    />
+    >
+      <template
+        v-if="schema.inputType === 'password'"
+        #after
+      >
+        <VisibilityOffIcon
+          v-if="eyeOpen"
+          role="button"
+          @click="() => { eyeOpen = !eyeOpen }"
+        />
+        <VisibilityIcon
+          v-else
+          role="button"
+          @click="() => { eyeOpen = !eyeOpen }"
+        />
+      </template>
+    </KInput>
 
     <!-- autofill -->
     <component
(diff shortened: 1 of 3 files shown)


packages/core/forms/src/components/fields/FieldInput.vue
```diff
@@ -17,11 +17,27 @@
...
+          role="button"
+          @click="() => { eyeOpen = !eyeOpen }"
+        />
+        <VisibilityIcon
+          v-else
+          role="button"
+          @click="() => { eyeOpen = !eyeOpen }"
+        />
+      </template>
+    </KInput>
```
Justineo, line 40: I think this feature should be built-in in Kongponents. Let's implement as-is and I'll propose it to the design system.

Justineo (approved):
LGTM

packages/core/forms/src/components/fields/FieldInput.vue
```diff
@@ -85,6 +102,14 @@ const emit = defineEmits<{
 }>()
 
 const propsRefs = toRefs(props)
+const eyeOpen = ref(false)
```
nekolab, line 105: nit: Would you mind changing `eyeOpen` to `maskPassword` for a clearer meaning?

packages/core/forms/src/components/fields/FieldInput.vue
```diff
@@ -85,6 +102,14 @@ const emit = defineEmits<{
...
+
+const displayInputType = computed(() => {
```
nekolab, line 107: Do we have a specific reason for using a separate `displayInputType` rather than directly modifying the `inputType`?

packages/core/forms/src/components/fields/FieldInput.vue
TT1228: Good point, merge it too soon. Since [Justineo](https://github.com/Justineo) will propose this as built-in Kongponents, i think we'll be refactoring this in near future.

packages/core/forms/src/components/fields/FieldInput.vue
Leopoldthecoder: I'd suggest we stick to `inputType` because it is the single source of truth in terms of determining the type of the input. IMO having another `displayInputType` introduces more complexity.

PR #1701, opened 2024-10-08
fix(keyform): unpass set,jwk and pem instead of passing null [KM-562]

TT1228:
# Summary

KM-562

When there is no value, instead of passing `null`, we don't pass `pem`, `jwk` and `set` in key form, cause koko now enforces null check for these parameters.


Leopoldthecoder:
LGTM, have you tested in KM?

TT1228:
Closing this since it was a bug of koko

PR #1702, opened 2024-10-08
fix: set custom max-widths for path names

andrewwylde:
# Summary

Context: Swagger UI Web Component uses shadow root, which makes overriding styles difficult from consuming applications.

Goal: Stop the operation path name from overflowing its container.

Solution: Add CSS Variables with sane defaults, and make changes to the renderer in order to prevent drastic overflow.

Before:
image

After:
image

NB: There is a [sister PR](https://github.com/Kong/swagger-ui-kong-theme/pull/183) to add the `title` property in order to get the full string to render on hover.


HenriPro (approved):
LGTM

PR #1705, opened 2024-10-09
fix: chart dataset sorting issue

filipgutica:
# Summary

Sorting logic was only looking at one segment within the bar rather than taking all segments into account.

https://konghq.atlassian.net/browse/MA-3157

## After
image

## Before
image

### DIFF
--- a/packages/analytics/analytics-chart/src/composables/useExploreResultToDatasets.spec.ts
+++ b/packages/analytics/analytics-chart/src/composables/useExploreResultToDatasets.spec.ts
@@ -178,214 +178,300 @@ describe('useVitalsExploreDatasets', () => {
     )
   })
 
-})
-
-it('handles no dimension', () => {
-  const exploreResult: ComputedRef<ExploreResultV4> = computed(() => ({
-    data: [
-      {
-        timestamp: '2022-01-01T01:01:02Z',
-        event: {
-          request_count: 1,
+  it('handles no dimension', () => {
+    const exploreResult: ComputedRef<ExploreResultV4> = computed(() => ({
+      data: [
+        {
+          timestamp: '2022-01-01T01:01:02Z',
+          event: {
+            request_count: 1,
+          },
         },
-      },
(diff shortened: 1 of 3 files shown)


packages/analytics/analytics-chart/src/composables/useExploreResultToDatasets.ts
```diff
@@ -99,7 +99,18 @@ export default function useExploreResultToDatasets(
             return [label, value]
           }))
 
-        const sortedDatasetIds = Object.entries(pivotRecords).sort(([, a], [, b]) => Number(b) - Number(a)).map(([key]) => (key.split(',')[0]))
+        const aggregatedPivotRecords = Object.keys(pivotRecords).reduce((acc, key) => {
```
adorack, line 102: Would it be possible to add a unit test to `useExploreResultToDatasets` that checks this case?  I see it impacted the component test, but I think it'd be helpful to have something more explicit.

adorack (changes requested):

packages/analytics/analytics-chart/src/composables/useExploreResultToDatasets.spec.ts
```diff
@@ -178,214 +178,300 @@ describe('useVitalsExploreDatasets', () => {
...
-          500: {
-            name: '500',
-            deleted: false,
+        } as GroupByResult,
+        {
+          timestamp: '2022-01-01T01:01:02Z',
+          event: {
+            route: 'route3',
+            status_code: 200,
+            metric1: 6,
```
adorack, line 440: I think that this doesn't actually test it, right?  In order to test it, we'd need 1 segment in route 1 or 2 to be larger than any 1 segment in route 3, but for the sum of segments in route 3 to be larger than the sum of segments in route 2.

Here, route 3 would be sorted first based on both the old criteria and the new criteria.

adorack (approved):

PR #1719, opened 2024-10-15
feat(portal): hide default tag when all tagged [TDX-4666]

amerc15:
# Summary

https://konghq.atlassian.net/browse/TDX-4666

When all endpoints had tags the default section would appear with nothing in it. So do not render the section when there are no untagged endpoints.


Mierenga (approved):
LGTM

andrewwylde (approved):
LGTM, thanks!

PR #1720, opened 2024-10-16
fix(entities-plugins): plugin list display incorrect scopes [KM-591]

2eha0:
# Summary
Compatible with different data structures in the List and Filter APIs
[KM-591](https://konghq.atlassian.net/browse/KM-591)

[KM-591]: https://konghq.atlassian.net/browse/KM-591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Leopoldthecoder:
> # Summary
> Compatible with different data structures in the List and Filter APIs [KM-591](https://konghq.atlassian.net/browse/KM-591)

I think it should be KM-590

TT1228 (commented):
I think we can apply same logic by taking `row.consumer_id` and `row.consumer_group_id` into account. I've observed same problem on `consumer` and `consumer group` scoped plugins

PR #1725, opened 2024-10-17
feat: add chart zoom custom plugin

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-3041
https://www.loom.com/share/74dc8af6086c434682fff43be22ce1a3

### DIFF
--- a/packages/analytics/analytics-chart/src/components/chart-plugins/VerticalLinePlugin.ts
+++ b/packages/analytics/analytics-chart/src/components/chart-plugins/VerticalLinePlugin.ts
@@ -2,6 +2,7 @@ import type { Chart, ChartType, Plugin, TooltipItem } from 'chart.js'
 
 interface VerticalLinePlugin extends Plugin {
   clickedSegment?: TooltipItem<ChartType>
+  pause?: boolean
 }
 
 const drawLine = (ctx: CanvasRenderingContext2D, x: number, top: number, bottom: number) => {
@@ -19,13 +20,13 @@ const drawLine = (ctx: CanvasRenderingContext2D, x: number, top: number, bottom:
 export const verticalLinePlugin: VerticalLinePlugin = {
   id: 'linePlugin',
   afterDatasetsDraw: function(chartInstance: Chart) {
-    if (chartInstance.tooltip && chartInstance.tooltip.getActiveElements() && chartInstance.tooltip.getActiveElements().length && !this.clickedSegment) {
+    if (!this.pause && chartInstance.tooltip && chartInstance.tooltip.getActiveElements() && chartInstance.tooltip.getActiveElements().length && !this.clickedSegment) {
       const { x } = chartInstance.tooltip.dataPoints[0].element
       const ctx = chartInstance.ctx
       drawLine(ctx, x, chartInstance.scales.y.top, chartInstance.scales.y.bottom)
     }
 
-    if (this.clickedSegment) {
+    if (!this.pause && this.clickedSegment) {
       const { x } = (this.clickedSegment as TooltipItem<ChartType>).element
       const ctx = chartInstance.ctx
       drawLine(ctx, x, chartInstance.scales.y.top, chartInstance.scales.y.bottom)
--- a/packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
+++ b/packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
@@ -197,8 +197,8 @@ const plugins = computed(() => [
   ...(props.type === 'timeseries_line' ? [verticalLinePlugin] : []),
 ])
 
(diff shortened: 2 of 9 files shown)


packages/analytics/analytics-chart/src/components/chart-plugins/VerticalLinePlugin.ts
```diff
@@ -2,6 +2,7 @@ import type { Chart, ChartType, Plugin, TooltipItem } from 'chart.js'
 
 interface VerticalLinePlugin extends Plugin {
   clickedSegment?: TooltipItem<ChartType>
+  pause?: boolean
```
filipgutica, line 5: Added a way for the vertical line plugin to be "paused" so we don't render it while were doing a zoom selection

packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
```diff
@@ -228,6 +246,26 @@ watch(() => props.type, () => {
   delete verticalLinePlugin.clickedSegment
 })
 
+
+nextTick(() => {
+  if (chartInstance.value?.chart) {
```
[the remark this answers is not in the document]
adorack: Isn't `nextTick` firing every time Vue applies updates?  That seems like it could potentially apply a ton of redundant event listeners, unless I'm missing something.

As long as it's only applying the listeners once, I don't think I have an opinion; we just want to make sure the `zoom` prop is reactive if at all possible.

packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
```diff
@@ -139,15 +141,26 @@ const props = defineProps({
...
+
 const { translateUnit } = composables.useTranslatedUnits()
 const chartInstance = ref<{ chart: Chart }>()
 const legendID = ref(uuidv4())
 const chartID = ref(uuidv4())
 const legendItems = ref<LegendItem[]>([])
 const tooltipElement = ref()
 const legendPosition = ref(inject('legendPosition', ChartLegendPosition.Right))
+const isDoingSelection = ref(false)
+const tz = (new Intl.DateTimeFormat()).resolvedOptions().timeZone
```
adorack, line 163: I _think_ it's better to not plug the timezone in at this point.  (It should be an optional member of `AbsoluteTimeRangeV4`.)

adorack (changes requested):

packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
```diff
@@ -176,7 +190,15 @@ const htmlLegendPlugin = {
...
-const plugins = computed(() => [htmlLegendPlugin, highlightPlugin, ...(props.type === 'timeseries_line' ? [verticalLinePlugin] : [])])
+const plugins = computed(() => [
+  htmlLegendPlugin,
+  highlightPlugin,
+  ...(props.zoom ? [dragSelectPlugin] : []),
+  ...(props.type === 'timeseries_line' ? [verticalLinePlugin] : []),
+])
+
+const remountLineKey = computed(() => plugins.value.map(p => p.id).join('-'))
+const remountBarKey = computed(() => plugins.value.map(p => p.id).join('-'))
```
filipgutica, line 201: Had to make two distinct keys even though they're the same, because of this: 
image

packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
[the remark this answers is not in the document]
filipgutica: because the reactivity of the zoom prop requires remounting of the underlying chart components in order to update plugins, event listeners need to be removed/re-added whenever the chart instance updates. Updated to use `watch` here.

packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
adorack: I think this is telling you that it's using the value of the key to manage the DOM: https://v3-migration.vuejs.org/breaking-changes/key-attribute

Maybe prefix the computed value with `'line'` or `'bar'`?

packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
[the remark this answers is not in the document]
adorack: That makes sense; I think it's fine.

adorack (changes requested):

packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
filipgutica: done 👍

adorack (approved):

PR #1730, opened 2024-10-18
fix(analytics): trend range text in multi-entity queries [MA-3250]

adorack:
In the case where we can't query a trend (either because the props said not to or because we're doing a multi-entity query), it doesn't make sense to render any trend text.

# Summary


mihai-peteu (approved):
Appreciate the Table View example in the sandbox :)  I tend to forget this is an actual use case in Gateway Mananger...

PR #1734, opened 2024-10-21
chore(deps): update to sass@1.8 and switch to @use

Justineo:
# Summary

Update to `sass@1.8`.

- Switch to `@use`

KM-620


packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -230,7 +230,8 @@ const getValue = (record: AnalyticsExploreRecord): string => {
 </script>
 
 <style lang="scss" scoped>
-@import '../styles/base';
+@use "../styles/base" as *;
```
adamdehaven, line 233: I don’t think you need to add the namespace if these aren’t variables (same comment for all instances)

packages/analytics/analytics-chart/src/styles/base.scss
```diff
@@ -1 +1,2 @@
-@import "./mixins";
+@forward "@kong/design-tokens/tokens/scss/variables";
```
adamdehaven, line 1: issue: these should be `@use` not `@forward`: https://github.com/sass/sass/issues/3782#issuecomment-2421142866

adamdehaven (changes requested):
A few issues

packages/analytics/analytics-chart/src/components/TopNTable.vue
adamdehaven: (I couldn’t pull up the file on mobile for some reason so maybe these aren’t vars)

packages/analytics/analytics-chart/src/styles/base.scss
Justineo: I saw that comment the other day and I wasn't quite following why we shouldn't use `@forward`. According to the [docs](https://sass-lang.com/documentation/at-rules/forward/), `@forward "foo"` should work similar to `export * from 'foo'` in JavaScript modules.

In this case changing the `@forward` here to `@use` does not work. `base.scss` acts like a barrel file in JS here.

packages/analytics/analytics-chart/src/components/TopNTable.vue
Justineo: Originally there's only `@import "./mixins.scss";` in it, I added variable inclusion into it as well. There's no CSS output in it.

packages/analytics/analytics-chart/src/styles/base.scss
adamdehaven: forward should be for mixins, functions, whereas use should be variables and "includes" as I understand it

adamdehaven: But you can't import variables into the global namespace with `@forward`? There is no namespace

Justineo:
```scss
// vars.scss
$heading-color: red;

// mixins.scss
@mixin heading {
  font-weight: 500;
  font-size: 20px;
}

// lib.scss
@forward "vars";
@forward "mixins";

// app.scss
@use "lib" as *;

h3 {
  @include heading;
  color: $heading-color;
}
```

Justineo: There is no more global namespace with `@use` so the best we can do is to inject variables to style file with Vite (those `@use`ed or `@forward`ed files won't be processed by Vite).

adamdehaven: > namespace

well there is, with `as *` 😬

Justineo: I mean there's no global namespace that can work across files. Anyway, I think it is correct to use `@forward` to compose a "library file" here.

And I don't quite like the file name either as it kind of implies that there is style output in it, which isn't the case. But as it named like this before, I'd say lets keep it as-is here.

adamdehaven: if `base` is just mixins/variables, feel free to rename them in this PR to clear things up as long as functionality remains the same

Justineo: Do we have a naming convention? Like "lib.scss"?

adamdehaven: No, we try to name more explicitly, e.g. `mixins` or `variables` etc.

In this case it could likely just be something like `globals` since that's what it is?

Justineo: 👍

Justineo: Renamed `base.scss` to `globals.scss`.

adamdehaven (approved):
Didn’t functionally test. Please verify in MFE repo

Justineo:
Manually tested at https://github.com/Kong/konnect-ui-apps/pull/4653.

Leopoldthecoder (approved):
Tests in `konnect-ui-apps` are passing. LGTM.

adamdehaven (approved):

PR #1738, opened 2024-10-23
fix(analytics): trigger chart and dashboard patch version builds

mihai-peteu:
# Summary

Attempting to unblock https://github.com/Kong/konnect-ui-apps/pull/4642

Slack thread:
https://kongstrong.slack.com/archives/C05UE909XU5/p1729651157682919

### DIFF
--- a/packages/analytics/analytics-chart/src/types/chart-types.ts
+++ b/packages/analytics/analytics-chart/src/types/chart-types.ts
@@ -4,6 +4,7 @@ export const chartTypes = [
   'vertical_bar',
   'timeseries_line',
   'timeseries_bar',
+  'choropleth_map',
 ] as const
 
 export type ChartType = typeof chartTypes[number]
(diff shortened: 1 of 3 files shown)


packages/analytics/analytics-chart/src/types/chart-types.ts
```diff
@@ -4,6 +4,7 @@ export const chartTypes = [
   'vertical_bar',
   'timeseries_line',
   'timeseries_bar',
+  'choropleth_map',
```
mihai-peteu, line 7: This should be the larger Venn diagram that includes all the possible `reportChartTypes` , therefore should include chloropleth as well

filipgutica (changes requested):
Map chart type should not be going  into analytics-chart

packages/analytics/analytics-chart/src/types/chart-types.ts
filipgutica: We need to look into it more tomorrow. Map chart type doesn't result in anything this package is capable of so it shouldn't be here 🤔

PR #1757, opened 2024-10-31
fix(entities-plugins): custom plugins disappeared after page reload [KM-658]

2eha0:
# Summary

For the reproduction method, please refer to: [KM-658](https://konghq.atlassian.net/browse/KM-658)

in fact, this problem is caused by the konnect, which hides the app container during initialization ([see here](https://github.com/Kong/shared-ui-components/blob/defb8926e9a7ae6b2c29239fdd895d4d813dfe8d/packages/core/konnect-app-shell/src/components/KonnectAppShell.vue#L305)), during which the card height is always calculated to be 0px.

This is a quick and easy fix that provide a minimum height.

[preview link](https://pr-4755--gateway-manager.cloud-preview.konghq.tech/gateway-manager)

[KM-658]: https://konghq.atlassian.net/browse/KM-658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-plugins/src/components/PluginIcon.vue
+++ b/packages/entities/entities-plugins/src/components/PluginIcon.vue
@@ -2,6 +2,7 @@
   <img
     ref="img"
     :alt="alt"
+    :height="size"
     :src="iconSrc"
     :width="size"
     @error="onError"
--- a/packages/entities/entities-plugins/src/components/custom-plugins/PluginCustomGrid.vue
+++ b/packages/entities/entities-plugins/src/components/custom-plugins/PluginCustomGrid.vue
@@ -185,7 +185,8 @@ const handleClose = (revalidate?: boolean): void => {
   selectedPlugin.value = null
 }
 
-const tallestPluginCardHeight = ref<number>(310) // begin with default height
+const minHeight = 310
+const tallestPluginCardHeight = ref<number>(minHeight) // begin with default height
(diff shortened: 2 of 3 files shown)


packages/entities/entities-plugins/src/composables/usePluginHelpers.ts
```diff
@@ -154,8 +154,8 @@ export default function useHelpers() {
   /**
    * Get the height of the tallest card
    */
-  const getTallestPluginCardHeight = (elements: Array<InstanceType<typeof PluginSelectCard>>): number => {
-    let tallestCardHeight = 0
+  const getTallestPluginCardHeight = (elements: Array<InstanceType<typeof PluginSelectCard>>, minHeight: number = 0): number => {
```
TT1228, line 157: Since we're passing `minHeight = 310` for `getTallestPluginCardHeight` in every reference, could we make it default in this function as `minHeight: number = 310`

2eha0: @TT1228 I modified the implementation a bit, I think we shouldn't modify the function `getTallestPluginCardHeight`, I use `Math.max(xxx, minHeight)` inside in component to ensure the minimum height

TT1228 (approved):
Manually tested and LGTM

PR #1758, opened 2024-11-01
feat(dashboards and charts): editable tiles | slottable menu items

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-3367

- add new edit-tile event to dashboard renderer
- forward `go-to -explore` prop to analytics chart through tile definition
- add new editable property to tile definition that will add an "edit" item to the analytics chart kebab menu. Clicking on edit, will emit an @edit-tile event with the entire GridTile<T> object containing the tile layout, definition and index

## `@edit-tile` event
```
{
    "layout": {
        "position": {
            "col": 0,
            "row": 2
        },
        "size": {
            "cols": 3,
            "rows": 2
        }
    },
    "meta": {
        "chart": {
            "type": "horizontal_bar",
            "chartTitle": "Horizontal bar chart of mock data",
            "allowCsvExport": true
        },
        "query": {
            "datasource": "basic",
            "dimensions": [
                "route"
            ]
        },
        "editable": true
    },
    "id": 3
}
```

image

### DIFF
--- a/packages/analytics/dashboard-renderer/src/components/BarChartRenderer.vue
+++ b/packages/analytics/dashboard-renderer/src/components/BarChartRenderer.vue
@@ -1,42 +1,19 @@
+<!-- BarChartRenderer.vue -->
 <template>
-  <QueryDataProvider
-    v-slot="{ data }"
+  <BaseAnalyticsChartRenderer
+    :chart-options="chartOptions"
     :context="context"
+    :editable="editable"
+    :extra-props="{ showAnnotations: false }"
+    :height="height"
     :query="query"
     :query-ready="queryReady"
-  >
-    <div class="analytics-chart">
-      <AnalyticsChart
-        :allow-csv-export="chartOptions.allowCsvExport"
-        :chart-data="data"
-        :chart-options="options"
-        :chart-title="chartOptions.chartTitle"
-        legend-position="bottom"
-        :show-annotations="false"
-        :synthetics-data-key="chartOptions.syntheticsDataKey"
-        tooltip-title=""
-      />
-    </div>
-  </QueryDataProvider>
+  />
 </template>
+
 <script setup lang="ts">
+import BaseAnalyticsChartRenderer from './BaseAnalyticsChartRenderer.vue'
 import type { BarChartOptions, RendererProps } from '../types'
-import QueryDataProvider from './QueryDataProvider.vue'
-import { computed } from 'vue'
-import type { AnalyticsChartOptions } from '@kong-ui-public/analytics-chart'
-import { AnalyticsChart } from '@kong-ui-public/analytics-chart'
 
-const props = defineProps<RendererProps<BarChartOptions>>()
-
-const options = computed((): AnalyticsChartOptions => ({
-  type: props.chartOptions.type,
-  stacked: props.chartOptions.stacked,
-  chartDatasetColors: props.chartOptions.chartDatasetColors,
-}))
+defineProps<RendererProps<BarChartOptions>>()
 </script>
-
-<style scoped lang="scss">
-.analytics-chart {
-  height: v-bind('`${height}px`');
-}
-</style>
--- a/packages/analytics/dashboard-renderer/src/components/BarChartRenderer.vue
+++ b/packages/analytics/dashboard-renderer/src/components/BarChartRenderer.vue
@@ -3,7 +3,6 @@
   <BaseAnalyticsChartRenderer
     :chart-options="chartOptions"
     :context="context"
-    :editable="context.editable"
     :extra-props="{ showAnnotations: false }"
     :height="height"
     :query="query"
--- a/packages/analytics/dashboard-renderer/src/components/BarChartRenderer.vue
+++ b/packages/analytics/dashboard-renderer/src/components/BarChartRenderer.vue
@@ -3,7 +3,7 @@
   <BaseAnalyticsChartRenderer
     :chart-options="chartOptions"
     :context="context"
-    :editable="editable"
+    :editable="context.editable"
     :extra-props="{ showAnnotations: false }"
     :height="height"
     :query="query"
--- a/packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
+++ b/packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
@@ -4,6 +4,7 @@
       :is="componentData.component"
       v-if="componentData"
       v-bind="componentData.rendererProps"
+      @edit-tile="onEditTile"
     />
   </div>
 </template>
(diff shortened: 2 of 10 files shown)


adorack (changes requested):
I'd like to review.

packages/analytics/dashboard-renderer/src/components/GoldenSignalsRenderer.vue
```diff
@@ -46,18 +46,24 @@ const overrideTimeframe: Ref<Timeframe> = computed(() => {
   return relativePeriod
 })
 
-const options = computed<ProviderProps>(() => ({
```
filipgutica, line 49: There was a type error here that's somehow getting past CI... 🤔  

```
No overload matches this call.
  Overload 1 of 2, '(getter: ComputedGetter<Partial<{ description: string; containerTitle: string; refreshInterval: number; datasource: "advanced" | "basic"; dimension: "api_product" | "api_product_version" | "control_plane" | ... 9 more ... | "iso_code"; ... 8 more ...; percentileLatency: boolean; }> & Omit<...>>, debugOptions?: DebuggerOptions | undefined): ComputedRef<...>', gave the following error.
    Type '"advanced" | "basic" | "ai"' is not assignable to type '"advanced" | "basic" | undefined'.
      Type '"ai"' is not assignable to type '"advanced" | "basic" | undefined'.
  Overload 2 of 2, '(options: WritableComputedOptions<Partial<{ description: string; containerTitle: string; refreshInterval: number; datasource: "advanced" | "basic"; dimension: "api_product" | "api_product_version" | "control_plane" | ... 9 more ... | "iso_code"; ... 8 more ...; percentileLatency: boolean; }> & Omit<...>, Partial<...> & Omit<...>>, debugOptions?: DebuggerOptions | undefined): WritableComputedRef<...>', gave the following error.
    Argument of type '() => { datasource: "advanced" | "basic" | "ai"; overrideTimeframe: Timeframe; tz: string; additionalFilter: ExploreFilter[]; longCardTitles: boolean | undefined; containerTitle: string | undefined; description: string | undefined; percentileLatency: boolean | undefined; refreshInterval: number; queryReady: boolean; }' is not assignable to parameter of type 'WritableComputedOptions<Partial<{ description: string; containerTitle: string; refreshInterval: number; datasource: "advanced" | "basic"; dimension: "api_product" | "api_product_version" | "control_plane" | ... 9 more ... | "iso_code"; ... 8 more ...; percentileLatency: boolean; }> & Omit<...>, Partial<...> & Omit<....'.ts-plugin(2769)
MetricsProvider.vue.d.ts(67, 5): The expected type comes from property 'datasource' which is declared here on type 'Partial<{ description: string; containerTitle: string; refreshInterval: number; datasource: "advanced" | "basic"; dimension: "api_product" | "api_product_version" | "control_plane" | ... 9 more ... | "iso_code"; ... 8 more ...; percentileLatency: boolean; }> & Omit<...>'
```

filipgutica: I think it's from updates to analytics-utilities which introduced the `ai` datasource, while the provider props in metrics provider defines datasource property like so: `datasource?: "basic" | "advanced" | undefined;` 

Not sure if my change is the right one, or we should add `ai` to the acceptable datasources in MetricsProvider

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
```diff
@@ -55,6 +55,14 @@ const allowCsvExport = {
   type: 'boolean',
 } as const
 
+const goToExplore = {
```
adorack, line 58: I don't think either of these properties should be exposed in the schema.  The things in the schema are eventually going to be part of the public API for dashboards; there's no case where an end user should be able to directly configure these props.

I'd recommend:

- `editable` is a new property exposed via `context`.  I think it's OK (especially for now) to say that either all tiles are editable or none are.
- `goToExplore` is probably a base URL that's configured via the analytics bridge, and the query data provider calculates the final URL to hand to `AnalyticsChart`.

packages/analytics/dashboard-renderer/src/components/GoldenSignalsRenderer.vue
adorack: Hmmm.  I'll make a note to look at this later.  I think this is the fault of the schema; it shouldn't allow setting `ai` as a datasource for metric cards.  I think what you have for now is fine.

adorack (changes requested):

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
filipgutica: How does the dashboard know if the tiles should have go to explore or not? 🤔

packages/analytics/dashboard-renderer/src/components/BarChartRenderer.vue
```diff
@@ -3,7 +3,7 @@
   <BaseAnalyticsChartRenderer
     :chart-options="chartOptions"
     :context="context"
-    :editable="editable"
+    :editable="context.editable"
```
adorack, line 6: You're already passing in `context` -- no need for the extra prop.  :)

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -61,7 +61,7 @@ const componentData = computed(() => {
       queryReady: props.queryReady,
       chartOptions: props.definition.chart,
       height: props.height - PADDING_SIZE * 2,
-      editable: props.definition.editable,
+      editable: props.context.editable,
```
adorack, line 64: I don't think you need to pass in this prop.

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
```diff
@@ -59,10 +60,6 @@ const goToExplore = {
   type: 'string',
 } as const
```
adorack, line 61: Please remove `goToExplore` as well.

adorack (changes requested):

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
filipgutica: done

packages/analytics/dashboard-renderer/src/components/BarChartRenderer.vue
filipgutica: done

packages/analytics/dashboard-renderer/src/types/dashboard-renderer-types.ts
filipgutica: done

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
adorack: It looks like it's still there...

adorack (changes requested):

adorack (approved):

PR #1759, opened 2024-11-04
fix(pluginlist): adapt empty ordering object for filtered plugin

TT1228:
# Summary

KM-662

Filtered plugin item in Konnect contains `ordering:{}` in response, this results in current UI showing dynamic badge for static ordering. This pr allows passing empty ordering object too.


Leopoldthecoder (approved):
LGTM

PR #1765, opened 2024-11-05
feat: add explore base url to query bridge for analytics jump to explore

filipgutica:
# Summary

- Add optional `exploreBaseUrl` property to AnalyticsQueryBridge interface
- If Query Bridge is provided with an `exploreBaseUrl` this will be forwarded to `analytics-chart` along with the explore query, datasource and chart type as the `go-to-explore` link, which will show the go to explore option in the analytics chart kebab menu
- Updated "go to explore" text to match figma

https://konghq.atlassian.net/browse/MA-3380

image

### DIFF
--- a/packages/analytics/analytics-utilities/src/types/query-bridge.ts
+++ b/packages/analytics/analytics-utilities/src/types/query-bridge.ts
@@ -30,4 +30,5 @@ export interface AnalyticsBridge {
   //  https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html#the-noinfer-utility-type
   //  See note in DashboardRenderer tests.
   evaluateFeatureFlagFn: <T = boolean>(key: string, defaultValue: T) => T,
+  exploreBaseUrl?: string,
 }
--- a/packages/analytics/analytics-utilities/src/types/query-bridge.ts
+++ b/packages/analytics/analytics-utilities/src/types/query-bridge.ts
@@ -30,5 +30,5 @@ export interface AnalyticsBridge {
   //  https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html#the-noinfer-utility-type
   //  See note in DashboardRenderer tests.
   evaluateFeatureFlagFn: <T = boolean>(key: string, defaultValue: T) => T,
-  exploreBaseUrl?: string,
+  exploreBaseUrl?: () => string,
 }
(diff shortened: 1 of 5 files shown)


adorack (changes requested):
I'd like to review.

packages/analytics/analytics-utilities/src/types/query-bridge.ts
```diff
@@ -30,4 +30,5 @@ export interface AnalyticsBridge {
   //  https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html#the-noinfer-utility-type
   //  See note in DashboardRenderer tests.
   evaluateFeatureFlagFn: <T = boolean>(key: string, defaultValue: T) => T,
+  exploreBaseUrl?: string,
```
adorack, line 33: Just to be safe, let's make this a function: `exploreBaseUrl?: () => string`.  I can't think right now of a case it'd have trouble with, but sometimes the query bridge doesn't have full information at the time it's constructed.

adorack (changes requested):

packages/analytics/dashboard-renderer/src/components/BaseAnalyticsChartRenderer.vue
```diff
@@ -50,6 +54,26 @@ const options = computed((): AnalyticsChartOptions => ({
...
+          ...filter,
+          type: filter.type,
+        } as ExploreFilter
+      }) ?? [],
+      metrics: props.query.metrics as ExploreAggregations[] | AiExploreAggregations[] ?? [],
+      dimensions: props.query.dimensions as QueryableExploreDimensions[] | QueryableAiExploreDimensions[] ?? [],
+      time_range: props.query.time_range as TimeRangeV4 || props.context.timeSpec,
+
+    } as ExploreQuery | AiExploreQuery
+    return `${queryBridge.exploreBaseUrl()}?q=${JSON.stringify(exploreQuery)}&d=${props.query.datasource}&c=${props.chartOptions.type}`
```
adorack, line 66: Sorry I missed this earlier: `props.query.datasource` might not be populated at this stage; it's not required in the schema yet and the fallback is in `QueryDataProvider`.  You can probably do `props.query.datasource || 'advanced'` in this case.

adorack (changes requested):

packages/analytics/dashboard-renderer/src/components/BaseAnalyticsChartRenderer.vue
filipgutica: done

adorack (approved):

adorack (approved):

PR #1769, opened 2024-11-06
fix(entities-plugins): ai-proxy-advanced form

Leopoldthecoder:
# Summary

This is a temporary workaround for the issue that the `ai-proxy-advanced` plugin cannot be created if users don't provide `config.embeddings` or `config.vectordb`.

KM-675

### DIFF
--- a/packages/entities/entities-plugins/src/components/PluginForm.vue
+++ b/packages/entities/entities-plugins/src/components/PluginForm.vue
@@ -166,6 +166,7 @@ import {
   type PluginFormFields,
   type PluginFormState,
   type PluginOrdering,
+  type CustomSchemas,
 } from '../types'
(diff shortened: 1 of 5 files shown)


sumimakito (approved):
Manually tested and the submission via the button worked as expected.

One more thing: Would we want to keep the model shown in View Configuration consistent with the final payload that will be submitted by us?

|Configuration|Request payload|
|:-|:-|
|Screenshot 2024-11-08 at 14 46 53|Screenshot 2024-11-08 at 14 50 22|

Leopoldthecoder:
Ideally yes, we should keep them consistent. But it is not easy to do given the current implementation, as we are copying the payload so that it can be transformed. Will revisit if necessary.

PR #1772, opened 2024-11-07
fix: chart title should not wrap

filipgutica:
# Summary

## Before
image

## After
image

### DIFF
--- a/packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
+++ b/packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
@@ -7,6 +7,7 @@
       <div
         v-if="chartTitle"
         class="chart-title"
+        :title="chartTitle"
       >
         {{ chartTitle }}
       </div>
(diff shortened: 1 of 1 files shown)


packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
```diff
@@ -7,6 +7,7 @@
       <div
         v-if="chartTitle"
         class="chart-title"
+        :title="chartTitle"
```
mihai-peteu, line 10: this is for accessibility reasons I assume 

@adamdehaven | @portikM - preference on `title` vs `aria-label` here?

adamdehaven: If you need to hover and see the text, `title`

Otherwise, `aria-label`

adamdehaven: You may also want a `role="figure` https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/figure_role

mihai-peteu (approved):

PR #1775, opened 2024-11-11
fix(i18n): compatible with timestamp in milliseconds or null values

Justineo:
# Summary

The filter API is returning timestamps in milliseconds instead of seconds and `null` as `updated_at` values for those weren't modified since created. The issue affects both filtered services and routes.

Before:

image

After:

image

### DIFF
--- a/packages/core/i18n/src/i18n.ts
+++ b/packages/core/i18n/src/i18n.ts
@@ -70,10 +70,10 @@ export const createI18n = <MessageSource extends Record<string, any>>
    * @param {Number} timestamp a unix timestamp in seconds
    * @returns a date string formatted like 'Apr 6, 2022 10:50'
    */
-  const formatUnixTimeStamp = (timestamp?: number | null): string => {
+  const formatUnixTimeStamp = (timestamp: number): string => {
     const invalidDate = 'Invalid Date'
     if (!timestamp) {
-      return '-'
+      return invalidDate
     }
 
     try {
--- a/packages/core/i18n/src/i18n.ts
+++ b/packages/core/i18n/src/i18n.ts
@@ -53,18 +53,31 @@ export const createI18n = <MessageSource extends Record<string, any>>
   const intl = otherProps
 
   /**
+   * Shamefully normalize a timestamp to be in seconds as some APIs are returning timestamps in milliseconds
+   * TODO: Remove this function once all timestamps are normalized from the backend
+   * @param {number} timestamp a unix timestamp in seconds *or milliseconds*
+   * @returns {number} a unix timestamp in seconds
+   */
+  const shamefullyNormalizeTimeStamp = (timestamp: number): number => {
+    if (timestamp.toString().length === 13) {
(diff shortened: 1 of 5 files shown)


packages/core/i18n/src/i18n.ts
```diff
@@ -52,19 +52,32 @@ export const createI18n = <MessageSource extends Record<string, any>>
...
    * Formats a unix timestamp into a formatted date string
    * @param {Number} timestamp a unix timestamp in seconds
    * @returns a date string formatted like 'Apr 6, 2022 10:50'
    */
-  const formatUnixTimeStamp = (timestamp: number): string => {
+  const formatUnixTimeStamp = (timestamp?: number | null): string => {
     const invalidDate = 'Invalid Date'
     if (!timestamp) {
-      return invalidDate
+      return '-'
```
ValeryG, line 76: not saying this is wrong change, but would some places where user expects to see `Invalid Date` and now sees `-` be "broken" or "changed without the user's content" ? Also are we sure there are no tests that expects `Invalid Date` ? 
Seems that the change is not related to shameful normalize and can be excluded from this PR, unless it's absolutely needed?

Justineo: I'm not sure if there are valid use cases where users should see `Invalid Date`. I can revert this change, as it was intended to handle potential `null` values returned by backend APIs, and we are already providing the correct fallback value (use `created_at` if `updated_at` is `null`). I’m just curious to understand if there are any scenarios in our current design where we might expect to see `Invalid Date`.

ValeryG: I don't see this string anywhere in the tests. So looks like safe change. If someone complains we just file JIRA and fix :)

Justineo: Sounds good. Can you stamp on this one?

kaiarrowood, line 76:
```suggestion
      return '—'
```

Can we use emdash instead please?

Justineo: I think I should just remove this. We are already handling `null` before this.

packages/core/i18n/src/i18n.ts
```diff
@@ -52,6 +52,19 @@ export const createI18n = <MessageSource extends Record<string, any>>
   const { $t, ...otherProps } = intlOriginal
   const intl = otherProps
 
+  /**
+   * Shamefully normalize a timestamp to be in seconds as some APIs are returning timestamps in milliseconds
+   * TODO: Remove this function once all timestamps are normalized from the backend
```
Justineo, line 57: Ticket for the backend: [MA-3412](https://konghq.atlassian.net/browse/MA-3412)

Justineo:
I've updated the code and no longer handles `null` value when formatting the timestamp. PTAL @ValeryG @kaiarrowood

kaiarrowood (approved):
LGTM! 🚀

PR #1776, opened 2024-11-11
chore(entity-base-table): fix missing prop warning

TT1228:
# Summary

Defining `preferencesStorageKey` cause missing prop warning in console. Since `preferencesStorageKey` has a default value, it can be removed.


Leopoldthecoder (approved):
LGTM

PR #1784, opened 2024-11-12
feat: make dashboards "refreshable"

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-3089

- Expose a "refresh" function from dashboard-renderer so host app can initiate "refresh"
- Use state validating and validating_has_data to determine loading state appropriately 

https://www.loom.com/share/afcd026de712408da886e722cb044395


adorack (changes requested):
I'd like to review.

packages/analytics/analytics-metric-provider/src/components/metricsProviderUtil.ts
```diff
@@ -39,6 +39,7 @@ interface FetcherOptions {
   queryFn: AnalyticsBridge['queryFn']
   averageLatencies: Ref<boolean>,
   abortController?: AbortController
+  refreshCounter?: Ref<number>
```
adorack, line 42: Since the prop has a default value, this option will always be available.
```suggestion
  refreshCounter: Ref<number>
```

packages/analytics/analytics-metric-provider/src/composables/useMetricFetcher.ts
```diff
@@ -173,7 +173,7 @@ export default function useMetricFetcher(opts: MetricFetcherOptions): FetcherRes
   })
 
   return {
-    isLoading: computed(() => STATE.PENDING === metricRequestState.value),
+    isLoading: computed(() => [STATE.PENDING, STATE.VALIDATING, STATE.VALIDATING_HAS_DATA].includes(metricRequestState.value)),
```
adorack, line 176: Hmmm.  This makes the auto-refresh UX very noisy -- the page drops to loading states every time the charts update.  I'm not aware of many uses of auto-refresh right now, so this is probably an acceptable trade-off right now, but I'll poke Christian and Helen.

packages/analytics/analytics-metric-provider/src/types/fetcher-types.ts
```diff
@@ -16,4 +16,5 @@ export interface MetricFetcherOptions {
   queryReady: Ref<boolean>
   queryFn: AnalyticsBridge['queryFn']
   abortController?: AbortController
+  refreshCounter?: Ref<number>
```
adorack, line 19:
```suggestion
  refreshCounter: Ref<number>
```

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -31,8 +31,10 @@ const props = withDefaults(defineProps<{
   context: DashboardRendererContextInternal,
   height?: number,
   queryReady: boolean,
+  refreshCounter?: number,
```
adorack, line 34: Does this need to be optional?  The renderer is the only thing that renders tiles, and it always provides the counter...

adorack (changes requested):
Please fix the PR title before merging.

packages/analytics/analytics-metric-provider/src/composables/useMetricFetcher.ts
filipgutica: You're right. I removed this change, I don't think refresh should trigger the loading state unless Christian/Helen really want it, so we can add that back in if they want that.

packages/analytics/analytics-metric-provider/src/composables/useMetricFetcher.ts
```diff
@@ -173,7 +173,7 @@ export default function useMetricFetcher(opts: MetricFetcherOptions): FetcherRes
   })
 
   return {
-    isLoading: computed(() => STATE.PENDING === metricRequestState.value),
+    isLoading: computed(() => metricRequestState.value === STATE.PENDING || [STATE.PENDING, STATE.VALIDATING, STATE.VALIDATING_HAS_DATA].includes(metricRequestState.value)),
```
adorack, line 176:
```suggestion
    isLoading: computed(() => [STATE.PENDING, STATE.VALIDATING, STATE.VALIDATING_HAS_DATA].includes(metricRequestState.value)),
```

packages/analytics/dashboard-renderer/src/components/QueryDataProvider.vue
```diff
@@ -133,6 +134,6 @@ const { state, swrvState: STATE } = useSwrvState(v4Data, error, isValidating)
 
 const errorMessage = ref<string | null>(null)
 const hasError = computed(() => state.value === STATE.ERROR || !!errorMessage.value)
-const isLoading = computed(() => !props.queryReady || state.value === STATE.PENDING)
+const isLoading = computed(() => !props.queryReady || state.value === STATE.PENDING || state.value === STATE.VALIDATING_HAS_DATA || state.value === STATE.VALIDATING)
```
adorack, line 137: Minor, but could you make this use the same logic as the one above?  Easier to change that way if we decide want to.

adorack (changes requested):

packages/analytics/analytics-metric-provider/src/composables/useMetricFetcher.ts
filipgutica: done

adorack (approved):

PR #1785, opened 2024-11-13
fix: analytics chart kebab menu positioning

filipgutica:
# Summary

https://www.loom.com/share/5fa43f74fcbe4764888c49e0d1dc38ea

- show kebab menu on hover only
- if there's no title overlap canvas and show on hover


packages/analytics/analytics-chart/src/components/AnalyticsChart.cy.ts
```diff
@@ -302,7 +302,7 @@ describe('<AnalyticsChart />', () => {
     })
   })
 
-  it('Renders an "Export" button, and tabulated data in the modal preview', () => {
+  it.only('Renders an "Export" button, and tabulated data in the modal preview', () => {
```
adorack, line 305: Please remove the `.only`.

packages/analytics/dashboard-renderer/sandbox/pages/RendererDemo.vue
```diff
@@ -153,7 +153,6 @@ const dashboardConfig: DashboardConfig = {
     {
       definition: {
         chart: {
-          chartTitle: 'Timeseries line chart of mock data',
```
adorack, line 156: Minor, but please remove this change; we might still trigger a new renderer version because it depends on `analytics-chart`, but it's nice to avoid code changes when possible.

adorack (changes requested):
(Still reviewing; a couple of minor things so far.)

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
```diff
@@ -429,6 +438,22 @@ const chartTooltipSortFn = computed(() => {
   }
 })
 
+const chartHeaderPosition = computed(() => {
+  return props.chartTitle || !hasKebabMenuAccess || (resultSetTruncated.value && maxEntitiesShown.value) ? 'relative' : 'absolute'
```
adorack, line 442: `resultSetTruncated && maxEntitiesShown` -- I know this is the old logic, but shouldn't this be `||` rather than `&&`?  We need to show a header if either of these warnings is visible?  (Honest question, I may not be up to date on the designs.)

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
```diff
@@ -364,10 +377,6 @@ const hasValidChartData = computed(() => {
   return props.chartData && props.chartData.meta && props.chartData.data.length
 })
 
-const showChartHeader = computed(() => {
```
adorack, line 367: If a chart doesn't have a title and doesn't have valid data, does this behave the same way as before?  (It seems like we're showing a header in that case where before we didn't.)

filipgutica: Yes, because even in those cases, we may still want to show the kebab menu that may still contain valid actions a user can take, despite the lack of valid chart data. 

For example, a chart in a dashboard may not have data for whatever reason. A user may still want to "go to explore" or "edit" the tile in order to investigate why the chart has no data, and make changes. So I think the old way has to be changed.

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
filipgutica: I don't remember how this part of explore works, but may be this was based off an old faulty assumption. 😅 

Those computed props are based off the `limit` and `truncated` in the explore response. Is the presence of both required to determine that the results are truncated? If not, then may be we can just use `trancated` ?

packages/analytics/analytics-chart/src/components/AnalyticsChart.cy.ts
filipgutica: Wooops 🙈

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
```diff
@@ -43,10 +47,17 @@
...
-        <MoreIcon
-          :color="KUI_COLOR_TEXT_NEUTRAL"
-          :size="KUI_ICON_SIZE_40"
-        />
+        <button
+          appearance="none"
+          :aria-label="i18n.t('more_actions')"
+          class="kebab-action-menu"
+          data-testid="kebab-action-menu"
+        >
```
mihai-peteu, line 55: good call here, without the button element, the nested SVG can sometimes "steal" the click action

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
adorack: I think that makes sense; my main concern is that this is changing outside of the feature flag.  I took a look at the sandbox, and I think the change is basically invisible, so I won't worry about it.

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
adorack: Aaagh, I'm sorry.  I dug a bit deeper and these two things are complementary: I lost the thread and thought they were two different things.  Technically, you could just use `truncated`, but the current logic is fine.  No need for changes.

adorack (approved):

PR #1790, opened 2024-11-18
fix(entities-*): revamp hide toolbar logic

Justineo:
# Summary

[KM-709]

Preview PR for `konnect-ui-apps`: https://github.com/Kong/konnect-ui-apps/pull/4896

> [!NOTE]
> I've updated this PR to a more comprehensive fix for several current issues.

Currently, logic related to table state and toolbar visibility is scattered across multiple places, leading to redundancy, inaccuracies, and visual inconsistencies. Here's a breakdown of the existing problems and their respective issues:

1. Each entity list observes `fetcherState` to update a `hasData` flag, which determines the visibility of the create button:

    https://github.com/Kong/public-ui-components/blob/1ea901eefe483102499f4aa3ee0d309ecf2df787/packages/entities/entities-gateway-services/src/components/GatewayServiceList.vue#L562-L565

    > Problem: It does not account for cache states in `KTableData`, causing layout shifts during revalidation.

2. Each entity list observes `fetcherState` to toggle the `hideTableToolbar` flag and set it to `true` when the state is `FetcherStatus.NoRecords`:

    https://github.com/Kong/public-ui-components/blob/1ea901eefe483102499f4aa3ee0d309ecf2df787/packages/entities/entities-gateway-services/src/components/GatewayServiceList.vue#L567-L571

    > Problem: Similar cache-awareness issues result in layout shifts when navigating away and returning to the list.

3. `EntityBaseTable` also controls the visibility of toolbar *content* using the `KTableData`'s state event:

    https://github.com/Kong/public-ui-components/blob/1ea901eefe483102499f4aa3ee0d309ecf2df787/packages/entities/entities-shared/src/components/entity-base-table/EntityBaseTable.vue#L264-L266

    > Problem: Toolbar visibility and its content can desynchronize (e.g., filter inputs hidden but the toolbar remains visible).

## Refactored Logic

To address these issues, all related logic has been centralized within the `EntityBaseTable`, ensuring consistent behavior across all use cases. The refactored workflow includes:

1. We no longer show/hide create button or toolbar content separate to the toolbar itself
2. The toolbar is hidden if nothing has been added to the table (initial state)
3. `KTableData`'s `hideToolbar` is now calculated by table states and query
4. If a filter query exists, the toolbar remains visible regardless of the table state (e.g., no records or no filter results)
5. If the table has previously contained data (even if none is visible now due to clearing a query), the toolbar will remain visible as we are confident that we'll see unfiltered records
6. Toolbar visibility remains configurable through the current `hideToolbar` prop, which now defaults to `undefined` so that we can offer the default behavior if it's not specified

[KM-709]: https://konghq.atlassian.net/browse/KM-709?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


2eha0 (dismissed):
LGTM

TT1228 (dismissed):

nekolab (approved):
LGTM

kaiarrowood (approved):
This looks good to me, awesome work @Justineo

portikM (approved):
LGTM 🚀

PR #1802, opened 2024-11-20
feat: new dashboard tile header

filipgutica:
# Summary

- introduce a header to dashboard tiles
- render title and kebab menu for tile actions in this header
- behind FF no need for major version upgrade

### DIFF
--- a/packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
+++ b/packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
@@ -3,7 +3,7 @@
     class="analytics-chart-shell"
   >
     <div
-      v-if="hasValidChartData && resultSetTruncated && maxEntitiesShown || (!hasKebabMenuAccess &&allowCsvExport) || props.chartTitle"
+      v-if="hasValidChartData && resultSetTruncated && maxEntitiesShown || (!hasKebabMenuAccess && (allowCsvExport || props.chartTitle))"
       class="chart-header"
     >
       <div
@@ -32,7 +32,7 @@
         </KTooltip>
       </div>
       <div
-        v-if="allowCsvExport && hasValidChartData"
+        v-if="!hasKebabMenuAccess && allowCsvExport && hasValidChartData"
         class="chart-export-button"
       >
         <CsvExportButton
@@ -129,11 +129,6 @@ const props = defineProps({
     required: false,
     default: false,
   },
-  goToExplore: {
-    type: String,
-    required: false,
-    default: '',
-  },
   chartData: {
     type: Object as PropType<ExploreResultV4>,
     required: true,
@@ -188,11 +183,6 @@ const props = defineProps({
     default: true,
   },
   timeseriesZoom: {
-    type: Boolean,
-    required: false,
-    default: false,
-  },
-  showMenu: {
     type: Boolean,
     required: false,
     default: false,
@@ -387,13 +377,6 @@ provide('legendPosition', toRef(props, 'legendPosition'))
   height: 100%;
   position: relative;
   width: 100%;
-
-  &.is-hovering {
-    .chart-header :deep(.popover-trigger-wrapper) {
-      opacity: 1;
-      visibility: visible;
-    }
-  }
 
   .analytics-chart-parent {
     height: inherit;
--- a/packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
+++ b/packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
@@ -1,23 +1,22 @@
 <template>
   <div
     class="analytics-chart-shell"
-    :class="{ 'is-hovering': isHovering }"
-    @mouseenter="handleMouseEnter"
-    @mouseleave="handleMouseLeave"
   >
     <div
+      v-if="hasValidChartData && resultSetTruncated && maxEntitiesShown || (!hasKebabMenuAccess &&allowCsvExport) || props.chartTitle"
       class="chart-header"
     >
       <div
-        v-if="chartTitle"
         class="chart-title"
         :title="chartTitle"
       >
         {{ chartTitle }}
       </div>
-      <div class="chart-header-icons-wrapper">
+      <div
+        v-if="resultSetTruncated && maxEntitiesShown"
+        class="chart-header-icons-wrapper"
+      >
         <KTooltip
-          v-if="hasValidChartData && resultSetTruncated && maxEntitiesShown"
           class="tooltip"
           max-width="500"
           placement="right"
@@ -33,7 +32,7 @@
         </KTooltip>
       </div>
       <div
-        v-if="allowCsvExport && hasValidChartData && !hasKebabMenuAccess"
+        v-if="allowCsvExport && hasValidChartData"
         class="chart-export-button"
       >
         <CsvExportButton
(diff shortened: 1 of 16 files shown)


adorack (changes requested):
Once component tests pass, I'd like to review.

mihai-peteu (dismissed):

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
adorack, line 1: Is the intent to roll AnalyticsChart back to how it was as of 7fa378462442693dff395737150a42e0862825b7?  If so, there are some differences -- for example, why is this still referring to `hasKebabMenuAccess`?  Why is the `resultSetTruncated` check no longer checking for valid chart data?  Why are there still references to `.is-hovering` in the CSS?

It might be better to just revert the file to 7fa3784 and see if that diff makes sense.  If the intent isn't to fully roll it back, it might be easier to talk about some of the differences.

packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -234,9 +234,8 @@ const getValue = (record: AnalyticsExploreRecord): string => {
 
 .kong-ui-public-top-n-table {
   border: none !important;
-  height: 100%;
+  max-height: 100%;
```
adorack, line 237: This seems like a meaningful change.  Isn't this going to break TopNTable styling in dashboards?

packages/analytics/dashboard-renderer/sandbox/pages/RendererDemo.vue
```diff
@@ -162,7 +162,7 @@ const dashboardConfig: DashboardConfig = {
     {
       definition: {
         chart: {
-          chartTitle: 'Timeseries line chart of mock data',
+          // chartTitle: 'Timeseries line chart of mock data',
```
adorack, line 165: Please remove dead code.

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -182,7 +182,6 @@ defineExpose({ refresh: refreshTiles })
     border: var(--kui-border-width-10, $kui-border-width-10) solid var(--kui-color-border, $kui-color-border);
     border-radius: var(--kui-border-radius-20, $kui-border-radius-20);
     height: 100%;
-    padding: var(--kui-space-70, $kui-space-70);
```
adorack, line 185: This change is not behind the feature flag and makes the gateway manager overview page look substantially different.

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -40,6 +114,36 @@ const emit = defineEmits<{
...
+    const datasource = ['advanced', 'ai'].includes(props.definition.query.datasource) ? props.definition.query.datasource : 'advanced'
+    return `${queryBridge.exploreBaseUrl()}?q=${JSON.stringify(exploreQuery)}&d=${datasource}&c=${props.definition.chart.type}`
+  }
+
+  return ''
+})
+
+const csvFilename = computed<string>(() => i18n.t('csvExport.defaultFilename'))
+const canShowKebabMenu = computed(() => hasKebabMenuAccess && !['golden_signals', 'top_n', 'gauge'].includes(props.definition.chart.type))
+const showDescriptionInHeader = computed(() => ['top_n', 'golden_signals'].includes(props.definition.chart.type))
```
adorack, line 145: Instead of special-casing this, would it be possible to only show the description if it's present?  Or would that break something I don't see?

adorack (changes requested):

packages/analytics/dashboard-renderer/sandbox/pages/RendererDemo.vue
filipgutica: done

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
filipgutica: done

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
filipgutica: done

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
```diff
@@ -1,23 +1,22 @@
...
-    :class="{ 'is-hovering': isHovering }"
-    @mouseenter="handleMouseEnter"
-    @mouseleave="handleMouseLeave"
   >
     <div
+      v-if="hasValidChartData && resultSetTruncated && maxEntitiesShown || (!hasKebabMenuAccess && (allowCsvExport || props.chartTitle))"
       class="chart-header"
     >
       <div
-        v-if="chartTitle"
```
adorack, line 12: This `v-if` still needs to be there to avoid rendering an extra div.

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
```diff
@@ -490,10 +393,8 @@ provide('legendPosition', toRef(props, 'legendPosition'))
...
     display: flex;
     justify-content: flex-start;
-    padding-bottom: var(--kui-space-60, $kui-space-60);
-    position: v-bind('chartHeaderPosition');
     right: 0;
-    width: v-bind('chartHeaderWidth');
+    width: 100%;
     z-index: 999;
 
     &:hover :deep(.popover-trigger-wrapper) {
```
adorack, line 400: This style doesn't need to be there anymore.

packages/analytics/analytics-chart/src/components/TopNTable.vue
adorack: Double-checked; looks good.

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
adorack: I don't see any change?

packages/analytics/dashboard-renderer/src/components/GoldenSignalsRenderer.vue
```diff
@@ -69,10 +74,15 @@ const options = computed<ProviderProps>(() => {
 
 <style scoped lang="scss">
 .metric-card-tile-wrapper {
+  align-items: center;
+  display: flex;
+  height: 75%;
```
adorack, line 79: This looks like it'll also have an un-feature-flagged impact on appearance.

adorack (changes requested):

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
filipgutica: done

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
filipgutica: done

packages/analytics/dashboard-renderer/src/components/GoldenSignalsRenderer.vue
filipgutica: It doesn't actually, if you try it out. What it does is just better center the metric cards in the tile, when the FF is on and the tile header is rendered and padding is tightened up. When you don't have the FF, I cannot notice any difference as a result of this change from what's in main.

If you disable the FF it looks like this with the change, and I don't notice any difference from what's currently in main.
image

However, without this change, if you enable the FF it looks like this, notice how the metrics card is no longer well centered. 
image

filipgutica: Ok, see my latest push I made any style changes to the golden signals renderer be behind the FF.

adorack (commented):
One comment while I'm checking the FF stuff.

adorack (approved):

PR #1809, opened 2024-11-25
fix(forms): number values in map fields

Leopoldthecoder:
# Summary

Problematic field example:
```json
 {
    "trace_services": {
      "description": "Map of routes matchers with sampling rates, use when plugin mode is global",
      "keys": {
        "type": "string"
      },
      "type": "map",
      "values": {
        "between": [
          0,
          1
        ],
        "default": 0,
        "required": true,
        "type": "number"
      }
    }
  }
```

This PR ensures the form respects `"type": "number"` and `"default": 0`.

FTI-6360

### DIFF
--- a/packages/core/forms/src/components/fields/FieldObjectAdvanced.vue
+++ b/packages/core/forms/src/components/fields/FieldObjectAdvanced.vue
@@ -220,7 +220,9 @@ export default {
         this.subSchema.fields.length === 1 &&
         this.subSchema.fields[0].type === 'array'
       const defaultValue = this.schema.values?.default ?? (this.valueInputType === 'number' ? 0 : '')
-      const initialValue = this.subSchema ? valueIsArray ? [''] : {} : defaultValue
+      const initialValue = this.subSchema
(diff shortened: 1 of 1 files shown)


packages/core/forms/src/components/fields/FieldObjectAdvanced.vue
```diff
@@ -216,9 +219,10 @@ export default {
         Array.isArray(this.subSchema.fields) &&
         this.subSchema.fields.length === 1 &&
         this.subSchema.fields[0].type === 'array'
-      const type = this.subSchema ? valueIsArray ? [''] : {} : ''
+      const defaultValue = this.schema.values?.default ?? ''
```
hangrao, line 222: If value's type is not string, would it cause problem to set default value as ''?

Leopoldthecoder: Updated.

sumimakito (approved):
Manually verified the fix. LGTM

PR #1821, opened 2024-11-27
fix(i18n): formatIsoDate handles trailing 0s

andrewwylde:
This pull request includes changes to improve the date formatting in the i18n module. The most important changes include adding new test cases for various ISO date formats and modifying the `formatIsoDate` function to exclude milliseconds.

This resolves a bug where a timestamp with trailing 0s will result in dividing the number by 1000 twice, resulting in an incorrect date.

---
Improvements to date formatting:

* [`packages/core/i18n/src/i18n.spec.ts`](diffhunk://#diff-39efa8e72b2e74acd22901626beabcd9d5521455aaa59f62b17b491f0e39858bR161-R167): Added test cases to ensure `formatIsoDate` correctly formats ISO dates with and without milliseconds, and with different time zone offsets.

Codebase simplification:

* [`packages/core/i18n/src/i18n.ts`](diffhunk://#diff-e7652b19e52ba221117c0d8d125453eae4868c2e8daa82bab1fc6a16dbc01b65L96-R98): Modified the `formatIsoDate` function to exclude milliseconds by flooring the date value before formatting it.

### DIFF
--- a/packages/core/i18n/src/i18n.spec.ts
+++ b/packages/core/i18n/src/i18n.spec.ts
@@ -158,6 +158,13 @@ describe('i18n', () => {
       const decimalTimestamp = formatIsoDate('2019-10-03T14:13:15.000Z')
       const integerTimestamp = formatIsoDate('2019-10-03T14:13:15.000Z')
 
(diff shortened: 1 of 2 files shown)


kaiarrowood (approved):
LGTM! 🚀

PR #1824, opened 2024-11-28
fix(rla-form): place enable switch in top of form

TT1228:
# Summary

KM-752

Place the enable switch from hidden advanced fields into top of form like other plugins for rate limiting advanced form.
Before:
截屏2024-11-28 下午12 54 11

After: 
截屏2024-11-28 下午1 39 58

### DIFF
--- a/packages/core/forms/src/components/forms/RLAForm.vue
+++ b/packages/core/forms/src/components/forms/RLAForm.vue
@@ -3,7 +3,7 @@
     <VueFormGenerator
       :model="formModel"
       :options="formOptions"
-      :schema="scopingSchema"
+      :schema="globalFields"
       @model-updated="(value: any, model: string) => onModelUpdated(value, model)"
     />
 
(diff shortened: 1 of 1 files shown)


packages/core/forms/src/components/forms/RLAForm.vue
```diff
@@ -337,7 +337,9 @@ const scopingSchema = computed(() => {
     return undefined
   }
 
-  return { fields: [selectionGroup] }
+  const enableSwitch = props.formSchema?.fields.find((field: any) => field.pinned && field.type === 'switch' && field.model === 'enabled')
```
sumimakito, line 340: nit: Could this be simplified to something like:

```suggestion
  const enableSwitch = props.formSchema?.fields.find((field: any) => field.model === 'enabled')
```

sumimakito (dismissed):
LGTM. I left a non-blocking comment.

Leopoldthecoder (approved):
LGTM

PR #1827, opened 2024-12-02
feat(analytics): add new granularities with support for FF [MA-2149]

adorack:
- Add new granularity values and definitions
- Add support for passing one feature flag to query time and `allowedGranularities`
- Add new timeframe -> granularity mappings

# Summary

### DIFF
--- a/packages/analytics/analytics-utilities/src/queryTime.ts
+++ b/packages/analytics/analytics-utilities/src/queryTime.ts
@@ -12,8 +12,9 @@ import type { Timeframe } from './timeframes'
 abstract class BaseQueryTime implements QueryTime {
   protected readonly timeframe: Timeframe
   protected readonly tz?: string
+  protected readonly dataGranularity: GranularityValues
 
-  constructor(timeframe: Timeframe, tz?: string) {
+  constructor(timeframe: Timeframe, tz?: string, dataGranularity?: GranularityValues) {
     // This is an abstract class.
     if (this.constructor === BaseQueryTime) {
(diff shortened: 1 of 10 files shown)


packages/analytics/analytics-utilities/src/queryTime.ts
```diff
@@ -79,11 +81,16 @@ abstract class BaseQueryTime implements QueryTime {
...
+  constructor(timeframe: Timeframe, granularity?: GranularityValues, tz?: string, dataGranularity?: GranularityValues, fineGrain?: boolean) {
+    super(timeframe, tz, dataGranularity)
 
-    if (granularity && timeframe.allowedGranularities().has(granularity)) {
+    if (granularity && timeframe.allowedGranularities(fineGrain).has(granularity)) {
       this.granularity = granularity
+    } else if (fineGrain) {
+      // TODO: when removing the feature flag, consider redefining `defaultResponseGranularity`
+      // in the timeframes constructor: it should probably handle this calculation on its own.
+      const finestGranularity = timeframe.allowedGranularities(fineGrain).keys().next().value
```
hfukada, line 92: nit: feels almost like this should just be a list instead of a Set

adorack: Hmm.  I think that's fair, but would prefer not to change it right now since it's a breaking change to the API.  I think a lot of this needs an overhaul, and plan to advocate for that overhaul when we do fine-grained absolute time.

hfukada (approved):

PR #1829, opened 2024-12-03
fix(tracing): fix height for skeleton container

sumimakito:
# Summary

This pull request fixes the height for skeleton containers to avoid shaking while switching between the ON and OFF states of skeletons in `TraceViewer`.


Leopoldthecoder (approved):
LGTM

PR #1833, opened 2024-12-04
fix(spec-renderer): tolowercase is not a function [KHCP-14300]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-14300

### DIFF
--- a/packages/portal/spec-renderer/src/components/SpecOperationsList.vue
+++ b/packages/portal/spec-renderer/src/components/SpecOperationsList.vue
@@ -32,7 +32,7 @@
           v-for="section in sectionHeadings"
           :key="section"
           class="section-wrapper"
-          :data-testid="`spec-operations-list-section-${section.toLowerCase().replaceAll(' ', '-')}`"
+          :data-testid="`spec-operations-list-section-${String(section).toLowerCase().replaceAll(' ', '-')}`"
         >
           <KCollapse
             :model-value="false"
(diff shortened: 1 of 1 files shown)


packages/portal/spec-renderer/src/components/SpecOperationsList.vue
```diff
@@ -32,7 +32,7 @@
           v-for="section in sectionHeadings"
           :key="section"
           class="section-wrapper"
-          :data-testid="`spec-operations-list-section-${section.toLowerCase().replaceAll(' ', '-')}`"
+          :data-testid="`spec-operations-list-section-${String(section).toLowerCase().replaceAll(' ', '-')}`"
```
adamdehaven, line 35: I don't understand what this fixes? If `section` is undefined, you're going to be trimming a string of `'undefined'`

adamdehaven (changes requested):

packages/portal/spec-renderer/src/components/SpecOperationsList.vue
portikM: It shouldn't be undefined as in `sectionHeadings` computed checks whether it's not falsy here: https://github.com/Kong/public-ui-components/blob/5c52d1fa5ac3ad6708911374d2794f35000e06f3/packages/portal/spec-renderer/src/components/SpecOperationsList.vue#L229

portikM: I'm not sure how this error happened but it happened. I reworked my fix a bit

adamdehaven (approved):

PR #1834, opened 2024-12-05
chore(viteconfig): add me in availableRegions

hangrao:
# Summary

Add me (Middle East) support in availableRegion,  it has announced GA release in Nov.

### DIFF
--- a/vite.config.shared.ts
+++ b/vite.config.shared.ts
@@ -160,7 +160,7 @@ export const getApiProxies = (pathToRoot: string = '../../../.') => {
     : undefined
 
   // Add additional regions as they become available
-  const availableRegions = ['au', 'eu', 'us', 'me', ‘in’]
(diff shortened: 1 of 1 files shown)


Leopoldthecoder (dismissed):

vite.config.shared.ts
```diff
@@ -160,7 +160,7 @@ export const getApiProxies = (pathToRoot: string = '../../../.') => {
     : undefined
 
   // Add additional regions as they become available
-  const availableRegions = ['au', 'eu', 'us']
+  const availableRegions = ['au', 'eu', 'us', 'me']
```
adamdehaven, line 163: Let’s go ahead and add the India region as well
```suggestion
  const availableRegions = ['au', 'eu', 'us', 'me', ‘in’]
```

hangrao: Updated with adding 'in'

PR #1843, opened 2024-12-09
feat: add filtering for country code

alxgrant:
# Summary

### DIFF
--- a/packages/analytics/analytics-utilities/src/types/explore/advanced.ts
+++ b/packages/analytics/analytics-utilities/src/types/explore/advanced.ts
@@ -6,6 +6,7 @@ export const queryableExploreDimensions = [
   ...queryableBasicExploreDimensions,
   'application',
   'consumer',
+  'country_code',
   'iso_code',
 ] as const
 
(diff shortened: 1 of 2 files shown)


packages/analytics/analytics-utilities/src/types/explore/advanced.ts
```diff
@@ -6,6 +6,7 @@ export const queryableExploreDimensions = [
   ...queryableBasicExploreDimensions,
   'application',
   'consumer',
+  'country_code',
   'iso_code',
```
adorack, line 10: Just to check: is `iso_code` going to go away?  I'm not up to date on the status.

adorack (approved):

packages/analytics/analytics-utilities/src/types/explore/advanced.ts
alxgrant: Yes it will, but right now it is still being used by the front end in both explore and api requests. I've updated the FE ticket to include a comment to remove `iso_code` in favor of `country_code` when updating the filtering. After that we can remove. https://konghq.atlassian.net/browse/MA-3448

PR #1855, opened 2024-12-17
feat(tracing): allow zooming in trace viewer

sumimakito:
# Summary

This pull request adds support for zooming interaction to the trace viewer.

https://github.com/user-attachments/assets/7eab6362-4a44-40cf-b0be-aa6bfcc2c6b0

- The `left` and `right` attributes of span bars are clamped within the viewport to ensure performance under a high zoom level.
- We will show color at the edge when the view is zoomed in and the span is out of the viewport to indicate that the span is in that direction.
- Cleaned up unused config `viewportShift` in the waterfall view

KM-755


Leopoldthecoder (approved):
Manually tested and LGTM

PR #1856, opened 2024-12-17
fix(tracing): handle missing fields in span

sumimakito:
# Summary

Screenshot 2024-12-18 at 10 43 11

- Marking some fields as optional as they might be missing in real-world use cases
- Better handling of these cases and showing an icon to indicate the span is potentially malformed
- Providing an alert in the span detail pane when the span is potentially malformed
- Type and misc fixes

KM-824


Leopoldthecoder (dismissed):
LGTM

Leopoldthecoder (approved):
LGTM

PR #1869, opened 2024-12-20
fix(tracing): check i18n key in a better way

sumimakito:
# Summary

This pull request updates the logic to check the availability of an i18n key to avoid printing errors in the console:

Screenshot 2024-12-20 at 14 44 20


Leopoldthecoder (approved):
LGTM

PR #1870, opened 2024-12-20
fix(tracing): map some span names to their new names

sumimakito:
# Summary

This pull request maps the names of some spans to their new names.

KM-837


Leopoldthecoder (approved):
LGTM

PR #1871, opened 2024-12-23
fix(tracing): overall and per-span latency display

sumimakito:
# Summary

- Updates the overall and per-span latency display
  Screenshot 2024-12-23 at 19 55 33
- Exported span attributed were updated and aligned with Gateway
- Legend colors were removed to avoid confusion
- Will round-up and format latencies in milliseconds with one decimal place

KM-822 + KM-838


Leopoldthecoder (approved):
LGTM

PR #1872, opened 2024-12-24
fix(analytics-chart): render timeseries charts with a single point [MA-2682]

ryanmoore:
# Summary

When a line chart is asked to display a single data point, display that datapoint with a radius smaller than the hover radius instead of displaying a deceptive "No data" error.

Test dataset:
```
{
  "data": [
    {
      "event": {
        "request_count": 100000
      },
      "timestamp": "2024-12-23T07:00:00.000Z"
    }
  ],
  "meta": {
    "start_ms": 1734937200000,
    "end_ms": 1735023600000,
    "limit": 50,
    "granularity_ms": 86400000,
    "metric_names": [
      "request_count"
    ],
    "truncated": false,
    "display": {},
    "metric_units": {
      "request_count": "count"
    }
  }
}
```

Before:
image

After:

image

### DIFF
--- a/packages/analytics/analytics-chart/README.md
+++ b/packages/analytics/analytics-chart/README.md
@@ -510,3 +510,11 @@ const fileName = ref('exportFilename')
 </script>
 
 ```
+
+## Development
+
+Follow the setup instructions for the [top level README](../../../README.md). Then run the sandbox with:
+
+```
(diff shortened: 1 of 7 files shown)


ryanmoore:
reworking based on feedback from @filipgutica

PR #1873, opened 2024-12-25
feat: KM-235 redis form plugin adapation

TT1228:
# Summary

This PR applies redis partials to plugins


packages/core/forms/src/components/FormGenerator.vue
```diff
@@ -11,8 +11,21 @@
         v-for="field in fields"
         :key="field.model"
       >
+        <form-redis
+          v-if="field.model === '__redis_partial' && enableRedisPartial"
```
raoxiaoyan, line 15: Why `__redis_partial`? It looks very wired. Is there any particular consideration?

PR #1875, opened 2024-12-26
feat(plugin-select): enable scrolling to anchor

TT1228:
# Summary

When cancelling edition/creation of a plugin, navigate back to plugin select list and scroll to the plugin card.

https://github.com/user-attachments/assets/077365d6-d9b7-497f-91b3-c1d8ed9d8818


Justineo:
How about we navigate back to the plugin selection list first and figure out how to restore its internal states (scroll position, filtering, expand/collapse states) later?

PR #1877, opened 2025-01-02
fix(tracing): hide basic info when no description is available

sumimakito:
# Summary

This pull request hides the basic info section when no description is available for the selected span to save space.

KM-849


Leopoldthecoder (approved):
LGTM

PR #1878, opened 2025-01-02
fix(entities-consumers): provide correct description for cascade delete

TT1228:
# Summary

KM-862

Provide confirm texts for cascade delete of plugins in consumer and consumer_group.


Leopoldthecoder (approved):
LGTM

PR #1880, opened 2025-01-06
ci: rename workflow jobs [KHCP-14529]

adamdehaven:
# Summary

Renaming workflow jobs in order to manually disable old workflows that used `pull_request_trigger`

### DIFF
--- a/.github/workflows/pr-on-close.yaml
+++ b/.github/workflows/pr-on-close.yaml
@@ -1,4 +1,4 @@
-name: On PR close
+name: PR on close
 on:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-on-close.yaml
```diff
@@ -1,4 +1,4 @@
-name: On PR close
+name: PR on close
```
jillztom, line 1: The name reads a bit weird.

jillztom (approved):

PR #1882, opened 2025-01-06
feat(app-layout): expose top offset

adamdehaven:
# Summary

Expose the `topOffset` from the `AppLayout` - can be used to determine the top offset for downstream apps

## Example Usage

```vue
<template>
  <AppLayout ref="appLayout" />
</template>

<script setup lang="ts">
const appLayout = useTemplateRef('appLayout')

// Calculate the top offset for the Learning Hub slideout, always adding 60 to account for the height of the navbar
const learningHubSlideoutTopOffset = computed((): number => (appLayout.value?.topOffset || 0) + 60)
</script>
```


aanchalm01 (approved):
lgtm!

PR #1885, opened 2025-01-09
feat(field-multiselect): setup syntax [khcp-11341]

kaiarrowood:
# Summary

Convert KMultiselect to setup syntax for [KHCP-11341](https://konghq.atlassian.net/browse/KHCP-11341).

Other fixes:
- Add testing of style classes to other entities
- add support for missing props to FieldMultiselect:
  - `id`
  - `classNames`
  - `disabled`
  - `name`
  - add `data-testid` to KPop

[KHCP-11341]: https://konghq.atlassian.net/browse/KHCP-11341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/forms/src/components/fields/FieldMultiselect.vue
+++ b/packages/core/forms/src/components/fields/FieldMultiselect.vue
@@ -17,7 +17,7 @@
 </template>
 
 <script lang="ts" setup>
-import { computed, toRefs, type PropType } from 'vue'
+import { computed, toRef, type PropType } from 'vue'
 import type { MultiselectItem } from '@kong/kongponents'
 import composables from '../../composables'
 
@@ -60,10 +60,10 @@ const emit = defineEmits<{
   (event: 'modelUpdated', value: any, model: Record<string, any>): void
 }>()
 
-const propsRefs = toRefs(props)
+const modelRef = toRef(props)
 
 const { getLabelId, getFieldID, clearValidationErrors, value: inputValue } = composables.useAbstractFields({
-  model: propsRefs.model,
+  model: modelRef,
   schema: props.schema,
   formOptions: props.formOptions,
   emitModelUpdated: (data: { value: any, model: Record<string, any> }): void => {
--- a/packages/core/forms/src/components/fields/FieldMultiselect.vue
+++ b/packages/core/forms/src/components/fields/FieldMultiselect.vue
@@ -1,41 +1,89 @@
 <template>
   <KMultiselect
+    :id="getFieldID(schema)"
+    v-model="inputValue"
     :aria-labelledby="getLabelId(schema)"
+    :class="schema.fieldClasses"
     data-testid="field-multiselect"
+    :disabled="disabled || undefined"
     :items="items"
+    :kpop-attributes="{ 'data-testid': `${getFieldID(schema)}-items` }"
     :label-attributes="{ info: schema.help }"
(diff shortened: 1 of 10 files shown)


packages/core/forms/src/components/fields/FieldMultiselect.vue
```diff
@@ -1,41 +1,89 @@
...
+   */
+  errors: {
+    type: Array,
+    default: () => [],
+  },
+  hint: {
+    type: String,
+    default: '',
+  },
+})
```
Justineo, line 57: Do we still prefer value syntax instead of type syntax for `defineProps`?

packages/core/forms/src/components/fields/FieldMultiselect.vue
```diff
@@ -1,41 +1,89 @@
...
-      if (this.schema.values) {
-        return this.schema.values
-      }
-      if (this.schema.elements?.one_of?.length) {
-        return this.schema.elements.one_of.map(value => ({ label: value, value }))
-      }
-      return []
-    },
-  },
+const propsRefs = toRefs(props)
```
Justineo, line 63: If we are only using `propsRefs.model`, maybe it's better to use `const model = toRef(props, 'model')`.

Leopoldthecoder (approved):
Manually tested and LGTM

packages/core/forms/src/components/fields/FieldMultiselect.vue
```diff
@@ -1,41 +1,89 @@
...
+  clearValidationErrors,
+})
+
+const items = computed((): MultiselectItem[] => {
+  if (props.schema.values) {
+    return props.schema.values
+  }
+
+  if (props.schema.elements?.one_of?.length) {
+    return props.schema.elements.one_of.map((value: string | number | boolean) => ({ label: String(value), value: String(value) } as MultiselectItem))
```
Justineo, line 84:
```suggestion
    return props.schema.elements.one_of.map((value: string | number | boolean) => ({ label: String(value), value: String(value) } satisfies MultiselectItem))
```

packages/core/forms/src/components/fields/FieldMultiselect.vue
```diff
@@ -1,41 +1,89 @@
 <template>
   <KMultiselect
+    :id="getFieldID(schema)"
+    v-model="inputValue"
     :aria-labelledby="getLabelId(schema)"
+    :class="schema.fieldClasses"
     data-testid="field-multiselect"
+    :disabled="disabled || undefined"
```
Justineo, line 8:
```suggestion
    :disabled="disabled"
```

I believe it's unnecessary to fallback to `undefined` here?

Justineo (commented):
Could you also describe any other changes made in addition to the refactoring to `<script setup>` in the PR?

packages/core/forms/src/components/fields/FieldMultiselect.vue
kaiarrowood: I'm not sure there is a preference? The rest of the files in there are using `defineProps` so that's what I went with.

Justineo (dismissed):
LGTM!

PR #1888, opened 2025-01-10
fix(app-layout): add pageinfosection component [KHCP-14578]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-14578

Adds CollapsibleSection component exported by `app-layout` package

### DIFF
--- a/packages/core/app-layout/docs/page-info-section.md
+++ b/packages/core/app-layout/docs/page-info-section.md
@@ -1,6 +1,6 @@
-# CollapsibleSection.vue
+# PageInfoSection.vue
 
-A Kong UI simple collapsible section component.
+A Kong UI simple optionally collapsible section component.
 
 - [Requirements](#requirements)
 - [Usage](#usage)
--- a/packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
+++ b/packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
@@ -1,0 +1,156 @@
+<template>
+  <component
+    :is="collapsible ? 'details' : 'div'"
+    class="collapsible-section"
+    data-testid="collapsible-section"
+  >
+    <component
+      :is="collapsible ? 'summary' : 'div'"
+      class="collapsible-section-header"
+      data-testid="collapsible-section-header"
+    >
+      <slot name="header">
+        <div class="collapsible-section-default-header">
+          <div
+            v-if="title"
+            class="collapsible-section-title"
+          >
+            {{ title }}
+          </div>
+          <div
+            v-if="description"
+            class="collapsible-section-description"
+          >
+            {{ description }}
+          </div>
+        </div>
+      </slot>
+
+      <ChevronRightIcon
+        v-if="collapsible"
+        class="collapsible-section-chevron-icon"
+        decorative
+      />
+      <slot
+        v-else
+        name="actions"
+      />
+    </component>
+
+    <div
+      v-if="$slots.default"
+      class="collapsible-section-content"
+      data-testid="collapsible-section-content"
+    >
+      <slot />
+    </div>
+  </component>
+</template>
+
+<script setup lang="ts">
+import { ChevronRightIcon } from '@kong/icons'
+
+defineProps({
+  collapsible: {
+    type: Boolean,
+    default: true,
+  },
+  title: {
+    type: String,
+    default: '',
+  },
+  description: {
+    type: String,
+    default: '',
+  },
+})
+</script>
+
+<style lang="scss" scoped>
+.collapsible-section {
+  border: $kui-border-width-10 solid $kui-color-border;
+  border-radius: $kui-border-radius-30;
+
+  .collapsible-section-header {
+    padding: $kui-space-70;
+    align-items: center;
+    cursor: pointer;
+    display: flex;
+    gap: $kui-space-20;
+    justify-content: space-between;
+
+    .collapsible-section-default-header {
+      display: flex;
+      flex-direction: column;
+      gap: $kui-space-40;
+      max-width: 700px;
+
+      .collapsible-section-title {
+        color: $kui-color-text;
+        font-size: $kui-font-size-40;
+        font-weight: $kui-font-weight-bold;
+        line-height: $kui-line-height-30;
+      }
+
+      .collapsible-section-description {
+        color: $kui-color-text-neutral;
+        font-size: $kui-font-size-30;
+        line-height: $kui-line-height-30;
+        margin: $kui-space-0;
+      }
+    }
+  }
+
+  .collapsible-section-content {
+    background: $kui-color-background-neutral-weakest;
+    border-top: $kui-border-width-10 solid $kui-color-border;
+    padding: $kui-space-70;
+    display: flex;
+    flex-direction: column;
+    gap: $kui-space-40;
+
+    :deep(.k-table-view) {
+      background-color: var(--kui-color-background-neutral-weakest, $kui-color-background-neutral-weakest);
+
+      tr.is-scrolled {
+        background-color: var(
+          --kui-color-background-neutral-weakest,
+          $kui-color-background-neutral-weakest
+        );
+      }
+    }
+
+    :deep(.k-empty-state) {
+      background-color: $kui-color-background-neutral-weakest;
+    }
+  }
+}
+
+details.collapsible-section {
+  &[open] {
+    .collapsible-section-chevron-icon {
+      transform: rotate(90deg);
+    }
+  }
+
+  &:has(>summary:focus-visible) {
+    box-shadow: $kui-shadow-focus;
+  }
+
+  summary {
+    outline: none;
+
+    &::-webkit-details-marker, &::marker {
+      display: none;
+    }
+  }
+
+  .collapsible-section-header {
+    .collapsible-section-chevron-icon {
+      flex-shrink: 0;
+      pointer-events: none;
+      transition: transform $kui-animation-duration-20 ease;
+    }
+  }
+}
+</style>
--- a/packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
+++ b/packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
@@ -11,12 +11,13 @@
     >
       <slot name="header">
         <div class="collapsible-section-default-header">
-          <div
+          <component
+            :is="titleTag"
             v-if="title"
             class="collapsible-section-title"
           >
             {{ title }}
-          </div>
+          </component>
           <div
             v-if="description"
             class="collapsible-section-description"
@@ -48,7 +49,10 @@
 </template>
 
 <script setup lang="ts">
+import type { PropType } from 'vue'
 import { ChevronRightIcon } from '@kong/icons'
+import type { HeaderTag } from '@kong/kongponents'
+import { HeaderTags } from '@kong/kongponents'
 
 defineProps({
   collapsible: {
@@ -63,6 +67,11 @@ defineProps({
     type: String,
     default: '',
   },
+  titleTag: {
+    type: String as PropType<HeaderTag>,
+    default: 'div',
+    validator: (value: HeaderTag): boolean => HeaderTags.includes(value),
+  },
 })
 </script>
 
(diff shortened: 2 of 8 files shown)


packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
```diff
@@ -0,0 +1,156 @@
+<template>
+  <component
+    :is="collapsible ? 'details' : 'div'"
```
adamdehaven, line 3: Since you're using `details` and `summary` can you please functionally test in Safari as well? 

If I recall there were some styling issues that popped up in spec renderer specifically around these elements in Safari

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
```diff
@@ -0,0 +1,156 @@
...
+      data-testid="collapsible-section-header"
+    >
+      <slot name="header">
+        <div class="collapsible-section-default-header">
+          <div
+            v-if="title"
+            class="collapsible-section-title"
+          >
+            {{ title }}
+          </div>
```
adamdehaven, line 19: suggestion: should this be an actual header tag? e.g. `h3`

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
```diff
@@ -0,0 +1,156 @@
...
+        name="actions"
+      />
+    </component>
+
+    <div
+      v-if="$slots.default"
+      class="collapsible-section-content"
+      data-testid="collapsible-section-content"
+    >
+      <slot />
```
adamdehaven, line 45: nit: just for readability
```suggestion
      <slot name="default" />
```

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  &:has(>summary:focus-visible) {
+    box-shadow: $kui-shadow-focus;
+  }
+
+  summary {
+    outline: none;
+
+    &::-webkit-details-marker, &::marker {
```
adamdehaven, line 143: I think this covers my Safari concern, but wouldn't hurt just to visually verify in the browser 🙏🏼

adamdehaven (commented):
suggestion: can you add this to the appropriate README documentation in the package?

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
portikM: Confirmed it looks and works in Safari

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
portikM: Added a `titleTag` prop

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
portikM: Done

portikM:
> suggestion: can you add this to the appropriate README documentation in the package?

Done

vaibhavrajsingh2001 (commented):
LGTM, just one comment about the naming

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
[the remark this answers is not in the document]
portikM: Valid point but `PageSection` doesn't suggest that it can be collapsible, engineers would have to know it is from previous experience in order to know to use this component when they need collapsible section.. not sure what's the right approach here. @adamdehaven thought?

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
johncowen: What do you think of having a slot to control this instead of multiple props? Then I can just `<template #title><h3 id="bookmarkable-h3">{{ title }}</h3></template>`

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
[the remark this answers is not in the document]
johncowen: Great point 👏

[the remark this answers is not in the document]
adamdehaven: How about something like `PageConfigSection` or `PageInfoSection`; something that conveys the use case? 

`PageSection` is a bit too vague IMO. 

Knowing that it can be collapsible is really just a documentation issue.

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
adamdehaven: It is slottable `<slot name="header">`

johncowen:
How would I add an icon here?

Screenshot 2025-01-10 at 16 53 49

Do we support collapsible sections with icons?

Maybe this is all TBD, but lemme know!

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
vaibhavrajsingh2001: There's already a `header` slot for this. It wraps only the title and description.

adamdehaven: Are you saying You'd like separate slots for title and description?

johncowen: Well I'm not totally sure what I'm saying, I guess if I use the header slot it looks like I lose the description, and if we are expected to just use the header slot, why not just make that the single way to do things and not have `title` and `titleTag`?

vaibhavrajsingh2001:
> How would I add an icon here?
> 
> Screenshot 2025-01-10 at 16 53 49
> 
> Do we support collapsible sections with icons?
> 
> Maybe this is all TBD, but lemme know!

For the image you shared, the icon would go in the `actions` slot and it'll be a non-collapsible section.
If an icon is supposed to be there then it shouldn't be a collapsible section, because where would the chevron go.

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
adamdehaven: it's if you don't want to use the slot. Utilizing props allows the component to control the styles rather than depending on consistent slot usage

johncowen: But if I use the `header` slot you are still not controlling the styles? Also, not totally sure but isn't `::slotted` available to us to influence styling? At least if you have a single way to do this (instead of several), you aren't assuming you are controlling styles when you are possibly not?

johncowen: > We can't use ::slotted -- we don't know what the slot contents are.

What styles are we wanting to control in a theoretical title slot, just the font styles? Would setting the font-styles on `::slotted` work for everything/all tags? I'm not totally sure what effect that would have without trying it out.

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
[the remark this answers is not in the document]
portikM: Renamed to `PageInfoSection`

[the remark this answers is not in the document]
johncowen: How about KSection? Come to think of it how come this isn't being added to kongponents? Feels like it should live over there?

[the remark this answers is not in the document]
adamdehaven: Just noticed we typically prefix with `App*` in this repo. `AppInfoSection` maybe?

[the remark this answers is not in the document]
adamdehaven: > How about KSection? Come to think of it how come this isn't being added to kongponents? Feels like it should live over there?

This is Konnect specific and will likely change over time.

johncowen:
> For the image you shared, the icon would go in the actions slot and it'll be a non-collapsible section.

Ah missed that, thanks!

portikM:
> > How would I add an icon here?
> > Screenshot 2025-01-10 at 16 53 49
> > Do we support collapsible sections with icons?
> > Maybe this is all TBD, but lemme know!
> 
> For the image you shared, the icon would go in the `actions` slot and it'll be a non-collapsible section. If an icon is supposed to be there then it shouldn't be a collapsible section, because where would the chevron go.

Just to add to this point: if the section was collapsible but didn't have a chevron, there would be no visual indication to user that it is collapsible. That's why we need the chevron there when it is collapsible.

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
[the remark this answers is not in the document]
portikM: > Just noticed we typically prefix with App* in this repo. AppPageConfigSection maybe?

Don't want to use the word config so it doesn't get mixed up with EntityConfigCard but AppPageInfoSection sounds good.

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
portikM: > is there a use-case for slotting the header in the first place?

There is in notifications MFE

Screenshot 2025-01-10 at 12 26 56 PM
Screenshot 2025-01-10 at 12 27 11 PM

vaibhavrajsingh2001 (approved):

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
johncowen: Cheers @portikM 👍 

I'd totally avoid trying to control too much styling of consumer content via using props then, with too much control you lose flexibility, I guess thats all a fine balance. But maybe thats just me

packages/core/app-layout/docs/page-info-section.md
```diff
@@ -0,0 +1,74 @@
+# PageInfoSection.vue
```
adamdehaven, line 1: You didn't update the docs for the naming, etc.

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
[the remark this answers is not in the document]
portikM: This PR was merged by mistake (I forgot to disable auto-merge). If there is anything to follow up on, please drop comments on https://github.com/Kong/public-ui-components/pull/1890

packages/core/app-layout/src/components/collapsibleSection/CollapsibleSection.vue
portikM: This PR was merged early by mistake (I forgot to disable auto-merge). If there is anything to follow up on, please drop comments on https://github.com/Kong/public-ui-components/pull/1890

johncowen: No prob! Umm did you have any thoughts on the props vs slots thing in relation to https://github.com/Kong/public-ui-components/pull/1888#discussion_r1910702471 or shall I move it to the other PR? FYI: Its totally fine to say, I hear your comment, just we don't want to get into it atm, we can discuss another time also.

adamdehaven: Here's my remaining question from this thread:

> is there a use-case for slotting the header in the first place? The whole point of the component is to allow for a consistent UI across apps, so if we can remove slots where not needed, that would be ideal TBH

johncowen: I suppose this is turning into a more general props vs slots comment, so I guess I'm asking whether we should generally default to using slots for these sorts of things instead of props because of the difficulties consumers have when they are just given props to work with. I suppose maybe depends if we are making these components to make the our lives easier or making these components to make consumers lives easier?

I guess theres maybe a lot of opinion here, so happy to finish this up and resolve this convo if nobody has any more to add?

PR #1896, opened 2025-01-15
feat(entities-plugin): add streaming custom plugin support

nekolab:
Changes include:
1. Remove `disableCustomPlugins` 
2. A new prop `customPluginSupport` is added to indicate the which type of custom plugin should be created for the CP
3. The plugin grid now renders schema-only and streaming custom plugins correctly.
4. `getCustomEditRoute` now provides `type: 'schema' | 'streaming'` for custom plugin type indication
5. Text and sandbox change

JIRA: KM-895

### DIFF
--- a/packages/entities/entities-plugins/docs/plugin-select.md
+++ b/packages/entities/entities-plugins/docs/plugin-select.md
@@ -61,7 +61,7 @@ A grid component for selecting Plugins.
     - The route for creating a custom plugin.
 
   - `getCustomEditRoute`:
-    - type: `(plugin: string) => RouteLocationRaw`
+    - type: `(plugin: string, type: 'schema' | 'streaming') => RouteLocationRaw`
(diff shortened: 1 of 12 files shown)


packages/entities/entities-plugins/docs/plugin-select.md
```diff
@@ -92,21 +92,19 @@ A grid component for selecting Plugins.
 
 The base konnect or kongManger config.
 
-#### `disableCustomPlugins`
+#### `customPluginSupport`
 
-- type: `boolean`
+- type: `false | 'disabled' | 'schema' | 'streaming'`
```
Justineo, line 97: Do you think it will be better to use `'hidden'` or `'none'` instead of keeping the boolean type here?

Justineo (approved):
LGTM

PR #1907, opened 2025-01-22
feat(expressions): add func to check if rhs autocompletion is necessary; dispose after use

sumimakito:
# Summary

* `<ExpressionsEditor />` now accepts `rhsValueCompletion` which is made up of `provide` and `shouldProvide`, where the 2nd function accepts an LHS identifier value and returns a `boolean` to indicate whether the RHS value autocompletion is necessary.
* Disposables associated with a `languageId` will now be disposed upon the next registration with the same `languageId`


Leopoldthecoder (approved):
LGTM. Thanks!

PR #1908, opened 2025-01-22
fix(tracing): lifecycle view and tab partials

sumimakito:
# Summary

* `<LifecycleView />` to show request lifecycles in flowcharts
* `<PayloadDisplay />` to show payload info like request/response headers/body
* `<SummaryViewTab />` to build up the tab for request lifecycle, info, and captured payload

KM-773


packages/core/tracing/src/partials/SummaryViewTab.vue
```diff
@@ -0,0 +1,171 @@
...
+      size="50"
+    >
+      <template v-if="rootSpan">
+        <RequestInfo
+          :root-span="rootSpan"
+          :show-skeleton="showSkeleton"
+        />
+
+        <PayloadDisplay
+          v-if="payloads?.headers?.request"
```
Leopoldthecoder, line 29: Optional chaining doesn't break GitHub's syntax highlight anymore?

Leopoldthecoder:
The code block always shows a horizontal scrollbar. Can this be removed?

image

Leopoldthecoder:
Sometimes the request latency is out of view when I open the summary view. Reproduction link: https://pr-5577--gateway-manager.cloud-preview.konghq.tech/us/gateway-manager/03a13dd9-cf6c-4f8c-b2f8-1c3fc34a4f29/active-tracing/sessions/01949224-c46e-7dcd-a2c6-9de98afcd7a8/traces/9513ffd6b98ea780b538ebd7fd292890

image

sumimakito:
> Sometimes the request latency is out of view when I open the summary view…

@Leopoldthecoder Good catch. It seems that the `fitView` function provided by VueFlow would zoom in on the chart too much when we only have a few nodes. After adding `maxZoom: 1` to `fitView`, it looks like this:

|Before|After|
|:-|:-|
|Screenshot 2025-01-23 at 16 27 29|Screenshot 2025-01-23 at 16 26 41|

What do you think?

packages/core/tracing/src/partials/SummaryViewTab.vue
sumimakito: It seems to be so that I didn't even notice that… 🥹

Leopoldthecoder:
> > Sometimes the request latency is out of view when I open the summary view…
> 
> @Leopoldthecoder Good catch. It seems that the `fitView` function provided by VueFlow would zoom in on the chart too much when we only have a few nodes. After adding `maxZoom: 1` to `fitView`, it looks like this:
> 
> Before	After
> Screenshot 2025-01-23 at 16 27 29	Screenshot 2025-01-23 at 16 26 41
> What do you think?

I think now it's better: it gives users the whole picture. 👍

packages/core/tracing/src/types/payloads.ts
```diff
@@ -0,0 +1,12 @@
...
+export type Body = string
+
+export type Payload =
+  | {
+    type: 'headers';
+    headers: Headers;
+  }
+  | {
+    type: 'body';
+    content: Body;
```
Leopoldthecoder, line 11: Shall we remove `;` in this file?

PR #1912, opened 2025-01-23
fix: make sure timeperiods from the singleton map are cloned

filipgutica:
# Summary


packages/analytics/analytics-metric-provider/src/components/MetricsProvider.vue
```diff
@@ -101,7 +102,7 @@ const resolvedDatasource = computed<QueryDatasource>(() => {
 // needs to be reactive as well.  Ideally, this would be the case; we don't have any guarantee that the
 // tier data has finished loading by the time the component mounts, for example.
 const timeframe = computed<Timeframe>(() => {
-  return props.overrideTimeframe || TimePeriods.get(TimeframeKeys.SEVEN_DAY)!
+  return props.overrideTimeframe || cloneDeep(TimePeriods.get(TimeframeKeys.SEVEN_DAY)) as Timeframe
```
adorack, line 105: I'd prefer to avoid this `cloneDeep` unless we actually find evidence that this usage is unsafe.  The metric cards don't expose this timeframe or mutate it, so I'm inclined to believe it's safe.

packages/analytics/dashboard-renderer/src/components/GoldenSignalsRenderer.vue
```diff
@@ -43,7 +44,7 @@ const overrideTimeframe: Ref<Timeframe> = computed(() => {
     })
   }
 
-  const relativePeriod = TimePeriods.get(timeSpec.time_range)
+  const relativePeriod = cloneDeep(TimePeriods.get(timeSpec.time_range))
```
adorack, line 47: Same as the metric cards, I'd prefer to avoid this unless we actually find out that we need it.   The dashboard renderer doesn't expose this value to the host app and doesn't mutate it.

adorack (changes requested):

adorack (approved):

PR #1917, opened 2025-01-27
fix(plugins): mocking spec max length

kaiarrowood:
# Summary

Add ability to customize the max character length for a textarea. Takes `boolean | number`, if provided `false` no max character limit is set.
Set no max character limit in client side for mocking plugin's spec length.

For https://konghq.atlassian.net/browse/KM-922.


Leopoldthecoder (approved):
LGTM, thank you!

PR #1918, opened 2025-01-28
fix: reduce font size and padding in tooltip

filipgutica:
# Summary

image


helen-tang27:
I'm good with the changes. Will revisit and fine-tune once have more time later.

adorack (approved):

PR #1921, opened 2025-01-30
feat(consumers): new empty state [khcp-14823]

kaiarrowood:
# Summary

Shared changes to `EntityBaseTable` for new empty states + usage in consumers.
Part of [KHCP-14823](https://konghq.atlassian.net/browse/KHCP-14823).

[KHCP-14823]: https://konghq.atlassian.net/browse/KHCP-14823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-consumers/src/components/ConsumerList.vue
+++ b/packages/entities/entities-consumers/src/components/ConsumerList.vue
@@ -665,6 +665,12 @@ onBeforeMount(async () => {
 </script>
 
 <style lang="scss" scoped>
+.button-row {
+  align-items: center;
+  display: flex;
+  gap: $kui-space-50;
+}
+
 .kong-ui-entities-consumers-list {
   width: 100%;
 
--- a/packages/entities/entities-consumers/src/components/ConsumerList.vue
+++ b/packages/entities/entities-consumers/src/components/ConsumerList.vue
@@ -18,6 +18,7 @@
       @click:row="(row: any) => rowClick(row as EntityRow)"
       @empty-state-cta-clicked="handleEmptyStateCtaClicked"
       @sort="resetPagination"
+      @state="handleStateChange"
     >
       <!-- Filter -->
       <template #toolbar-filter>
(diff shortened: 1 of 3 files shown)


packages/entities/entities-consumers/src/components/ConsumerList.vue
```diff
@@ -49,6 +49,31 @@
         </Teleport>
       </template>
 
+      <template
+        v-if="enableV2EmptyStates && config.app === 'konnect'"
```
afrzhussain, line 53: suggestion: can we move 'konnect' to a constant

kaiarrowood: Looking at it, that kind of refactor would be too large and I can't see an easy way forward without making the change for all entities in a single PR. Can you create a separate ticket for this work?

packages/entities/entities-consumers/src/components/ConsumerList.vue
```diff
@@ -33,22 +34,59 @@
...
             </KButton>
-          </PermissionsWrapper>
+            <PermissionsWrapper :auth-function="() => canCreate()">
+              <!-- Hide Create button if table is empty -->
+              <KButton
+                appearance="primary"
+                data-testid="toolbar-add-consumer"
+                :size="useActionOutside ? 'medium' : 'large'"
+                :to="config.consumerGroupId ? undefined : config.createRoute"
+                @click="() => config.consumerGroupId ? handleAddConsumerClick() : undefined"
```
adamdehaven, line 55: I'm a little confused here by both the `to` and `@click` handlers. If the user clicks the link, they could both fire since the `@click` doesn't have `prevent` etc.

Is there any context as to how this will work?

kaiarrowood: Should I have `@click.prevent` too?

adamdehaven: Could you just move all of the logic into a click.prevent handler instead of defining both?

kaiarrowood: Looks like this was old code, I had already added a function to do this, so updated to use it here.

mptap (approved):
Lgtm

PR #1937, opened 2025-02-07
fix: tile header overflow issue

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-3645

Also add a tooltip when title is truncated

image
image

### DIFF
--- a/packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
+++ b/packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
@@ -7,12 +7,19 @@
       v-if="hasKebabMenuAccess && definition.chart.type !== 'slottable'"
       class="tile-header"
     >
-      <div
-        class="title"
-        :data-testid="`tile-title-${tileId}`"
+      <KTooltip
+        class="title-tooltip"
+        :disabled="!isTitleTruncated"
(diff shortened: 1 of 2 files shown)


adorack (changes requested):
I'd like to review.

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -135,6 +142,16 @@ const hasKebabMenuAccess = evaluateFeatureFlag('ma-3043-analytics-chart-kebab-me
 
 const chartData = ref<ExploreResultV4>()
 const exportModalVisible = ref<boolean>(false)
+const titleRef = ref<HTMLElement>()
+const isTitleTruncated = ref(false)
+
+onMounted(async () => {
```
adorack, line 148: This needs to be a `watch`; it's possible for the title to change without the tile getting unmounted.

adorack (changes requested):

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
filipgutica: good call!

adorack (approved):

PR #1939, opened 2025-02-10
fix(entities-plugins): fields set to null

Leopoldthecoder:
# Summary

Fixing a bug where `vectordb` or `embeddings` will be set to `null` when editing an `ai-proxy-advanced` plugin.

KM-945


sumimakito (approved):
Manually tested and LGTM

PR #1947, opened 2025-02-11
fix: ff guard v2 empty state changes

afrzhussain:
# Summary
Fix FF guarding LH buttons


vaibhavrajsingh2001 (approved):
Code LGTM. Can you get a functional review from the person who reported this by sharing the PR preview URL?

PR #1954, opened 2025-02-14
fix(analytics-chart): render timeseries charts with a single point [MA-2682]

ryanmoore:
# Summary

Second attempt at: https://github.com/Kong/public-ui-components/pull/1872

When a line chart is asked to display a single data point, display that datapoint with a radius smaller than the hover radius instead of displaying a deceptive "No data" error.

Test dataset:

```
{
  "data": [
    {
      "event": {
        "request_count": 100000
      },
      "timestamp": "2024-12-23T07:00:00.000Z"
    }
  ],
  "meta": {
    "start_ms": 1734937200000,
    "end_ms": 1735023600000,
    "limit": 50,
    "granularity_ms": 86400000,
    "metric_names": [
      "request_count"
    ],
    "truncated": false,
    "display": {},
    "metric_units": {
      "request_count": "count"
    }
  }
}
```

Before:
image

After:
image

### DIFF
--- a/packages/analytics/analytics-chart/src/composables/useLineChartOptions.ts
+++ b/packages/analytics/analytics-chart/src/composables/useLineChartOptions.ts
@@ -124,7 +124,7 @@ export default function useLinechartOptions(chartOptions: LineChartOptions) {
       },
       elements: {
         point: {
-          radius: 0,
+          radius: chartOptions.pointsWithoutHover?.value ? 3 : 0,
           hitRadius: 4,
           hoverRadius: 4,
         },
--- a/packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
+++ b/packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
@@ -75,6 +75,7 @@ import type { GranularityValues, AbsoluteTimeRangeV4 } from '@kong-ui-public/ana
 import type { Chart } from 'chart.js'
 import { ChartLegendPosition } from '../../enums'
 import { formatByGranularity, generateLegendItems } from '../../utils'
+import { hasExactlyOneDatapoint } from '../../utils/commonOptions'
 
 const props = defineProps({
   chartData: {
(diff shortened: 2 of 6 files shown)


packages/analytics/analytics-chart/README.md
```diff
@@ -510,3 +510,11 @@ const fileName = ref('exportFilename')
 </script>
 
 ```
+
+## Development
+
+Follow the setup instructions for the [top level README](../../../README.md). Then run the sandbox with:
+
+```
+pnpm run dev
```
filipgutica, line 519: I think this will only work if you're in the `analytics-chart` package directory. 

If you're at the root the right command would be 
`pnpm --filter @kong-ui-public/analytics-chart run dev`

filipgutica: The main readme mentions it
image

packages/analytics/analytics-chart/src/composables/useLineChartOptions.ts
```diff
@@ -124,7 +124,7 @@ export default function useLinechartOptions(chartOptions: LineChartOptions) {
       },
       elements: {
         point: {
-          radius: 0,
+          radius: chartOptions.pointsWithoutHover?.value ? 2 : 0,
```
filipgutica, line 127: I wonder if the radius should be a bit larger for a single point. Testing it out, I find the "single point" hard to see in the chart

filipgutica (changes requested):
Just a couple comments

packages/analytics/analytics-chart/README.md
ryanmoore: deleted my readme changes

packages/analytics/analytics-chart/src/composables/useLineChartOptions.ts
ryanmoore: per our discussion, how's 3?

our reasoning was: 2 looked small. 4 is the on-hover size and we'd still like an effect, so that leaves 3.

filipgutica (approved):

PR #1959, opened 2025-02-17
fix(entities-keys): text area char limit

Leopoldthecoder:
# Summary

This PR removes the character limit on all `<KTextArea>` instances on the key form page.

KM-907


nekolab (approved):
LGTM

PR #1970, opened 2025-02-19
feat(dashboard-renderer): limit request concurrency [MA-3623]

adorack:
- Use `p-queue` to define windows for issuing queries
- Override `queryFn` for all components inside a dashboard to apply a queue
- Define default values for queue parameters, but allow overrides through config mechanism
- Remove `lazyLoad` until we can figure out how to get it working

# Summary


packages/analytics/dashboard-renderer/src/composables/useRequestQueue.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  if (!queryBridge) {
+    // If there's no query bridge injected, we can't make a request queue.
+    return
+  }
+
+  if (queryBridge.queryFnOverride) {
+    // Someone goofed; this composable isn't intended to be nested.
+    // If someone has a use case for this, we should revisit the design.
+    console.warn('Detected incorrect use of useRequestQueue.  This composable should not be used in nested components.  Please discuss with team-data or report an issue if you feel you have a use case.')
```
filipgutica, line 23: What does nested component mean here, is it nested relative to <DashboardRenderer> ?

adorack: Yes, exactly -- it shouldn't be possible to nest it as it's designed now; I just put a warning here in case this composable ends up getting used in places that I don't expect.

filipgutica (approved):

PR #1972, opened 2025-02-20
fix(dashboard-renderer): enable doughnut charts [MA-3650]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/MA-3650

Enabling doughnut charts in DashboardRenderer

screencapture-localhost-5173-2025-02-21-10_04_31

### DIFF
--- a/packages/analytics/dashboard-renderer/src/components/DoughnutChartRenderer.vue
+++ b/packages/analytics/dashboard-renderer/src/components/DoughnutChartRenderer.vue
@@ -2,7 +2,6 @@
   <BaseAnalyticsChartRenderer
     :chart-options="chartOptions"
     :context="context"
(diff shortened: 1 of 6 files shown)


packages/analytics/dashboard-renderer/src/components/DoughnutChartRenderer.vue
```diff
@@ -0,0 +1,20 @@
+<!-- DonutChartRenderer.vue -->
```
adorack, line 1: Minor: could you remove the comment?

portikM: Done

adorack (approved):

PR #1974, opened 2025-02-21
feat: publish previews to internal repository [KHCP-14702]

ValeryG:
# Summary

publishing previews to internal npm registry


package.json
```diff
@@ -102,7 +102,7 @@
   },
   "volta": {
     "node": "18.20.4",
-    "pnpm": "9.12.2"
+    "pnpm": "10.4.0"
```
adamdehaven, line 105: Before merging, we should run `pnpm approve-builds` so that the root `package.json` is updated to allow lefthook, etc. to run

adamdehaven (approved):

PR #1977, opened 2025-02-24
fix(forms): array elements are deleted

Leopoldthecoder:
# Summary

Array elements are mistakenly deleted.

FTI-6518


sumimakito (approved):
Manually tested and fixed as expected

PR #1985, opened 2025-02-25
fix: keep tile sorted by row/column order

filipgutica:
# Summary

Note: The indexes appearing on the tiles are just to demonstrate the tiles array now sorted. Change is not included in the PR.
https://www.loom.com/share/e88029be90854f32948776d26eb93f76

### DIFF
--- a/packages/analytics/dashboard-renderer/src/components/layout/DraggableGridLayout.vue
+++ b/packages/analytics/dashboard-renderer/src/components/layout/DraggableGridLayout.vue
@@ -49,6 +49,7 @@ const emit = defineEmits<{
 
 const gridContainer = ref<HTMLDivElement | null>(null)
 
+
 const tilesRef = ref<Map<string, GridTile<any>>>(new Map(props.tiles.map(t => [`${t.id}`, t])))
 
 let grid: GridStack | null = null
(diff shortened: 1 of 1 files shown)


packages/analytics/dashboard-renderer/src/components/layout/DraggableGridLayout.vue
```diff
@@ -49,7 +49,7 @@ const emit = defineEmits<{
 
 const gridContainer = ref<HTMLDivElement | null>(null)
 
-const tilesRef = ref<Map<string, GridTile<any>>>(new Map(props.tiles.map(t => [`${t.id}`, t])))
+const tilesRef = computed<Map<string, GridTile<any>>>(() =>new Map(props.tiles.map(t => [`${t.id}`, t])))
```
adorack, line 52: I'm not sure that this should be a `computed` given that you're calling `set` and `delete` outside of the ref.

adorack (changes requested):

packages/analytics/dashboard-renderer/src/components/layout/DraggableGridLayout.vue
filipgutica: done

PR #1987, opened 2025-02-26
fix(tracing): hide upstream nodes when not available; update text

sumimakito:
# Summary

* Hide upstream nodes when they are not available (e.g., no route matched with the request)
* Update UI text

| |Screenshot|
|-:|:-|
|**Before**|Screenshot 2025-02-26 at 16 45 01|
|**After**|Screenshot 2025-02-26 at 16 42 50|
|**After**|Screenshot 2025-02-26 at 16 43 29|

KM-982


Leopoldthecoder (approved):
Manually tested and LGTM

PR #1990, opened 2025-02-27
refactor: refactor part of the hard code into constants [KM-986]

raoxiaoyan:
# Summary

KM-986


packages/entities/entities-upstreams-targets/src/components/TargetForm.vue
```diff
@@ -259,9 +266,7 @@ const saveFormData = async (): Promise<void> => {
     if (formType.value === 'create') {
       response = await axiosInstance.post(submitUrl.value, requestBody.value)
     } else if (formType.value === 'edit') {
-      response = props.config?.app === 'konnect'
-        // Note: Konnect currently uses PUT because PATCH is not fully supported in Koko
```
hangrao, line 263: This Note looks still valid, should we keep it?

PR #2009, opened 2025-03-07
Revert "fix(forms): use Number.isFinite (#1996)"

TT1228:
This reverts commit 92a7ec26bbc5898e93191f0b214b9bbdd2e4a55d.

KAG-6576

We need to pass `NaN` to form component for it to function as expected, save `null` value when number input is cleared.


Leopoldthecoder (approved):
Thank you!

PR #2013, opened 2025-03-12
feat(redis): add tags field

2eha0:
# Summary
Add `tags` to redis partial
[KM-1036](https://konghq.atlassian.net/browse/KM-1036)

[KM-1036]: https://konghq.atlassian.net/browse/KM-1036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-redis-configurations/src/helpers.ts
+++ b/packages/entities/entities-redis-configurations/src/helpers.ts
@@ -83,4 +83,14 @@ export const standardize = {
     }
     return newObj
   },
+
+  stringToArray(value: string | null | undefined, separator = ','): string[] {
+    if (!value) return []
+    return value.split(separator).map((tag: string) => String(tag || '').trim()).filter((tag: string) => tag !== '')
+  },
+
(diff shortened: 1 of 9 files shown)


packages/entities/entities-redis-configurations/src/helpers.ts
```diff
@@ -83,4 +83,14 @@ export const standardize = {
     }
     return newObj
   },
+
+  stringToArray(value: string | null | undefined, separator = ','): string[] {
```
TT1228, line 87: I found that the standard operation of tags is
`fields.tags.split(',')?.map((tag: string) => String(tag || '')
      .trim())?.filter((tag: string) => tag !== '')` which filter empty string and can take inputs like `a,b,,`

TT1228 (approved):
Manually tested and LGTM

Leopoldthecoder (approved):

PR #2017, opened 2025-03-12
fix(tracing): hide nodes for missing spans

sumimakito:
# Summary

This pull request hides the nodes in the trace lifecycle view lacking the corresponding spans. Sandbox is updated to provide more controls to test out different combinations.

KM-1023


packages/core/tracing/sandbox/pages/TraceViewPage.vue
```diff
@@ -101,23 +124,77 @@
...
+
+const spanTrees = computed(() => {
+  const trees = buildSpanTrees(mergeSpansInTraceBatches(traceBatches))
+  trimTree(trees.roots[0])
+  return trees
+})
+
+watch(spanTrees, () => {
+  renderKey.value += 1
+}, { deep:true })
```
jw-foss, line 191: **Question:** I see this is updated whenever the `spanTrees` gets changed, but this key is applied to `KSlideout` instead of its children, is this expected to re-render the entire `slideout` instead of its children?

jw-foss (approved):
@sumimakito I have one question and left as comment, the rest of the code LGTM, feel free to ignore it :bow:

jw-foss (approved):
LGTM, thank you!

Leopoldthecoder (approved):
LGTM

packages/core/tracing/sandbox/pages/TraceViewPage.vue
sumimakito: Nice catch. This is intended as of today since we did not find a good way to reload and re-layout the lifecycle view after the span tree changes. It is only affecting the sandbox but not the production as we won't update the tree there. We will revisit and improve in the future.

sumimakito: @jw-foss I opened another PR https://github.com/Kong/public-ui-components/pull/2021 to improve this. Could you take a look at it?

PR #2021, opened 2025-03-13
fix(tracing): re-layout and re-render on tree change

sumimakito:
# Summary

Adds (relative) unique prefixes to nodes to ensure re-layout and re-render on tree change

### DIFF
--- a/packages/core/tracing/src/utils/lifecycle.ts
+++ b/packages/core/tracing/src/utils/lifecycle.ts
@@ -70,7 +70,11 @@ const buildPluginNodeData = (node: SpanNode): [LifecycleNodeData & { id: string 
   }, durationExcluded]
 }
 
-export const buildLifecycleGraph = (root: SpanNode): LifecycleGraph => {
+export interface BuildLifecycleGraphOptions {
+  idPrefix?: string
+}
+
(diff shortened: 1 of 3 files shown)


packages/core/tracing/src/utils/lifecycle.ts
```diff
@@ -145,7 +149,7 @@ export const buildLifecycleGraph = (root: SpanNode): LifecycleGraph => {
     nodeTree: {
       client: {
         node: {
-          id: LifecycleNodeType.CLIENT,
+          id: `${options?.idPrefix ?? ''}${LifecycleNodeType.CLIENT}`,
```
jw-foss, line 152: @sumimakito WDYT about evaluate the expression only once on the top of the `buildLifecycleGraph` method 🤔 

```ts
export const buildLifecycleGraph = (root: SpanNode, options?: BuildLifecycleGraphOptions): LifecycleGraph => {
  const prefix = options?.idPrefix ?? '';
  // while using
{ 
  // ...other attributes
  id: `${prefix}${LifecycleNodeType.CLIENT_OUT}`,
}
```

sumimakito: Good idea and I've just pushed another commit

Leopoldthecoder (approved):
LGTM

jw-foss (approved):
LGTM, thank you! 👍 Good work!

PR #2022, opened 2025-03-13
feat: disable jump to explore for api queries

alxgrant:
# Summary

### DIFF
--- a/packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
+++ b/packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
@@ -165,10 +165,11 @@ watch(() => props.definition, async () => {
 const exploreLink = computed(() => {
   const filters = [...props.context.filters, ...props.definition.query.filters ?? []]
   const dimensions = props.definition.query.dimensions as QueryableExploreDimensions[] | QueryableAiExploreDimensions[] ?? []
-  // TODO: remove once portal has been added as option in Explore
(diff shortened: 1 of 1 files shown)


packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -165,10 +165,11 @@ watch(() => props.definition, async () => {
 const exploreLink = computed(() => {
   const filters = [...props.context.filters, ...props.definition.query.filters ?? []]
   const dimensions = props.definition.query.dimensions as QueryableExploreDimensions[] | QueryableAiExploreDimensions[] ?? []
-  // TODO: remove once portal has been added as option in Explore
-  if (filters.some(filter => ('dimension' in filter && filter.dimension === 'portal') ||
-    ('field' in filter && filter.field === 'portal')) ||
-    dimensions.some(dim => dim === 'portal')
+  // TODO: remove once portal and api are available in Explore
+  if (filters.some(filter =>
+    ('dimension' in filter && (filter.dimension === 'portal' || filter.dimension === 'api')) ||
```
filipgutica, line 170: We should probably do something like this now that we're adding more dimensions make it easier to maintain what dimensions are being excluded.

```
const excludeList = new Set(['portal', 'api'])

if ('dimension' in filter && excludeList.has(filter.dimension) || 'field' in filter && excludeList.has(filter.field) || 
    dimensions.some(dim => excludeList.has(dim)) {
    return ''
}
```

PR #2037, opened 2025-03-17
feat(entities-routes): split route-rule-form from route-form

sumimakito:
# Summary

This pull request splits the components to configure the rules in the route form into a separate component `<RouteFormRulesComposer />` to allow them to be reused on a finer-grained level.

A sandbox for this component has also been added:

Screenshot 2025-03-26 at 17 17 25

KM-1062


packages/entities/entities-routes/src/components/RouteForm.cy.ts
```diff
@@ -253,6 +253,7 @@ describe('<RouteForm />', { viewportHeight: 700, viewportWidth: 700 }, () => {
           cy.get('.expression-editor .monaco-editor').should('be.visible')
 
           // base advanced fields
+          cy.getTestId('collapse-trigger-content').click()
```
Leopoldthecoder, line 256: Question: why do we need to click this element?

packages/entities/entities-routes/src/components/RouteFormExpressionsEditorLoader.vue
```diff
@@ -33,20 +33,20 @@
...
-  render: () => h('div', t('form.expressions_editor.loading')),
+  render: () => h('div', i18n.t('form.expressions_editor.loading')),
 }
 
 const errorComponent: Component = {
-  render: () => h('div', t('form.expressions_editor.error')),
+  render: () => h('div', i18n.t('form.expressions_editor.error')),
 }
 
 const RouteFormExpressionsEditor = defineAsyncComponent({
```
Leopoldthecoder, line 52: Are changes in this file necessary?

packages/entities/entities-routes/src/components/RouteForm.cy.ts
sumimakito: This is because previously, the "Traditional" and "Expressions" tabs shared the same container for "Advanced fields", but this was changed in this PR, where each tab will have its own "Advanced fields" container and the collapsed state. Therefore, we will have to click the "Advanced fields" again to show these fields after switching to another tab.

Leopoldthecoder: I see. Thanks for the explanation!

adamdehaven (approved):

PR #2046, opened 2025-03-20
feat(redis): set default redis type to enterprise

2eha0:
[KM-1071](https://konghq.atlassian.net/browse/KM-1071)

[KM-1071]: https://konghq.atlassian.net/browse/KM-1071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-redis-configurations/src/components/RedisConfigurationForm.vue
+++ b/packages/entities/entities-redis-configurations/src/components/RedisConfigurationForm.vue
@@ -439,6 +439,8 @@ import ClusterNodes from './ClusterNodes.vue'
 import composables from '../composables'
 import SentinelNodes from './SentinelNodes.vue'
 import { useLinkedPluginsFetcher } from '../composables/useLinkedPlugins'
+import { DEFAULT_REDIS_TYPE } from '../constants'
+import { mapRedisTypeToPartialType } from '../helpers'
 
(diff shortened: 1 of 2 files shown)


packages/entities/entities-redis-configurations/src/components/RedisConfigurationForm.vue
```diff
@@ -556,6 +558,25 @@ const getSelectedText = (item: any) => {
...
+  if (defaultPartialType === props.disabledPartialType) {
+    return props.disabledPartialType === PartialType.REDIS_CE
+      ? RedisType.HOST_PORT_EE
+      : RedisType.HOST_PORT_CE
+  }
+
+  // Default type is valid
+  return DEFAULT_REDIS_TYPE
+}
+
```
jw-foss, line 579: **Question:** I saw the usage of `props` would it be better if we convert this to a `computed`? Do we need it be reactive?

2eha0: No, the default type shouldn't be reactive

raoxiaoyan (approved):

PR #2050, opened 2025-03-21
feat(analytics): make explore URL determination async [MA-3590]

adorack:
- Allow permissions checks to happen before generating an explore URL.
- Clean up dead code.

BREAKING CHANGE: change signature of analytics bridge

# Summary

### DIFF
--- a/packages/analytics/analytics-utilities/src/types/query-bridge.ts
+++ b/packages/analytics/analytics-utilities/src/types/query-bridge.ts
@@ -27,11 +27,12 @@ export interface AnalyticsBridge {
   configFn: () => Promise<AnalyticsConfigV2>,
 
   // Evaluate feature flags (if applicable)
-  // TODO: Use `NoInfer` once we upgrade to TS 5.4:
-  //  https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html#the-noinfer-utility-type
-  //  See note in DashboardRenderer tests.
   evaluateFeatureFlagFn: <T = boolean>(key: string, defaultValue: T) => T,
-  exploreBaseUrl?: () => string,
 
+  // Define the location of explore to enable jump-to-explore.
+  // Async because there might need to be permissions checks.
+  exploreBaseUrl?: () => Promise<string>,
+
+  // Dynamically provide certain components that aren't available in all environments
   fetchComponent?: (name: string) => Promise<Component>,
 }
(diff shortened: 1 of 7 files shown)


packages/analytics/dashboard-renderer/src/components/BaseAnalyticsChartRenderer.vue
```diff
@@ -61,22 +58,6 @@ const options = computed((): AnalyticsChartOptions => ({
   threshold: props.chartOptions.threshold,
 }))
 
-const exploreLink = computed(() => {
```
filipgutica, line 64: Why is this not needed anymore?

packages/analytics/analytics-utilities/src/types/query-bridge.ts
```diff
@@ -27,11 +27,12 @@ export interface AnalyticsBridge {
...
 
   // Evaluate feature flags (if applicable)
-  // TODO: Use `NoInfer` once we upgrade to TS 5.4:
-  //  https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html#the-noinfer-utility-type
-  //  See note in DashboardRenderer tests.
   evaluateFeatureFlagFn: <T = boolean>(key: string, defaultValue: T) => T,
-  exploreBaseUrl?: () => string,
 
+  // Define the location of explore to enable jump-to-explore.
+  // Async because there might need to be permissions checks.
```
filipgutica, line 33: Should you also add `canUserAccess: boolean` here?

filipgutica (changes requested):
Just one comment

packages/analytics/analytics-utilities/src/types/query-bridge.ts
adorack: I thought about that for a while, but currently think it's best to leave the permissions checks out of the component:

- There's some chance the signature could be as simple as `canUserAccess: boolean`, but this basically forces the permissions check to happen at the time the app is first initialized.  I don't think it's a good idea to rely on this.
- Say the signature were `canUserAccess: (krn) => Promise<boolean>`.  This would maximize flexibility, but requires that the components know which specific permission they need to check.  At least as of right now, this feels too Konnect-specific to put in the component.

It's something that we might change later?  But for now, making `exploreBaseUrl` async means that the check can happen in the bridge (like so: https://github.com/Kong/shared-ui-components/pull/3157/files).

filipgutica (approved):

PR #2053, opened 2025-03-25
fix(tracing): incorrect type used in the legacy span sampler attribute

sumimakito:
# Summary

This pull request fixes [an incorrectly typed span sampler attribute](https://github.com/Kong/public-ui-components/commit/853f09c6c5564b58e5f1d2810d40fbc6bfa6fded#diff-4f0e8fcac8c74a38bf33506a167e4c4dd4423d1c1cc1a02f912e9cdcb5319b89R117) that has caused the downstream editor to fail to initialize.

This pull request also enforces the value types for ATC-related fields.

A verification PR has been created.

KM-1080


Leopoldthecoder (approved):
Manually tested in the preview link and LGTM.

PR #2056, opened 2025-03-25
Fix/km 1029

TT1228:
# Summary

KM-1029

### DIFF
--- a/packages/entities/entities-shared/src/composables/useHelpers.ts
+++ b/packages/entities/entities-shared/src/composables/useHelpers.ts
@@ -1,3 +1,7 @@
+import isEqual from 'lodash.isequal'
+import isPlainObject from 'lodash.isplainobject'
+import sortBy from 'lodash.sortby'
+
 export default function useHelpers() {
   /**
    * propName could be 'rowValue' or 'row'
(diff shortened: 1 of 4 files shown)


packages/entities/entities-shared/src/composables/useHelpers.ts
```diff
@@ -10,15 +14,24 @@ export default function useHelpers() {
...
+    } else if (isPlainObject(obj)) {
+      const sortedObj: Record<string, any> = {}
+      Object.keys(obj)
+        .sort()
+        .forEach((key) => {
+          sortedObj[key] = deepSort(obj[key])
+        })
+      return sortedObj
+    } else {
+      return obj
```
jw-foss, line 29: I'm suspicious about this sort operation, in my understanding performing deep equal might not need a sort operation prior to comparison 🤔 Can we try a non-intrusive way for performing the comparison here? For instance, we compare the key first, if we found unmatched key, then the object is not equal, then compare the value, recursively perform the operation.

Justineo: +1, we can compare the keys first and exit early on any mismatch.

PR #2058, opened 2025-03-25
fix(entities-shared): expose entity base form actions [KHCP-15505]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-15505

Add `hideActions` props and expose EntityBaseForm action methods to allow triggering them from the host app


mptap (approved):
LGTM 🚀

PR #2059, opened 2025-03-25
fix: back to pnpm for package publishing [KHCP-15589]

ValeryG:
# Summary

lerna publish by design is not updating versions of peerDependencies. As a result published package still has the word `workspace`  as a version number. 

This is breaking npm installs when npm is trying to install peerDependencies automatically. 

Solution: Need to go back to pnpm but still have KHCP-15491 fixed (not publish packages that do not need to be published) 

🧵 on issue: https://kongstrong.slack.com/archives/C04A6LD904E/p1742840099558359 

So essentially this is the rollback of [PR](https://github.com/Kong/public-ui-components/pull/2029) but make pnpm publish command use correct filter to not re-introduce KHCP-15491 issue


adamdehaven (approved):
lgtm

PR #2060, opened 2025-03-26
feat(vfg): introduce custom field component

nekolab:
In this PR, we expose a `component` field in the `FormFieldSchema`, which can point to a Vue component and override the default field render.

This gives us more flexibility when handling complex requirements

Test result:
* Konnect adoption test: https://github.com/Kong/konnect-ui-apps/pull/6264
* Kong admin adoption test: https://github.com/Kong/kong-admin/pull/3837
* Kong admin cross browser test: https://github.com/Kong/kong-admin/actions/runs/14079674561

JIRA: KM-1035

### DIFF
--- a/packages/core/forms/src/components/FormGroup.vue
+++ b/packages/core/forms/src/components/FormGroup.vue
@@ -39,11 +39,11 @@
     <div class="field-wrap">
       <component
         v-bind="$attrs"
-        :is="getFieldType(field)"
+        :is="getFieldComponent(field)"
         ref="child"
         :disabled="fieldDisabled(field) || null"
         :form-options="options"
-        :hint="field.hint && getFieldType(field) === 'field-input' ? fieldHint(field) : undefined"
+        :hint="field.hint && field.type === 'input' ? fieldHint(field) : undefined"
         :model="model"
         :schema="schema"
         :vfg="vfg"
@@ -67,7 +67,7 @@
     </div>
 
     <div
-      v-if="field.hint && getFieldType(field) !== 'field-input'"
+      v-if="field.hint && field.type !== 'input'"
       class="hint"
       v-html="fieldHint(field)"
     />
@@ -185,8 +185,13 @@ export default {
     getTooltipId(schema) {
       return `${this.getFieldID(schema)}-tooltip`
     },
-    // Get type of field 'field-xxx'. It'll be the name of HTML element
-    getFieldType(fieldSchema) {
+    // Get type of field, it could be the name of a registered component or a Vue component
+    // a registered component should be named in 'field-{type}' format
+    getFieldComponent(fieldSchema) {
+      if (fieldSchema.component) {
+        return fieldSchema.component
+      }
+
       return 'field-' + fieldSchema.type
     },
     // Get type of button, default to 'button'
--- a/packages/core/forms/src/composables/useAbstractFields.ts
+++ b/packages/core/forms/src/composables/useAbstractFields.ts
@@ -26,9 +26,10 @@ interface AbstractFieldParams {
     errors: any[]
     field: Record<string, any>
   }) => void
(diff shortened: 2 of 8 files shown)


packages/core/forms/src/components/FormGroup.vue
```diff
@@ -186,7 +186,11 @@ export default {
       return `${this.getFieldID(schema)}-tooltip`
     },
     // Get type of field 'field-xxx'. It'll be the name of HTML element
```
jw-foss, line 188: **Question:** maybe we should update the comment as well.

Also, I was wondering if we can update the code to, this way we can keep the old method intact, just a random thought.

```ts
getFieldComponent(fieldSchema) {
       return field.component || this.getFieldType(fieldSchema)
    },
```

packages/core/forms/src/types/form-fields.ts
```diff
@@ -16,6 +18,7 @@ export interface FormFieldSchema extends Record<string, any> {
   hint?: string
   submitWhenNull?: boolean
   selectOptions?: Record<string, any>
+  component?: Component | null
```
jw-foss, line 21: Wondering under what situation the schema would pass a `null` value to this field.

packages/core/forms/src/composables/useAbstractFields.ts
```diff
@@ -86,19 +87,25 @@ export default function useAbstractFields(formData: AbstractFieldParams) {
...
+        if (!Array.isArray(schemaValidator)) {
+          validators.push(convertValidator(schemaValidator))
+        } else {
+          forEach(schemaValidator, validator => {
+            validators.push(convertValidator(validator))
+          })
+        }
+      }
+
+      if (formData.externalValidator && typeof formData.externalValidator === 'function') {
```
jw-foss, line 107: I saw the first line we imported `isFunction` here can we utilize that here instead of using `typeof`?

packages/core/forms/src/composables/useAbstractFields.ts
```diff
@@ -26,9 +26,10 @@ interface AbstractFieldParams {
     errors: any[]
     field: Record<string, any>
   }) => void
+  externalValidator?: (value: any, field: any, model: any) => string[] | Promise<string[]>
 }
 
-export default function useAbstractFields(formData: AbstractFieldParams) {
+export default function useAbstractFields<T = any>(formData: AbstractFieldParams) {
```
jw-foss, line 32: Would `unknown` be a better option here?

jw-foss (commented):
@nekolab fantastic job on the sandboxed demo! Got a few questions for your considerations 👍

packages/core/forms/src/types/form-fields.ts
nekolab: Emm, like what we demonstrated in the `CatFoodField`. If we'd like to use a conditional check to flip the component, it would be better to use `null` instead of an implicit `undefined`

jw-foss: Understood, resolved

packages/core/forms/src/composables/useAbstractFields.ts
nekolab: I don't think so. Using `any` here gives us maximum compatibility with the existing code, however using `unknown` here will break everything

jw-foss: Noted

packages/core/forms/src/components/FormGroup.vue
nekolab: Comment updated. However, I don't think we need to keep the original method intact. The method `getFieldType` is only used for indicating the Vue dynamic components, keeping it (in `field-xxx` form) might cause more confusion in the furtue

jw-foss (approved):

Justineo (approved):
LGTM

sumimakito (approved):

PR #2064, opened 2025-03-27
feat(analytics): add single value simple chart [MA-3730]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/MA-3730

Adds `SingleValue` simple chart, allows it to be used in dashboard renderer

Demo: https://www.loom.com/share/2e13cf753a074640a45a3ea061bc8eb1?sid=885a0cb2-9322-4206-aed8-0285f114d314


packages/analytics/analytics-utilities/src/dashboardSchema.ts
```diff
@@ -208,6 +209,25 @@ export const metricCardSchema = {
...
+  type: 'object',
+  properties: {
+    type: {
+      type: 'string',
+      enum: ['single_value'],
+    },
+    decimalPoints: {
+      type: 'number',
+    },
+    syntheticsDataKey,
```
adorack, line 222: It doesn't look like this chart accepts a synthetics data key; I think you can remove this line until we decide that it actually needs one.

packages/entities/entities-shared/src/components/entity-link/EntityLink.vue
adorack, line 1: Why does this file have changes?

packages/analytics/analytics-utilities/src/dashboardSchema.ts
portikM: Done

packages/entities/entities-shared/src/components/entity-link/EntityLink.vue
portikM: `pnpm stylelint:fix` command

packages/analytics/analytics-chart/src/components/chart-types/SingleValue.vue
```diff
@@ -0,0 +1,132 @@
...
+  /**
+   * Number of decimal points to display
+   */
+  decimalPoints: {
+    type: Number,
+    default: SINGLE_VALUE_DEFAULT_DECIMAL_POINTS,
+  },
+})
+
+const record = computed((): AnalyticsExploreRecord => props.data.data[0])
```
portikM, line 46: @adorack @filipgutica is my logic correct here in the sense that we only need the first record here? Explore queries without group by only return 1 event but asking in case I'm missing something here

adorack: I think this is a fine assumption.

filipgutica: May be it would be useful to throw a console warning if sigle value is being used with an explore result that has more than one record?

filipgutica (commented):
Just a comment

packages/analytics/analytics-chart/src/components/chart-types/SingleValue.vue
portikM: Good call, added a `console.warn` in `onMounted`

filipgutica (approved):

adorack (approved):

PR #2068, opened 2025-03-31
fix(dashboards): update grid size when tiles are updated [MA-3748]

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-3748

BREAKING CHANGE: Dashboard config is now passed via v-model instead of the config prop
- Replaces one-way config prop with two-way v-model binding
- Allows the dashboard renderer to modify the config directly
- Changes made within the component are synced with the parent’s ref
- remove `@update-tiles` event

This enables internal changes to the config (e.g., grid size changes, layout edits, widget updates) to automatically sync back to the parent, providing full two-way reactivity.


packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -215,6 +216,11 @@ const handleUpdateTiles = (tiles: GridTile<TileDefinition>[]) => {
       definition: tile.meta,
     } as TileConfig
   })
+  const newGridSize = {
+    cols: Math.max(...updatedTiles.map(t => t.layout.position.col + t.layout.size.cols)),
+    rows: Math.max(...updatedTiles.map(t => t.layout.position.row + t.layout.size.rows)),
+  }
+  model.value.gridSize = newGridSize
```
filipgutica, line 223: config is now bound using v-model, enabling two-way data binding. Changes made internally will automatically sync with the parent component.

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -215,7 +216,13 @@ const handleUpdateTiles = (tiles: GridTile<TileDefinition>[]) => {
       definition: tile.meta,
     } as TileConfig
   })
-  emit('update-tiles', updatedTiles.sort(tileSortFn))
+  const newGridSize = {
+    cols: Math.max(...[2, ...updatedTiles.map(t => t.layout.position.col + t.layout.size.cols)]),
```
filipgutica, line 220: dont let cols be less than 2

adorack (changes requested):
Can you please also update the tests to check that the 2-way binding works, rather than testing that `@update-tiles` works?

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.cy.ts
adorack, line 1: Ah, sorry I missed this -- can you please add a check for the new grid size adjustment as well?

adorack (changes requested):

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.cy.ts
```diff
@@ -908,11 +906,67 @@ describe('<DashboardRenderer />', () => {
...
-      .its('firstCall.args.0')
-      .then(tiles => {
-        expect(tiles.map((tile: TileConfig) => tile.id)).to.deep.equal(updatedTileIDOrder)
+    cy.wrap(configRef).should((ref: Ref<DashboardConfig>) => {
+      const currentOrder = ref.value.tiles.map((tile: TileConfig) => tile.id)
+      expect(currentOrder).to.deep.equal(updatedTileIDOrder)
+    })
+  })
+
+  it.only('Update gridsize when tile added', () => {
```
adorack, line 915: Please remove the `.only`.

adorack (changes requested):

adorack (approved):

PR #2070, opened 2025-04-04
fix: custom chart plugin states leaking across chart instances

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-3723

- Add a new sandbox page displaying multiple charts
- Refactored plugins to use class-based instances. Each chart now creates its own plugin instance rather than relying on a shared singleton.
- Custom tooltip positioning logic was leaking between chart instances of the same type. This was due to Chart.js requiring custom tooltip positioners to be registered globally. As a result, the shared positioner always accessed tooltipState from the last chart instance of that chart type, leading to unexpected behavior. To fix this, each chart instance registers its own unique positioner function. See https://konghq.atlassian.net/browse/MA-3820 
- Cleanup tooltip positioners `onUnomunted`

### DIFF
--- a/packages/analytics/analytics-chart/src/components/chart-plugins/DragSelectPlugin.ts
+++ b/packages/analytics/analytics-chart/src/components/chart-plugins/DragSelectPlugin.ts
@@ -38,7 +38,7 @@ const dispatchEvent = (eventName: string, chart: Chart, pluginInstance: DragSele
 }
 
 
-export const dragSelectPlugin: DragSelectPlugin = {
+export const createDragSelectPlugin = (): DragSelectPlugin => ({
   id: 'dragSelectPlugin',
   isDragging: false,
   startX: 0,
@@ -99,4 +99,4 @@ export const dragSelectPlugin: DragSelectPlugin = {
       drawSelectionArea(chart, this.startX, this.endX)
     }
   },
-}
+})
--- a/packages/analytics/analytics-chart/src/components/chart-plugins/DragSelectPlugin.ts
+++ b/packages/analytics/analytics-chart/src/components/chart-plugins/DragSelectPlugin.ts
@@ -1,6 +1,6 @@
 import type { Chart, Plugin } from 'chart.js'
 
-interface DragSelectPlugin extends Plugin {
+interface IDragSelectPlugin extends Plugin {
   isDragging: boolean
   startX: number
   endX: number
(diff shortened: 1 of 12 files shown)


packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
```diff
@@ -157,10 +157,13 @@ const emit = defineEmits<{
   (e: 'zoom-time-range', newTimeRange: AbsoluteTimeRangeV4): void,
 }>()
 
+const verticalLinePlugin = new VerticalLinePlugin()
+const highlightPlugin = new HighlightPlugin()
+const dragSelectPlugin = new DragSelectPlugin()
 const { translateUnit } = composables.useTranslatedUnits()
 const chartInstance = ref<{ chart: Chart }>()
-const legendID = ref(uuidv4())
```
filipgutica, line 162: None of these need to be reactive

packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
```diff
@@ -157,10 +157,13 @@ const emit = defineEmits<{
   (e: 'zoom-time-range', newTimeRange: AbsoluteTimeRangeV4): void,
 }>()
 
+const verticalLinePlugin = new VerticalLinePlugin()
```
filipgutica, line 160: Create a unique plugin instance per chart instance

packages/analytics/analytics-chart/src/components/chart-plugins/DragSelectPlugin.ts
```diff
@@ -37,14 +37,19 @@ const dispatchEvent = (eventName: string, chart: Chart, pluginInstance: DragSele
   }))
 }
 
+export class DragSelectPlugin implements IDragSelectPlugin {
```
filipgutica, line 40: Since these aren't being shared across packages, only used internally, I think it's a good use case for classes?

packages/analytics/analytics-chart/src/composables/useBarChartOptions.ts
```diff
@@ -29,8 +32,8 @@ export default function useBarChartOptions(chartOptions: BarChartOptions) {
     },
   }
 
-  Tooltip.positioners.barChartTooltipPosition = function(elements, position) {
-    if (!elements.length) {
+  Tooltip.positioners[positionKey] = function(elements, position) {
+    if (!elements.length || chartOptions.tooltipState.locked) {
```
filipgutica, line 36: Same as line chart options, shouldn't run the positioner if tooltip is `locked`

adorack (changes requested):
I'd like to review; will do so first thing next week.

adorack (approved):

PR #2078, opened 2025-04-14
feat(entities-config-editor): migrate from PoC

sumimakito:
# Summary

This pull request migrates the MVP from the previous PoC. The current implementation is not perfect and will be improved by future pull requests.

KM-1107


packages/core/entities-config-editor/src/components/EntitiesConfigEditor.vue
```diff
@@ -1,16 +1,591 @@
...
+
+onMounted(async () => {
+  editor = monaco.editor.create(editorRef.value as HTMLElement, {
+    theme: 'vs',
+    automaticLayout: true,
+    tabSize: 2,
+  })
+
+  makeDetailsVisible()
+
```
TT1228, line 212: Shall we consider adding something like below code to enhance the completion? I found that when key in editor blurred and regained focus, it no longer provide suggestions.
```
  editorDisposables.push(
    editor.onKeyUp((e) => {
      const key = e.browserEvent.key

      if (key.length === 1 && /[a-zA-Z0-9_]/.test(key) && triggerSuggestIfHasItems(editor)) {
        editor.trigger('keyboard', 'editor.action.triggerSuggest', {})
      }
    }),
  )
```

sumimakito: Nice catch. I have updated the config used to initialize the editor. Now users can pick up the suggestions from where they left off without any special key combinations:

demo

packages/core/entities-config-editor/src/composables/useMonacoEditor.ts
```diff
@@ -0,0 +1,88 @@
...
+  const schema = computed(() => {
+    const _luaSchema = unref(luaSchema)
+    if (!_luaSchema) {
+      return undefined
+    }
+
+    const rootSchema = cloneDeep(buildRecordSchema(_luaSchema, unref(language)))
+    rootSchema.properties = {
+      ...rootSchema.properties,
+    }
```
jw-foss, line 41: **q:** why are we copying this property here?

packages/core/entities-config-editor/src/composables/useMonacoEditor.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (existingModel) {
+      existingModel.dispose()
+    }
+    return monaco.editor.createModel(_language === 'json' ? '{}' : '', _language, uri.value)
+  })
+
+  const textDocument = computed(() =>{
+    const _language = unref(language)
+    return TextDocument.create(uri.value.toString(), _language, 0, _language === 'json' ? '{}' : '')
+  })
```
jw-foss, line 57: WDYT about isolating the ternary expression into computed like:

```ts
const langTemplate = computed(() => language.value === 'json' ? '{}' : '')
```

packages/core/entities-config-editor/src/types/fields.ts
```diff
@@ -0,0 +1,143 @@
...
+  schema: S
+  value?: V
+
+  /** DEBUGGING: An stack that contains the information of the parent fields */
+  parentStack?: string[]
+}
+
+export interface FieldEmits<V = any> {
+  (e: 'update-value', value: V): void
+}
```
jw-foss, line 28: **non-blocking** maybe we should use the same definition for Emitter

```suggestion
export interface FieldEmits<V = any> {
  'updateValue': [value: V],
}
```

packages/core/entities-config-editor/src/types/fields.ts
```diff
@@ -0,0 +1,143 @@
...
+  type: typeof FIELD_TYPES.NUMBER | typeof FIELD_TYPES.INTEGER
+
+  one_of?: number[]
+
+  between?: [min: number, max: number]
+  gt?: number
+}
+
+export const isNumberLikeField = (schema: FieldSchema): schema is NumberLikeFieldSchema =>
+  schema.type === FIELD_TYPES.NUMBER || schema.type === FIELD_TYPES.INTEGER
```
jw-foss, line 95: Personally I'd prefer

```suggestion
  [FIELD_TYPES.NUMBER, FIELD_TYPES].includes(schema.type)
```

packages/core/entities-config-editor/src/composables/useMonacoEditor.ts
sumimakito: Thank you for the suggestion. I'm marking this pull request as Draft for now as I plan to refactor the code a little bit. Let me come back with more improvements.

jw-foss: Cool, let me know when it's ready ✉️

PR #2083, opened 2025-04-16
feat(entities-plugins): poc of free-form improving

2eha0:
# Summary


TT1228 (commented):
We could add `id` or `test-id` in Fields for regression tests, maybe in a follow up PR.

packages/entities/entities-plugins/src/components/free-form/RequestCallout/CalloutRequestForm.vue
```diff
@@ -1,269 +1,184 @@
...
-      :placeholder="getPlaceholder('callouts.*.request.method')"
-      required
+      name="method"
     />
 
     <ObjectField
       label="Request › Headers"
-      :label-attributes="getLabelAttributes('callouts.*.request.headers')"
+      name="headers"
       required
```
TT1228, line 17: Can we remove the `required` prop here?

packages/entities/entities-plugins/src/components/free-form/shared/Form.vue
```diff
@@ -0,0 +1,70 @@
...
+    </template>
+  </component>
+</template>
+
+<script setup lang="ts" generic="T extends Record<string, any> = Record<string, any>">
+import { provide, reactive, useSlots, type Slot, watch, toValue } from 'vue'
+import { DATA_INJECTION_KEY, FIELD_RENDERER_MATCHERS_MAP, FIELD_RENDERER_SLOTS, SCHEMA_INJECTION_KEY, useSchemaHelpers, FIELD_RENDERERS } from './composables'
+import type { FormSchema } from '../../../types/plugins/form-schema'
+import Field from './Field.vue'
+import { omit } from 'lodash-es'
```
TT1228, line 26: Shall we use `import omit from 'lodash-es/omit'` instead? Not sure if it will smallen bundle size.

Justineo: IMO we should probably always import from `lodash-es` and let the bundler do the treeshaking.

2eha0:
> We could add `id` or `test-id` in Fields for regression tests, maybe in a follow up PR.

That's a good idea, how about giving each field a default test-id based on its path, like `calloout.*.name`?

packages/entities/entities-plugins/src/components/free-form/shared/Form.vue
2eha0: The tree-shaking still works in this way

packages/entities/entities-plugins/src/components/free-form/shared/StringField.vue
```diff
@@ -39,33 +52,60 @@ const attrs = useAttrs()
...
+  showVaultSecretPicker = undefined,
+  name,
+  ...props
+} = defineProps<StringFieldProps>()
 const emit = defineEmits<{
   'update:modelValue': [value: string | null]
 }>()
 
-const initialValue = modelValue
+const field = useField<string | null>(toRef(() => name))
```
TT1228, line 74: Maybe we could considering use following definition so that we can avoid `field.value?.value` pattern
`const { value: fieldValue, schema: fieldSchema, ...field }` = useField<string | null>(toRef(() => name))

TT1228:
> > We could add `id` or `test-id` in Fields for regression tests, maybe in a follow up PR.
> 
> That's a good idea, how about giving each field a default test-id based on its path, like `calloout.0.name`?

Current `test-ids` have patterns like `config-llm_providers-window_size-0-item-0`, we could follow this pattern for convenience(so that we can reuse current test codes) or use new pattern if we find it better. We can discuss this on stand-up.

packages/entities/entities-plugins/src/components/free-form/shared/StringField.vue
```diff
@@ -39,33 +52,60 @@ const attrs = useAttrs()
...
   return props.multiline ? KTextArea : KInput
 })
 
 const autofillSlot = inject<AutofillSlot | undefined>(AUTOFILL_SLOT, undefined)
-const schema = computed(() => ({ referenceable: showVaultSecretPicker }))
+
+const realShowVaultSecretPicker = computed(() => {
+  if ('showVaultSecretPicker' in props) {
+    return props.showVaultSecretPicker
+  }
```
TT1228, line 104:
```suggestion
  if (showVaultSecretPicker ! ==undefined) {
    return showVaultSecretPicker
  }
```

packages/entities/entities-plugins/src/components/free-form/RequestCallout/CalloutForm.vue
```diff
@@ -1,59 +1,50 @@
...
+          'GET',
+          'POST',
+          'PUT',
+          'DELETE',
+          'PATCH',
+          'HEAD',
+          'OPTIONS',
+          'CONNECT',
+          'TRACE',
+        ])"
```
2eha0, line 25: todo: set `oneOf` to schema, then remove it

packages/entities/entities-plugins/src/components/free-form/RequestCallout/CalloutForm.vue
```diff
@@ -1,62 +1,60 @@
...
-import type { RequestCallout } from './types'
+import ObjectField from '../shared/ObjectField.vue'
+
+import type { RequestCalloutPlugin } from './types'
 import type { SelectItem } from '@kong/kongponents'
 
 const props = defineProps<{
-  index: number;
+  index: number
+  name: string
```
Justineo, line 51: I think `fieldName` is clearer.

packages/entities/entities-plugins/src/components/free-form/shared/utils.ts
```diff
@@ -3,3 +3,30 @@ export function toSelectItems<T extends string>(
 ): { value: T; label: T }[] {
   return items.map((item) => ({ value: item, label: item }))
 }
+
+export const path = {
+  arraySymbol: '*',
+  rootSymbol: '$',
+  separator: '.',
```
Justineo, line 10: Maybe we shouldn't allow to customize these symbols. And I think it's better to export the following methods just as plain functions like `resolvePath`, `resolveRootPath`, etc.

packages/entities/entities-plugins/src/components/free-form/shared/ArrayField.vue
```diff
@@ -9,18 +17,17 @@
...
-        {{ label }}
+        {{ fieldAttrs.label }}
         <template
-          v-if="labelAttributes?.info"
+          v-if="fieldAttrs.labelAttributes?.info"
           #tooltip
         >
           <!-- eslint-disable-next-line vue/no-v-html -->
-          <div v-html="labelAttributes?.info" />
+          <div v-html="fieldAttrs.labelAttributes.info" />
```
jw-foss, line 30: Should we consider escaping the string here?

Also, I am wondering, maybe use a slot here for the tooltip to differentiate  with `tooltipAttributes` would have been nicer.

packages/entities/entities-plugins/src/components/free-form/shared/ArrayField.vue
```diff
@@ -47,10 +54,16 @@
         >
           <div class="ff-array-field-item-content">
             <slot
+              v-if="$slots.item"
+              data-autofocus
+              :field-name="String(index)"
```
jw-foss, line 59: Should we consider receiving a method for constructing the name? For instance:

```ts
const getNameOfField = (idx: number) => {
  return props.generateName(idx) || String(idx)
}
```

packages/entities/entities-plugins/src/components/free-form/shared/Field.vue
```diff
@@ -0,0 +1,88 @@
...
+  {
+    default?: Slot
+    [FIELD_RENDERERS]?: Slot<{ name: string }>
+  } & Record<string, Slot<{ name: string }>>
+>()
+
+const field = useField(toRef(props, 'name'))
+
+const fieldRenderer = computed<any>(() => {
+
```
jw-foss, line 67: I would do something like following to early return from this function:

```ts
if ('one_of' in field.schema.value) {
  return EnumField
}
```

packages/entities/entities-plugins/src/components/free-form/shared/FieldRenderer.vue
```diff
@@ -0,0 +1,35 @@
...
+const inheritedMatchMap = inject<MatchMap>(FIELD_RENDERER_MATCHERS_MAP)
+
+if (!inheritedMatchMap) {
+  throw new Error(FIELD_RENDERER_MATCHERS_MAP.toString() + 'not provided')
+}
+
+if (defaultSlot) {
+  inheritedMatchMap.set(props.match, defaultSlot)
+}
+</script>
```
jw-foss, line 35: Wondering, why don't we just make this a composable instead of `Renderless` component

packages/entities/entities-plugins/src/components/free-form/shared/Form.vue
```diff
@@ -0,0 +1,104 @@
...
+    </template>
+
+    <!-- auto rendering -->
+    <template v-else>
+      <Field
+        v-for="field in childFields"
+        :key="Object.keys(field)[0]"
+        :name="Object.keys(field)[0]"
+      />
+    </template>
```
jw-foss, line 17:
```suggestion
    <!-- custom rendering -->
    <slot>
      <Field
        v-for="field in childFields"
        :key="Object.keys(field)[0]"
        :name="Object.keys(field)[0]"
      />
    </slot>
```

packages/entities/entities-plugins/src/components/free-form/shared/NumberField.vue
```diff
@@ -1,46 +1,92 @@
...
+    fieldValue!.value = null
     emit('update:modelValue', null)
   } else {
+    fieldValue!.value = Number(value)
     emit('update:modelValue', Number(value))
   }
 }
+
+const isAutoFocus = computed(() => {
+  if (attrs['data-autofocus'] !== undefined) return attrs['data-autofocus']
```
jw-foss, line 79: I think for a early return we also need to wrap it into a `{}` pair, also, the return type should be aligned so we might as well just return `true` instead

```suggestion
  if (attrs['data-autofocus'] !== undefined) {
    return true
  }
```

packages/entities/entities-plugins/src/components/free-form/shared/ObjectField.vue
```diff
@@ -1,20 +1,51 @@
...
+      <!-- manually rendering -->
+      <slot v-if="$slots.default" />
+
+      <!-- auto rendering -->
+      <Field
+        v-for="field in childFields"
+        v-else
+        :key="Object.keys(field)[0]"
+        :name="Object.keys(field)[0]"
+      />
```
jw-foss, line 23:
```suggestion
      <slot>
        <Field
          v-for="field in childFields"
          :key="Object.keys(field)[0]"
          :name="Object.keys(field)[0]"
        />
      </slot>
      <!-- auto rendering -->
      
```

packages/entities/entities-plugins/src/components/free-form/shared/StringField.vue
```diff
@@ -39,33 +57,85 @@ const attrs = useAttrs()
...
+
+  // If is child of array then return true
+  const parent = field.ancestors?.value.parent
+  if (parent?.path) {
+    const parentType = getSchema(parent.path)?.type
+    return parentType === 'array'
+  }
+
+  return false
+})
```
jw-foss, line 133: Saw this quite a few times, maybe consider moving to a composable

packages/entities/entities-plugins/src/components/free-form/shared/composables.ts
```diff
@@ -189,5 +213,248 @@ export function useFormShared<T>() {
...
+    const realPath = parentLabelPath.value ?? fieldName
+    const parts = pathUtils.toArray(realPath)
+
+    const res = parts
+      .map(fieldName => fieldName
+        .replaceAll('_', ' ')
+        .split(' ')
+        .map(word => word.charAt(0).toUpperCase() + word.slice(1))
+        .map(replaceByDictionary)
+        .join(' '))
```
jw-foss, line 377: I am not sure where this is used, but seems like we can simplify this to 

```suggestion
        .split('_')
        .map(_.capitalize)
        .map(replaceByDictionary)
        .join(' '))
```

packages/entities/entities-plugins/src/components/free-form/shared/FieldRenderer.vue
```diff
@@ -0,0 +1,35 @@
...
+export type MatchMap = Map<Match, Slot<{ name: string }>>
+
+const props = defineProps<{
+  match: Match
+  priority?: number
+}>()
+
+defineSlots<Record<string, Slot<{ name: string }>>>()
+
+const { default: defaultSlot } = useSlots()
```
Justineo, line 24:
```suggestion
const { default: defaultSlot } = defineSlots<Record<string, Slot<{ name: string }>>>()
```

packages/entities/entities-plugins/src/components/free-form/shared/FieldRenderer.vue
Justineo: We use this component to collect render functions which are defined in the templates.

packages/entities/entities-plugins/src/components/free-form/shared/ArrayField.vue
2eha0: The field name should be part of the field path, I called it field-name because it conflicts with the slot's name, so it shouldn't be constructed by users.

packages/entities/entities-plugins/src/components/free-form/shared/ArrayField.vue
2eha0: Rendering HTML here is intended I think, because it's markdown in schema, we [parsed it to HTML](https://github.com/Kong/public-ui-components/blob/cdee76d8c708bf4275abdc8ce08ff154b8b3b810/packages/entities/entities-plugins/src/components/free-form/shared/composables.ts#L164) so it can be rendered correctly

Using a slot is a good idea, I'd wrap this `div` with a `<slot name="tooltip">`

jw-foss (approved):
No further questions from my side. 👍 Approving

PR #2090, opened 2025-04-24
fix(key-auth-plugin): temporary add region info to the realm scope

nekolab:
This is a workaround to address a DP-side identity realm issue https://kongstrong.slack.com/archives/C06FJ632KJS/p1745429119460859

JIRA: KM-1205

### DIFF
--- a/packages/entities/entities-plugins/src/components/fields/KeyAuthIdentityRealms.vue
+++ b/packages/entities/entities-plugins/src/components/fields/KeyAuthIdentityRealms.vue
@@ -58,6 +58,18 @@ const kCurrentCPSelectItem = {
 const formConfig = inject<KonnectBaseFormConfig | KongManagerBaseFormConfig | undefined>(FORMS_CONFIG)
 const { axiosInstance } = useAxios(formConfig?.axiosRequestConfig)
 
+// workaround for the fact that the DP does not support the region: null now
+// we should remove this line when the DP supports it
+const guessRegion = () => {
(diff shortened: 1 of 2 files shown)


packages/entities/entities-plugins/src/components/fields/KeyAuthIdentityRealms.vue
```diff
@@ -111,7 +123,7 @@ const onRealmsUpdate = (currentSelected: string[]) => {
     ? currentKonnectRealmSelection.filter((realmId) => !prevKonnectRealmSelection.includes(realmId))
     : currentKonnectRealmSelection
 
-  const nextKonnectRealms = nextKonnectRealmId.map((id) => ({ scope: 'realm', id } as const))
+  const nextKonnectRealms = nextKonnectRealmId.map((id) => ({ ...guessRegion(), scope: 'realm', id } as const))
```
jw-foss, line 126: Do you think if we can invoke this once instead of every item?

TT1228 (approved):

packages/entities/entities-plugins/src/components/fields/KeyAuthIdentityRealms.vue
nekolab: Good catch! I didn't realize it before.
Luckily, the `nextKonnectRealmId` typically carries only 1 item, so it's safe to write like this here.

PR #2091, opened 2025-04-24
fix: show tooltip on warning icon and limit size ma-3830 ma-1889

Chalks:
# Summary

Fixes [MA-3830](https://konghq.atlassian.net/browse/MA-3830) and [MA-1889](https://konghq.atlassian.net/browse/MA-1889).

* Limit/truncate warning tooltip works when hovering over the icon
* Tooltip content is restricted to 50% viewport width so it doesn't bleed off screen
* Sandbox "No Limit" toggle now actually does something

The tooltip wasn't working before because if an icon has the `decorative` prop, it applies the css `pointer-events: none` which means hover effects don't actually trigger. In this case, the icon was the only component that could trigger the tooltip. Merely wrapping it in a `<div>` would have fixed it as well, but I figured it was better to just remove the `decorative` prop.

"No Limit" I don't think has been working [since 2023](https://github.com/Kong/public-ui-components/commit/03df6ec7e33daf8596c9b5bf27b7af1109073732#diff-03e87b68d35d6befcad08a5c5df10076996c356dffafbbc413f59510280cf23bL381) (note that `limitToggle` got removed, probably by accident).

Icons have [pointer events none here](https://github.com/Kong/icons/blob/main/src/__template__/ComponentTemplate.vue#L101)

**Behavior after changes**

tooltip

no-limit-after

**No limit behavior before changes**

no-limit-before

[MA-3830]: https://konghq.atlassian.net/browse/MA-3830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MA-3829]: https://konghq.atlassian.net/browse/MA-3829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MA-1889]: https://konghq.atlassian.net/browse/MA-1889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics/analytics-chart/sandbox/pages/BarChartDemo.vue
+++ b/packages/analytics/analytics-chart/sandbox/pages/BarChartDemo.vue
@@ -328,17 +328,22 @@ const exploreResult = computed<ExploreResultV4>(() => {
     }
   }, {})
 
+  const metaOverrides: Partial<QueryResponseMeta> = {
+    truncated: limitToggle.value,
+    limit: limitToggle.value ? 10 : 50,
+  }
+
   if (multiDimensionToggle.value) {
     return generateCrossSectionalData([{
       name: selectedMetric.value.name,
       unit: selectedMetric.value.unit,
-    }], dimensionMap)
+    }], dimensionMap, metaOverrides)
   } else {
     return generateCrossSectionalData([{
       name: selectedMetric.value.name,
       unit: selectedMetric.value.unit,
     }, ...(multiMetricToggle.value ? secondaryMetrics.value : [])],
-    dimensionMap)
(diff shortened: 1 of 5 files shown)


adorack (changes requested):
I'd like to review.

adorack (approved):

PR #2093, opened 2025-04-25
feat(app-layout): expose slot `title` for `<AppAboutSection />`

nekolab:
This can help us do more customization on the title of the `<AppAboutSection />` component.

The `defineProps` now changes to the TypeScript-style definition with `withDefaults`, which makes Cypress typing system happy in the component test file

JIRA: KM-1134

### DIFF
--- a/packages/core/app-layout/src/components/aboutSection/AppAboutSection.cy.ts
+++ b/packages/core/app-layout/src/components/aboutSection/AppAboutSection.cy.ts
@@ -91,6 +91,8 @@ describe('<AppAboutSection />', () => {
     })
 
     cy.get('.kong-ui-app-about-section').should('exist')
+    cy.getTestId('about-section-title').should('be.visible')
+    cy.getTestId('about-section-title').should('contain.text', title)
     cy.getTestId('about-section-actions').should('be.visible')
     cy.getTestId('about-section-actions').should('contain.text', actionsText)
     cy.getTestId('about-section-content').should('be.visible')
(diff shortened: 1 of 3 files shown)


packages/core/app-layout/src/components/aboutSection/AppAboutSection.cy.ts
```diff
@@ -80,10 +80,10 @@ describe('<AppAboutSection />', () => {
 
     cy.mount(AppAboutSection, {
       props: {
-        title,
         description,
       },
       slots: {
+        title,
```
adamdehaven, line 86: Rather than changing the test for a slot, can we leave this test as-is and write a new test that covers the slot?

packages/core/app-layout/src/components/aboutSection/AppAboutSection.vue
```diff
@@ -4,14 +4,18 @@
...
       <span
         class="about-section-title"
         data-testid="about-section-title"
       >
-        {{ title }}
+        <slot
+          v-if="$slots.title"
+          name="title"
+        />
+        <template v-else>{{ title }}</template>
```
adamdehaven, line 18: This can be simplified by utilizing the prop value as the default slot content.
```suggestion
        <slot name="title">
          {{ title }}
        </slot>
```

packages/core/app-layout/src/components/aboutSection/AppAboutSection.cy.ts
nekolab: Updated

packages/core/app-layout/src/components/aboutSection/AppAboutSection.vue
nekolab: Yes, you are right!

packages/core/app-layout/src/components/aboutSection/AppAboutSection.cy.ts
```diff
@@ -100,6 +102,47 @@ describe('<AppAboutSection />', () => {
...
+        title: titleProps,
+      },
+    })
+
+    cy.get('.kong-ui-app-about-section').should('exist')
+    cy.getTestId('about-section-title').should('be.visible')
+    cy.getTestId('about-section-title').should('contain.text', titleProps)
+
+    // Now mount with slot title
+    cy.mount(AppAboutSection, {
```
adamdehaven, line 121: Can you split this into separate tests rather than mounting the component twice?

adamdehaven (approved):
lgtm

PR #2097, opened 2025-04-29
chore(deps): removes swrv from @kong-ui-public/core

johncowen:
# Summary

We noticed that it seems like swrv is no longer used in this package. This PR removes it form the package manifest.


adamdehaven (approved):
This LGTM but won’t generate a release since you used `chore` - you’d need to use `fix` or leave as-is and the next dependency bump will auto release

PR #2099, opened 2025-04-30
fix(analytics): single value chart show empty when null [KHCP-16024]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-16024

When metric value is `null`, show empty state instead of error message in SingleValue chart

https://github.com/user-attachments/assets/78a62887-36a1-4772-8dfa-c1a28990d4db

### DIFF
--- a/packages/analytics/analytics-chart/src/components/SimpleChart.cy.ts
+++ b/packages/analytics/analytics-chart/src/components/SimpleChart.cy.ts
@@ -117,6 +117,24 @@ describe('<SimpleChart />', () => {
     cy.getTestId('single-value-chart').contains(value)
   })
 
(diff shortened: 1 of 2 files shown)


filipgutica (changes requested):
Could you add test case or 2 in "AnalyticsChart.cy.ts" for single value chart to cover this case

portikM:
> Could you add test case or 2 in "AnalyticsChart.cy.ts" for single value chart to cover this case

Good call -- added test cases for this in `SimpleChart.cy.ts` (since SingleValue is simple chart)

filipgutica (approved):

PR #2100, opened 2025-05-01
fix(deps): remove recursive kongponents dependency

adamdehaven:
# Summary

Remove the `@kong/kongponents` dependency that ends up being recursive in the Kongponents repo. It's already a devDependency here at the root so shouldn't be an issue.

### DIFF
--- a/packages/core/sandbox-layout/package.json
+++ b/packages/core/sandbox-layout/package.json
@@ -40,7 +40,6 @@
   },
   "devDependencies": {
     "@kong/design-tokens": "1.17.3",
-    "@kong/kongponents": "9.29.3",
     "vue": "^3.5.13",
     "vue-router": "^4.4.5"
   },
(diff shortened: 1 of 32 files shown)


johncowen:
Just checking, but did you notice that the kongponents dependency that is needed to be installed here is no longer specified in the package manifest for this package (unless I missed something)?

adamdehaven:
> Just checking, but did you notice that the kongponents dependency that is needed to be installed here is no longer specified in the package manifest for this package (unless I missed something)?

Yes, thinking about this again I supposed I could leave it as a devDependency and just remove the peer

johncowen:
Ah this:

> It's already a devDependency here at the root so shouldn't be an issue.

👀

packages/core/sandbox-layout/package.json
```diff
@@ -62,7 +62,6 @@
     "errorLimit": "200KB"
   },
   "peerDependencies": {
-    "@kong/kongponents": "^9.29.3",
```
johncowen, line 65: This is the only relevant change here right? If so I'm glad we are removing unnecessary peerDependencies 🎉

johncowen (approved):
LGTM! 🎉

portikM (approved):

PR #2110, opened 2025-05-07
feat(analytics): support dashboard tile duplication [ma-3824]

Chalks:
# Summary

Satisfies [MA-3824](https://konghq.atlassian.net/browse/MA-3824)

* Add a `duplicate-tile` event to `<DashboardTile>`
* Bubble up the `duplicate-tile` event in `<DashboardRenderer>` - anything that uses DashboardRenderer will need to handle this event.
* Slightly refactor `DashboardTile.cy.ts` to have a common `mount` method so that the tests are easier to read.
* **fix**: long tile titles looked very bad (visible in prod). Now they don't.

In action:

https://github.com/user-attachments/assets/036caf05-9563-49e7-bbbf-783f2a6b4c96

**Before tile title fix:**

Screenshot 2025-05-07 at 4 26 54 PM

**After tile title fix:**

Screenshot 2025-05-07 at 4 27 14 PM

[MA-3824]: https://konghq.atlassian.net/browse/MA-3824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
+++ b/packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
@@ -31,6 +31,7 @@
           :query-ready="queryReady"
           :refresh-counter="refreshCounter"
           :tile-id="tile.id"
+          @duplicate-tile="onDuplicateTile(tile)"
           @edit-tile="onEditTile(tile)"
           @remove-tile="onRemoveTile(tile)"
           @zoom-time-range="emit('zoom-time-range', $event)"
@@ -42,7 +43,7 @@
 
 <script setup lang="ts">
 import type { DashboardRendererContext, DashboardRendererContextInternal, GridTile } from '../types'
-import type { AbsoluteTimeRangeV4, DashboardConfig, TileConfig, TileDefinition } from '@kong-ui-public/analytics-utilities'
+import type { AbsoluteTimeRangeV4, DashboardConfig, TileConfig, SlottableOptions, TileDefinition } from '@kong-ui-public/analytics-utilities'
 import DashboardTile from './DashboardTile.vue'
 import { computed, inject, ref } from 'vue'
 import type { ComponentPublicInstance } from 'vue'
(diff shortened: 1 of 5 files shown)


packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -312,13 +324,13 @@ const exportCsv = () => {
...
-        overflow: hidden;
-        text-overflow: ellipsis;
-        white-space: nowrap;
-      }
+    .title {
+      font-weight: var(--kui-font-weight-bold, $kui-font-weight-bold);
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
     }
```
Chalks, line 334: this was the only change necessary to fix the tile titles. I'm pretty sure something in KTooltip is swallowing these styles since the dom structure is:

`<KTooltip class="title-tooltip"><div class="title">...`

Either that, or deeply nested scoped css is breaking?

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -199,6 +200,36 @@ const onEditTile = (tile: GridTile<TileDefinition>) => {
   emit('edit-tile', tile)
 }
 
+const isSlottable = (chart: any): chart is SlottableOptions => {
+  return chart.type === 'slottable'
+}
```
Chalks, line 205: `slottable` charts don't have a `chartTitle`. Is there a better way around this?

adorack: I don't think there's a better way around this... we don't currently have any known cases where `slottable` would ever show up in user-editable dashboards, but it's still a case the code needs to handle.

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
adorack: I _think_ the `KTooltip` doesn't automatically render a DOM element, so these styles only look nested.

adorack (approved):

PR #2120, opened 2025-05-13
fix(entities): fetcher url reactivity

Leopoldthecoder:
# Summary

This PR fixes a bug where `config` reactivity is lost in `useFetchUrlBuilder`.

KM-920
KM-1264

### DIFF
--- a/packages/entities/entities-shared/src/composables/useFetchUrlBuilder.ts
+++ b/packages/entities/entities-shared/src/composables/useFetchUrlBuilder.ts
@@ -1,4 +1,4 @@
-import { computed, ref, unref } from 'vue'
+import { computed, ref, unref, toValue, type MaybeRefOrGetter } from 'vue'
 import type {
   KongManagerBaseTableConfig,
   KonnectBaseTableConfig,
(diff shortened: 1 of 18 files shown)


packages/entities/entities-shared/src/composables/useFetchUrlBuilder.ts
```diff
@@ -7,15 +7,15 @@ import type { MaybeRef } from '../types/utils'
 import type { TableDataFetcherParams } from '@kong/kongponents'
 
 export default function useFetchUrlBuilder(
-  config: MaybeRef<KonnectBaseTableConfig | KongManagerBaseTableConfig>,
+  config: MaybeRefOrGetter<KonnectBaseTableConfig | KongManagerBaseTableConfig>,
   baseUrl: MaybeRef<string>,
 ) {
```
jw-foss, line 12: @Leopoldthecoder, I saw the `baseUrl` was marked to `MaybeRef` as well, maybe it is better if we pass a ref as well?

jw-foss (approved):

packages/entities/entities-shared/src/composables/useFetchUrlBuilder.ts
Leopoldthecoder: Updated.

PR #2121, opened 2025-05-13
refactor(free-form): handle advanced fields in free form [KM-1265]

2eha0:
# Summary
[KM-1265](https://konghq.atlassian.net/browse/KM-1265)
Let the three fixed advanced fields (instance_name, tags and protocols) be handled by the free form.

[KM-1265]: https://konghq.atlassian.net/browse/KM-1265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-plugins/src/components/free-form/shared/Field.vue
+++ b/packages/entities/entities-plugins/src/components/free-form/shared/Field.vue
@@ -47,6 +47,7 @@ import ObjectField from './ObjectField.vue'
 import NumberField from './NumberField.vue'
 import EnumField from './EnumField.vue'
 import KeyValueField from './KeyValueField.vue'
+import TagField from './TagField.vue'
 
 defineOptions({ name: 'AutoField' })
 
@@ -76,8 +77,8 @@ const fieldRenderer = computed(() => {
     case 'array':
       return ArrayField
     case 'set':
-      if (utils.isStringSet(field.schema)) {
-        return StringField
+      if (utils.isTagField(field.schema)) {
+        return TagField
       }
       return EnumField
     case 'record':
--- a/packages/entities/entities-plugins/src/components/free-form/shared/Field.vue
+++ b/packages/entities/entities-plugins/src/components/free-form/shared/Field.vue
@@ -76,6 +76,9 @@ const fieldRenderer = computed(() => {
     case 'array':
       return ArrayField
     case 'set':
+      if (utils.isStringSet(field.schema)) {
+        return StringField
+      }
       return EnumField
     case 'record':
       return ObjectField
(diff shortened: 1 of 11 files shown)


packages/entities/entities-plugins/src/components/free-form/shared/composables.ts
```diff
@@ -170,11 +170,17 @@ export function useSchemaHelpers(schema: MaybeRefOrGetter<FormSchema>) {
...
   function getSelectItems(fieldPath: string): SelectItem[] {
     const schema = getSchema(fieldPath)
-    return utils.toSelectItems((schema?.one_of || []))
+    return utils.toSelectItems((schema?.one_of || (schema as ArrayLikeFieldSchema).elements?.one_of || []))
   }
 
   function getPlaceholder(fieldPath: string): string | null {
-    const defaultValue = getSchema(fieldPath)?.default
+    const schema = getSchema(fieldPath)
+
```
TT1228, line 178: The logic of placeholder is slightly different in VFG(https://github.com/Kong/public-ui-components/blob/main/packages/entities/entities-plugins/src/components/PluginForm.vue#L893). 
Do we have a reason for prioritizing `schema.help`? Also, we may need to consider whether the consuming app is Konnect.

packages/entities/entities-plugins/src/components/free-form/shared/Field.vue
```diff
@@ -76,6 +76,9 @@ const fieldRenderer = computed(() => {
     case 'array':
```
TT1228, line 76: How does `FreeForm` handles array item with elements of `one_of`?
VFG has them rendered with input. We could discuss how they should be rendered in `FreeForm`.

2eha0: We can stay consistent with the original util we have a new design, but maybe in another PR.

packages/entities/entities-plugins/src/components/free-form/shared/utils.ts
```diff
@@ -1,10 +1,12 @@
+import type { UnionFieldSchema } from 'src/types/plugins/form-schema'
+import { toValue, type MaybeRefOrGetter } from 'vue'
+
```
jw-foss, line 3: nit(non-blocking): can we unify the code style here by importing the type individually with a `import type` statement instead of `import { type } ` 🤔

jw-foss (approved):
I left a non-blocking comment regarding coding style, the rest LGTM. 👍

TT1228 (approved):
LGTM

packages/entities/entities-plugins/src/components/free-form/shared/utils.ts
2eha0: Sure, most likely it was generated by VSCode, will be fixed in next PR.

PR #2131, opened 2025-05-16
feat(entities-shared): add a prop to persist format preference

sumimakito:
# Summary

This pull request adds an optional prop to persist the format preference. When omitted, the preference will not be persisted.

KM-1278


Leopoldthecoder (approved):
LGTM

PR #2134, opened 2025-05-20
feat(consumer-groups): stop fetching the consumers when no need

jw-foss:
# Summary

Stop fetching consumers data when the consumer field is hidden from the user.


sumimakito (approved):
Looks Cool To Me

PR #2135, opened 2025-05-20
feat: make gridSize optional (and unused) calculate from tiles instead [ma-3916]

Chalks:
# Summary

[satisfies ma-3916](https://konghq.atlassian.net/browse/MA-3916)

* set default columns in a dashboard to 6 (using the `DASHBOARD_COLS` constant
* remove `gridSize` everywhere it exists in this repo
* calculate the values of gridsize where it was used from tile definitions instead

After this is merged, will need to merge the following PRs as well, please consider reviewing them:

* https://github.com/Kong/flink-jobs/pull/1639
* https://github.com/Kong/konnect-ui-apps/pull/6829
* https://github.com/Kong/kanalytics/pull/948
* https://github.com/Kong/portal/pull/1120

### DIFF
--- a/packages/analytics/analytics-utilities/src/dashboardSchema.ts
+++ b/packages/analytics/analytics-utilities/src/dashboardSchema.ts
@@ -553,7 +553,7 @@ export const dashboardConfigSchema = {
     },
     global_filter: filtersFn([...new Set([...filterableExploreDimensions, ...filterableBasicExploreDimensions, ...filterableAiExploreDimensions])]),
   },
-  required: ['tiles', 'gridSize'],
+  required: ['tiles'],
   additionalProperties: false,
 } as const satisfies JSONSchema
 
--- a/packages/analytics/dashboard-renderer/README.md
+++ b/packages/analytics/dashboard-renderer/README.md
@@ -92,11 +92,6 @@ const context: DashboardRendererContext = {
 }
 
 const config: DashboardConfig = {
-  // 4 x 1 grid
-  gridSize: {
-    cols: 4,
-    rows: 1,
-  },
   tiles: [
     {
       id: 'unique-tile-id', // Required for editable dashboards
@@ -116,9 +111,9 @@ const config: DashboardConfig = {
           col: 0,
           row: 0,
         },
-        // Spans 2 columns and 1 rows
+        // Spans 3 columns and 1 rows
         size: {
-          cols: 2,
+          cols: 3,
           rows: 1,
         }
       }
@@ -138,14 +133,14 @@ const config: DashboardConfig = {
         },
       },
       layout: {
-        // Position at column 2, row 0
+        // Position at column 3, row 0
         position: {
-          col: 2,
+          col: 3,
           row: 0,
         },
-        // Spans 2 columns and 1 rows
+        // Spans 3 columns and 1 rows
         size: {
-          cols: 2,
+          cols: 3,
           rows: 1,
         }
       }
@@ -184,11 +179,6 @@ const context: DashboardRendererContext = {
 }
 
 const config: DashboardConfig = {
-  // 4 x 1 grid
-  gridSize: {
-    cols: 4,
-    rows: 1,
-  },
   tiles: [
     {
       // Line chart
@@ -208,9 +198,9 @@ const config: DashboardConfig = {
           col: 0,
           row: 0,
         },
-        // Spans 2 columns and 1 rows
+        // Spans 3 columns and 1 rows
         size: {
-          cols: 2,
+          cols: 3,
           rows: 1,
         }
       }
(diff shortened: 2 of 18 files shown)


packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -249,9 +248,6 @@ const handleUpdateTiles = (tiles: GridTile<TileDefinition>[]) => {
     } as TileConfig
   })
 
-  // Update `rows` to match the number of tiles we've placed.
-  // `columns` remains fixed; this is set by design requirements rather than the number of tiles.
-  model.value.gridSize.rows = Math.max(1, ...updatedTiles.map(t => t.layout.position.row + t.layout.size.rows))
```
Chalks, line 254: no longer necessary, but is effectively the same code as what's happening in [grid-utils.ts](https://github.com/Kong/public-ui-components/pull/2135/files#diff-e3b34364ff092f11b2420a0a7acbed33eb37bc861ce717025db8540cf36195a7R12) now

packages/analytics/dashboard-renderer/src/components/layout/DraggableGridLayout.vue
```diff
@@ -29,22 +29,21 @@
...
-  gridSize: GridSize
 }
 
 const props = withDefaults(defineProps<{
   tiles: GridTile<T>[]
-  gridSize: GridSize
   tileHeight?: number
 }>(), {
-  tileHeight: 200,
+  tileHeight: DEFAULT_TILE_HEIGHT,
```
Chalks, line 46: was there a reason this was `200` instead of the default?

adamdehaven (changes requested):
issue: you're removing a prop. Shouldn't you issue a breaking change commit + major version bump?

Chalks:
> issue: you're removing a prop. Shouldn't you issue a breaking change commit + major version bump?

almost certainly. I haven't done that before here, can you show me how to do that (or point me to documentation)? @adamdehaven

adamdehaven:
> > issue: you're removing a prop. Shouldn't you issue a breaking change commit + major version bump?
> 
> almost certainly. I haven't done that before here, can you show me how to do that (or point me to documentation)? @adamdehaven

The easiest way is to stage your changes, and then run `pnpm commit` (instead of `git commit`) and it will walk you through creating the breaking commit message.

adorack (changes requested):
I'd like to review.

packages/analytics/analytics-utilities/src/dashboardSchema.ts
```diff
@@ -537,23 +537,9 @@ export const dashboardConfigSchema = {
       type: 'number',
       description: 'Height of each tile in pixels.',
     },
-    gridSize: {
```
adorack, line 540: I changed my mind about this: because it's going to take longer to get to the version migration than I had thought, and because there are a number of other `analytics-utilities` changes in flight, could you please only remove `gridSize` from `required` here, rather than getting rid of the whole schema?  We're going to remove it entirely as part of the version migration, but it'll make our lives easier to allow it in the schema for now, even if we don't reference it anywhere.

packages/analytics/dashboard-renderer/README.md
```diff
@@ -433,10 +414,6 @@ The root configuration type for a dashboard.
 interface DashboardConfig {
   tiles: TileConfig[]         // Array of tile configurations
   tileHeight?: number         // Optional height of each tile in pixels
-  gridSize: {                 // Required grid layout configuration
```
adorack, line 436: Thanks for updating this; even if we leave it in the schema for now, it's still good to remove it from the README.

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.cy.ts
```diff
@@ -912,63 +901,6 @@ describe('<DashboardRenderer />', () => {
...
-        },
-      })
-    })
-
-    cy.wrap(configRef).should((ref: Ref<DashboardConfig>) => {
-      expect(ref.value.gridSize).to.deep.equal({ cols: 8, rows: 8 })
-    })
-  })
-
   it('Preserve column counts', () => {
```
adorack, line 904: You can remove this entire test; its only purpose was making sure we set `gridSize` correctly.

packages/analytics/dashboard-renderer/src/components/layout/DraggableGridLayout.vue
adorack: I'm not sure.  I don't see any issues with changing it, however.

adorack (changes requested):

packages/analytics/analytics-utilities/src/dashboardSchema.ts
Chalks: I guess that means it's not a breaking change anymore, eh?

PR #2141, opened 2025-05-22
feat: combine context and global filters

alxgrant:
# Summary

### DIFF
--- a/packages/analytics/analytics-utilities/src/dashboardSchema.ts
+++ b/packages/analytics/analytics-utilities/src/dashboardSchema.ts
@@ -551,7 +551,7 @@ export const dashboardConfigSchema = {
       required: ['cols', 'rows'],
       additionalProperties: false,
     },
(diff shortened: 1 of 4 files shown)


packages/analytics/analytics-utilities/src/dashboardSchema.ts
```diff
@@ -551,7 +551,7 @@ export const dashboardConfigSchema = {
       required: ['cols', 'rows'],
       additionalProperties: false,
     },
-    global_filter: filtersFn([...new Set([...filterableExploreDimensions, ...filterableBasicExploreDimensions, ...filterableAiExploreDimensions])]),
+    global_filters: filtersFn([...new Set([...filterableExploreDimensions, ...filterableBasicExploreDimensions, ...filterableAiExploreDimensions])]),
```
alxgrant, line 554: `global_filter` was what was mentioned in ticket, but I think `global_filters` (plural) is more consistent with the `filters` we use elsewhere.

PR #2142, opened 2025-05-26
feat(entities-vaults): add cert based authentication to hcv [KM-1303]

2eha0:
# Summary
[KM-1303](https://konghq.atlassian.net/browse/KM-1303)
[Schema changing](https://github.com/Kong/kong-ee/pull/12418)

image

[KM-1303]: https://konghq.atlassian.net/browse/KM-1303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-vaults/src/components/VaultForm.vue
+++ b/packages/entities/entities-vaults/src/components/VaultForm.vue
@@ -277,6 +277,7 @@
                 { label: VaultAuthMethods.TOKEN, value: VaultAuthMethods.TOKEN as string },
                 { label: VaultAuthMethods.K8S, value: VaultAuthMethods.K8S as string },
                 ...(config.hcvAppRoleMethodAvailable ? [{ label: VaultAuthMethods.APP_ROLE, value: VaultAuthMethods.APP_ROLE as string }] : []),
+                ...(config.hcvCertMethodAvailable ? [{ label: VaultAuthMethods.CERT, value: VaultAuthMethods.CERT as string }] : []),
               ]"
               :label="t('form.config.hcv.fields.auth_method.label')"
               :readonly="form.isReadonly"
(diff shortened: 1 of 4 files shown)


packages/entities/entities-vaults/src/components/VaultForm.vue
```diff
@@ -371,6 +372,41 @@
...
+              class="vault-form-config-auth-method-container"
+            >
+              <KInput
+                v-model.trim="configFields[VaultProviders.HCV].cert_auth_role_name"
+                autocomplete="off"
+                data-testid="vault-form-config-hcv-cert_auth_role_name"
+                :label="t('form.config.hcv.fields.cert_auth_role_name.label')"
+                :readonly="form.isReadonly"
+                required
+                type="text"
```
jw-foss, line 386: I think we can remove line L#386 since the component `KInput` defaults to `type="text"`

packages/entities/entities-vaults/src/components/VaultForm.vue
```diff
@@ -371,6 +372,41 @@
...
+              v-else-if="configFields[VaultProviders.HCV].auth_method === VaultAuthMethods.CERT"
+              class="vault-form-config-auth-method-container"
+            >
+              <KInput
+                v-model.trim="configFields[VaultProviders.HCV].cert_auth_role_name"
+                autocomplete="off"
+                data-testid="vault-form-config-hcv-cert_auth_role_name"
+                :label="t('form.config.hcv.fields.cert_auth_role_name.label')"
+                :readonly="form.isReadonly"
+                required
```
jw-foss, line 385: **Nit:** I was wondering if we can do something like the following:

```ts
const formFieldsCommonAttrs = computed(() => ({
    autocomplete: "off",
    readonly: form.isReadonly,
    required: true,
  })
)
```

then use them as following:

```vue
<KInput v-bind="formFieldsCommonAttrs">
```

jw-foss (commented):
@2eha0 changes LGTM, only have two code style comments for your considerations 👍

packages/entities/entities-vaults/src/components/VaultForm.vue
2eha0: `required` is not that common I think.

2eha0: For these three new fields, it's just a **coincidence** that they are all required, for rest of fields, required is not that common, it depends on the schema.

PR #2148, opened 2025-05-28
feat(analytics): calculate the unit from aggregation name [ma-3976]

Chalks:
# Summary

Satisfies [MA-3976](https://konghq.atlassian.net/browse/MA-3976)

We need aggregation units for data generation to support dashboard previews. Currently this is determined [here](https://github.com/Kong/kanalytics/blob/eb58d6cd27a77083c18d7556e79d88aeb20bbd41/src/druid/druid.service.ts#L58-L73). Instead we're moving that logic into analytics-utilities as that's a more clear place for it to live and allows a lot more things to use it.

I'll have a followup PR to remove that code from kanalytics and import from here instead.


adorack:
(Pausing until we have a chance to discuss further.)

PR #2150, opened 2025-05-30
test(free-form): add test-id [KM-1313]

2eha0:
# Summary
[KM-1313](https://konghq.atlassian.net/browse/KM-1313)

[KM-1313]: https://konghq.atlassian.net/browse/KM-1313?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-plugins/src/components/free-form/shared/BooleanField.vue
+++ b/packages/entities/entities-plugins/src/components/free-form/shared/BooleanField.vue
@@ -9,7 +9,9 @@
   <KCheckbox
     v-else
     v-bind="fieldAttrs"
+    :id="field.path.value"
     class="ff-boolean-field"
+    :data-testid="field.path.value"
     :model-value="fieldValue ?? false"
     @update:model-value="handleUpdate"
   >
--- a/packages/entities/entities-plugins/src/components/free-form/shared/BooleanField.vue
+++ b/packages/entities/entities-plugins/src/components/free-form/shared/BooleanField.vue
@@ -9,7 +9,6 @@
   <KCheckbox
     v-else
     v-bind="fieldAttrs"
-    :id="field.path.value"
     class="ff-boolean-field"
     :data-testid="field.path.value"
     :model-value="fieldValue ?? false"
--- a/packages/entities/entities-plugins/src/components/free-form/shared/KeyValueField.vue
+++ b/packages/entities/entities-plugins/src/components/free-form/shared/KeyValueField.vue
@@ -16,7 +16,7 @@
         class="ff-kv-field-label"
         v-bind="fieldAttrs"
         :data-testid="`label-${field.path.value}`"
-        :for="`${field.path.value}.key`"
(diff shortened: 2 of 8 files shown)


packages/entities/entities-plugins/src/components/free-form/shared/BooleanField.vue
```diff
@@ -9,7 +9,9 @@
   <KCheckbox
     v-else
     v-bind="fieldAttrs"
+    :id="field.path.value"
```
Justineo, line 12: For per instance `id`s, I think we should use `useId` from `vue` directly.

packages/entities/entities-plugins/src/components/free-form/shared/KeyValueField.vue
```diff
@@ -43,16 +45,20 @@
       class="ff-kv-field-entry"
     >
       <KInput
+        :id="`${field.path.value}.key`"
         v-model.trim="entry.key"
         class="ff-kv-field-entry-key"
         :data-key-input="index"
+        :data-testid="`${field.path.value}.key`"
```
Justineo, line 52: We should adopt a consistent naming convention for these `testid`s.

`{part}-{path}` or `{path}.{part}`? Maybe the former one is better as the latter may make `part` looks like a part of the `path`.

TT1228 (commented):
Shall we use 
```
<component
    :is="fieldRenderer"
    v-if="fieldRenderer"
    :id="field.path.value"
    :data-testid="field.path.value"
    :name="utils.resolveRoot(field.path.value)"
  />
```
in `Field.vue` instead of writing in each field implementation.

packages/entities/entities-plugins/src/components/free-form/shared/BooleanField.vue
2eha0: Why `useId` is better than a meaningful id?

2eha0:
> Shall we use
> 
> ```
> <component
>     :is="fieldRenderer"
>     v-if="fieldRenderer"
>     :id="field.path.value"
>     :data-testid="field.path.value"
>     :name="utils.resolveRoot(field.path.value)"
>   />
> ```
> 
> in `Field.vue` instead of writing in each field implementation.

It only works when use `Field` for auto rendering, if I use specific field than we miss the test-id

packages/entities/entities-plugins/src/components/free-form/shared/KeyValueField.vue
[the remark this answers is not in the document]
2eha0: Why do we need a test-id for field wrapper? If we do need it, we should add it for `ObjectField`, `ArrayField` as well.

TT1228 (approved):
LGTM

PR #2157, opened 2025-06-03
fix(entities-plugins): use symbol to prune private fields

Justineo:
# Summary

[KM-1304](https://konghq.atlassian.net/browse/KM-1304)

- Skip pruning config data against plugin schemas, as the new `subschema-*` keys are not yet recognized.
- If a free-form config needs to define private fields that should be pruned on submission, use `Symbol` instead.
- Updated RequestCallout forms to reflect the changes above.

[KM-1304]: https://konghq.atlassian.net/browse/KM-1304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-plugins/src/components/free-form/RequestCallout/ConfigForm.vue
+++ b/packages/entities/entities-plugins/src/components/free-form/RequestCallout/ConfigForm.vue
@@ -148,7 +148,7 @@ function onChange(newVal?: RequestCalloutPlugin) {
   if (!newVal) return
 
   // replace callout `depends_on` ids with actual callout names
-  const pluginConfig = JSON.parse(JSON.stringify(newVal)) as RequestCalloutPlugin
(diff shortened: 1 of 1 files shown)


packages/entities/entities-plugins/src/components/free-form/RequestCallout/types.ts
```diff
@@ -88,8 +88,10 @@ export interface RedisClusterNode {
  *                    Callout                   *
  ************************************************/
 
+export const CalloutId: unique symbol = Symbol()
```
sumimakito, line 91: nit: Maybe we can also give this Symbol a description to allow a better debugging experience in the future

sumimakito (approved):
LGTM

PR #2173, opened 2025-06-09
fix: chart tooltip cutoff when dragged outside chart boundary

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-3663

- Teleport tooltip to body, adjust position relative to chart parent container.
- Prevent tooltip drag from selecting text on the page

### DIFF
--- a/packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
+++ b/packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
@@ -1,5 +1,6 @@
 <template>
   <div
+    ref="chartParentRef"
     class="chart-parent"
     :class="chartFlexClass(legendPosition)"
     data-testid="line-chart-parent"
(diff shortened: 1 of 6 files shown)


packages/analytics/analytics-chart/src/components/chart-types/TimeSeriesChart.vue
```diff
@@ -168,6 +171,7 @@ const legendItems = ref<EnhancedLegendItem[]>([])
 const tooltipElement = ref()
 const legendPosition = ref(inject('legendPosition', ChartLegendPosition.Right))
 const isDoingSelection = ref(false)
+const chartParentRef = ref<HTMLElement | null>(null)
```
filipgutica, line 174: Super weird... I cannot use `useTemplateRef` in this file. Soon as I do so, get this TS build error. Only thing I can think of is that it's some sort of conflict with `vue-chartjs` library this component in particular uses. 

For bar charts for example, which does not use `vue-chartjs` there is no issue.

```
src/components/chart-types/TimeSeriesChart.vue:884:1 - error TS4082: Default export of the module has or is using private name 'ExtendedDataPoint'.

884 export default (await import('vue')).defineComponent({
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
885 setup() {
    ~~~~~~~~~
... 
967 __typeRefs: {} as __VLS_TemplateResult['refs'],
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
968 });
    ~~~

Found 1 error in src/components/chart-types/TimeSeriesChart.vue:884
```

adorack (approved):

PR #2176, opened 2025-06-10
feat(tracing): migrate to traces v1 api [KHCP-16625]

vaibhavrajsingh2001:
# Summary

- migrate to Traces v1 API:
  - `attributes` are now simple objects instead of array of object. Now in the `attributes` object, key is attribute name and value is the actual value (no more `unwrapAnyValue` needed to get the actual value)
  - `Trace` now has a `resource_spans` field (which is an array of spans) instead of `batches`
    - refactored `mergeSpansInTraceBatches` to `mergeSpansInTrace` on account of the above removal
  - casing of fields has changed from camelCase to snake_case
  - all timestamps are strings now, so added handling for `string` → `BigInt` conversion before calculations
- add `@kong/sdk-konnect-js-internal` as a dev dependency for using types
- update fixture for playground

Jira ticket: https://konghq.atlassian.net/browse/KHCP-16625


packages/core/tracing/package.json
```diff
@@ -67,6 +68,7 @@
   },
   "peerDependencies": {
     "@kong/kongponents": "9.30.0",
+    "@kong/sdk-konnect-js-internal": "^0.6.0",
```
adamdehaven, line 71: question: why is this a peerDependency?

adamdehaven: It looks like you're just utilizing type imports from the package? If that's the case this only needs to be included in devDependencies as it will be compiled away at build time

vaibhavrajsingh2001: Yepp, just using it for types. Will remove it from `peerDependencies` when I open the PR on shared-ui. Will be closing this PR then.

PR #2180, opened 2025-06-12
feat(tracing): add prop for view span logs button [KHCP-16669]

vaibhavrajsingh2001:
# Summary

We need to add a prop to `TraceViewTab` to disable the "View span logs" button because the feature is behind a feature flag

## Preview

https://github.com/user-attachments/assets/7395077b-9574-4f66-8417-7611aff317c6


sumimakito (approved):
LGTM

PR #2188, opened 2025-06-17
feat: consistent formatting of value/units in tooltips [MA-4007]

Chalks:
[satisfies MA-4007](https://konghq.atlassian.net/browse/MA-4007)

* added a `formatUnit.ts` function and tests for it
* updated `commonOptions.ts` (used to populate tooltip data) to use the `formatUnit` function instead of calculating it inline
* updated `format-timestamps.ts` to be (imo) more clear
* updated `format-timestamps.spec.ts` to cover more test cases that became apparent when I refactored `format-timestamps.ts`
* updated `useLineChartOptions` to be that instead of `useLinechartOptions` because it was messing with my grep-foo (and because that's how `useBarChartOptions` is)

**How to test**

I'm genuinely not sure. The lowest value I could get with our dev data is in [this dashboard](https://cloud.konghq.tech/us/analytics/dashboards/983f725c-248d-42e4-8f36-c281a5fb8e97). The ticket says this:

> Sample prod org for verification: fbb54d39-ddf4-4dab-bdb6-978b8f334e4b (growth team org)

but I'm not sure how to use that to test.

So... if you have any good ideas, let me know. I tested this with temporary mockdata in dashboard renderer with analytics-chart linked to it. it... was a bit jank.

---

**Tooltip costs before**

*anything below 0.01*

Screenshot 2025-06-17 at 5 34 05 PM

*anything above 0.01*

Screenshot 2025-06-17 at 5 58 13 PM

**Tooltip costs after**

*sub 0.0001*

Screenshot 2025-06-17 at 5 55 24 PM

*sub 0.01*

Screenshot 2025-06-17 at 5 53 28 PM

*everything else*

Screenshot 2025-06-17 at 5 59 20 PM


packages/analytics/analytics-chart/src/utils/formatUnit.ts
```diff
@@ -0,0 +1,102 @@
+import composables from '../composables'
+import { numberFormatter } from './constants'
+
+export const formatBytes = (value: number, isBytes1024?: boolean): string => {
```
filipgutica, line 4: We're already using https://www.npmjs.com/package/pretty-bytes to format byte display. 

Did you want to discuss further why we should use something custom/in-house instead of that package? 

We should at least be consistent and either use pretty-bytes everywhere, or update the area where pretty-bytes is currently used to use your new function

filipgutica (changes requested):
Just wanted to discuss the solution further around byte display

PR #2191, opened 2025-06-21
chore: sentence case everything

filipgutica:
# Summary


packages/analytics/analytics-chart/src/locales/en.json
jordojordo, line 1: Should this also be changed?
```json
    "exportPreview": "Export Preview",
```

filipgutica: yup, done

jordojordo (approved):

PR #2208, opened 2025-07-10
feat(entities-plugins): datakit visual editor modal

Justineo:
# Summary

[KM-1453](https://konghq.atlassian.net/browse/KM-1453)
[KM-1455](https://konghq.atlassian.net/browse/KM-1455)

In this PR:

- Refactored `StandardLayout` to allow adding the editor mode toggle.
- Extracted the YAML code editor into a separate component.
- Added a toggle between code editor and visual editor, with the preference saved to `localStorage`
- Implemented basic functionality to open the visual editor modal from the plugin form
- Built the initial layout and interactions for the editor modal, including sidebar toggle and overall structure, scroll locking, focus trapping, etc.

[KM-1453]: https://konghq.atlassian.net/browse/KM-1453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KM-1455]: https://konghq.atlassian.net/browse/KM-1455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-plugins/package.json
+++ b/packages/entities/entities-plugins/package.json
@@ -91,6 +91,7 @@
     "@kong-ui-public/entities-routes": "workspace:^",
     "@kong-ui-public/entities-vaults": "workspace:^",
     "@kong-ui-public/forms": "workspace:^",
+    "focus-trap": "^7.6.0",
     "marked": "^14.1.3",
     "monaco-editor": "0.52.2"
   }
--- a/packages/entities/entities-plugins/package.json
+++ b/packages/entities/entities-plugins/package.json
@@ -30,6 +30,8 @@
     "@kong-ui-public/i18n": "workspace:^",
     "@kong/icons": "^1.31.0",
     "@kong/kongponents": "^9.34.2",
+    "@vueuse/core": "^13.5.0",
+    "@vueuse/integrations": "^13.5.0",
     "axios": "^1.7.7",
     "vue": ">= 3.3.13 < 4",
     "vue-router": "^4.4.5"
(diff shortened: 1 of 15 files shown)


sumimakito (approved):
Tested and LGTM. I left some non-blocking comments, as this is an internal feature for now.

**Some thoughts**

1. It seems we don't persist the code editor's content upon switching to the visual editor, shall we retain its state or show a confirmation modal before we switch away?
2. Since we previously made the host app import `editor.api` when importing monaco-editor, I'm thinking if we should add import calls to the necessary contrib deps in this package. Although we will import these dependencies on the host app side when it comes to the route form page (expressions editor), should it be better to import them here because we only have the Datakit plugin using an editor on the form page right now. (Maybe we should also consider moving the import calls from the host app's route form page to the entities-routes package, as it is nearer to the editor's usage)

sumimakito:
> Identifying all the dependent modules is a bit challenging—do you have any suggestions?

@Justineo As we usually use a similar set of features (formatting, suggest, etc.) from monaco-editor, I'm thinking of exposing a function to set up (import) the dependencies for that default set of features, and importing the specially required dependencies in the component. We may also make these optional features handy "use" functions. 

I agree that we can address this issue in the follow-up pull requests, and I'm also happy to help.

sumimakito (approved):
LGTM, thanks.

jw-foss (approved):

PR #2210, opened 2025-07-11
feat(fine-granularity): add response type for getting back information about available granularity

hfukada:
Used for new fine-granularity feature.

https://konghq.atlassian.net/browse/MA-4089


adorack (approved):
I think this makes sense.

filipgutica (approved):

PR #2218, opened 2025-07-16
fix: aged out data warning on dashboard tiles

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-3901

image

### DIFF
--- a/packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
+++ b/packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
@@ -39,7 +39,9 @@
                 :size="KUI_ICON_SIZE_20"
               />
             </KTooltip>
-            {{ badgeData }}
+            <span class="badge-text">
+              {{ badgeData }}
+            </span>
           </div>
         </KBadge>
         <EditIcon
--- a/packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
+++ b/packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
@@ -26,7 +26,10 @@
         class="tile-actions"
         :data-testid="`tile-actions-${tileId}`"
       >
-        <KBadge v-if="badgeData">
+        <KBadge
+          v-if="badgeData"
+          data-testid="time-range-badge"
+        >
           <div class="badge-content">
             <KTooltip
               v-if="isAgedOutQuery"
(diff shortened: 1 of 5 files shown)


packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -26,8 +26,23 @@
...
+            >
+              <WarningIcon
+                :size="KUI_ICON_SIZE_20"
+              />
+            </KTooltip>
+            <span class="badge-text">
+              {{ badgeData }}
+            </span>
+          </div>
         </KBadge>
```
Chalks, line 46: **opinion that I think is a blocker**

This functionality is built into KBadge and I think we should use tooltips the same way in every badge (e.g. tooltip hover target is the whole badge instead of just the icon):

```suggestion
        <KBadge
          v-if="badgeData"
          data-testid="time-range-badge"
          :tooltip="isAgedOutQuery ? agedOutWarning : undefined"
        >
          {{ badgeData }}
          <template
            v-if="isAgedOutQuery"
            #icon
          >
            <WarningIcon :size="KUI_ICON_SIZE_20" />
          </template>
        </KBadge>
```

packages/analytics/dashboard-renderer/src/locales/en.json
```diff
@@ -18,5 +18,19 @@
     "defaultFilename": "Chart export",
     "exportAsCsv": "Export as CSV"
   },
-  "jumpToExplore": "Explore"
+  "jumpToExplore": "Explore",
+  "query_aged_out_warning": "The saved query's granularity of \"{granularity}\" is no longer available because the data has aged out. Update with new granularity to ensure accurate reporting.",
+  "granularities": {
```
Chalks, line 23: **question, definitely not a blocker**

Is there any way we can get these strings from dashboard utilities or something? This seems a little fragile if/when we introduce more granularities. I guess that's not a huge concern as I doubt it happens often. shrug.jpg

Chalks (changes requested):

packages/analytics/dashboard-renderer/src/locales/en.json
Chalks: the more I think about it the less possible this seems. I'm more or less fine with this. I think it would be a huuuuge lift to centralize locale stuff more.

PR #2230, opened 2025-07-20
fix: handle zooming right to left ensure end > start

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-4109


packages/analytics/analytics-chart/src/components/chart-plugins/DragSelectPlugin.ts
```diff
@@ -17,12 +17,14 @@ const dispatchEvent = (eventName: string, chart: Chart, pluginInstance: DragSele
...
-    detail: {
-      xStart: xStartValue,
-      xEnd: xEndValue,
-    },
-  }))
+  if (xStartValue && xEndValue) {
+    chart.canvas.dispatchEvent(new CustomEvent<DragSelectEventDetail>(eventName, {
+      detail: {
+        xStart: xStartValue < xEndValue ? xStartValue : xEndValue,
+        xEnd: xEndValue > xStartValue ? xEndValue : xStartValue,
```
jordojordo, line 24: nit: Why not just use `Math.min` and `Math.max` here?

filipgutica: Good idea! 😅

PR #2235, opened 2025-07-21
fix(dashboard): update wording for aged out data warning

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-4119

image


packages/analytics/dashboard-renderer/src/locales/en.json
```diff
@@ -19,7 +19,7 @@
     "exportAsCsv": "Export as CSV"
   },
   "jumpToExplore": "Explore",
-  "query_aged_out_warning": "The saved query's granularity of \"{granularity}\" is no longer available because the data has aged out. Update with new granularity to ensure accurate reporting.",
+  "query_aged_out_warning": "Showing {currentGranularity} granularity ({savedGranularity} is no longer available). Update the tile to reflect this.",
```
adorack, line 22: Maybe "update the tile to hide this warning"?

Chalks: I argued Filip out of that wording originally, lol. My reasoning was that it was weird to have the language "to hide this warning" when the warning also had an X to dismiss it (which is true on the report page, not here).

At this point I would almost prefer it was just "Showing 1 hour granularity (1 minute is no longer available)" in all cases. I think everything we've come up with is more confusing and/or requires us to differentiate between viewers/admins.

jordojordo: But if the user doesn't have permission to update the tile, does this "call to action" make sense? There would need to be some "Contact your administrator" message instead.

PR #2241, opened 2025-07-22
fix: split the tooltip behaviour functions

filipgutica:
# Summary


packages/analytics/analytics-chart/src/utils/commonOptions.ts
```diff
@@ -24,21 +22,54 @@ export const tooltipBehavior = (tooltipData: TooltipState, context: ExternalTool
     const colors = tooltip.labelColors
     const valueAxis = context.chart.config?.options?.indexAxis === 'y' ? 'x' : 'y'
 
-    const isBarChart = ['horizontal_bar', 'vertical_bar'].includes(tooltipData.chartType)
+    tooltipData.tooltipContext = formatByGranularity(new Date(tooltip.dataPoints[0].parsed.x), granularity, true)
```
filipgutica, line 25: Just moving around some old code. 

lineChartTooltipBehaviour needs to know about granularity while bar charts do not

packages/analytics/analytics-chart/src/utils/commonOptions.ts
```diff
@@ -1,16 +1,14 @@
...
+import type { GranularityValues } from '@kong-ui-public/analytics-utilities'
 
 export const isTooltipInteractive = (state: TooltipInteractionMode) => {
   return ['interactive', 'zoom-interactive'].includes(state)
 }
 
-// TODO: we should implement a separate tooltip behavior for each broad chart type
-// as the "tooltip behaviors" are beggining to diverge more across chart types.
-export const tooltipBehavior = (tooltipData: TooltipState, context: ExternalTooltipContext) : void => {
+export const lineChartTooltipBehaviour = (tooltipData: TooltipState, context: ExternalTooltipContext, granularity: GranularityValues) : void => {
```
jordojordo, line 11: nit: This is a silly one, but we should be consistent with our locality when naming these things. `Behaviour` (UK spelling) vs. `Behavior` (US spelling).

Since `tooltipBehavior` is defined below, we should go with that spelling.

jordojordo (changes requested):

jordojordo (approved):

PR #2254, opened 2025-07-26
feat(entities-plugins): add engine prop to plugin form

2eha0:
# Summary

### DIFF
--- a/packages/entities/entities-plugins/src/components/PluginEntityForm.vue
+++ b/packages/entities/entities-plugins/src/components/PluginEntityForm.vue
@@ -240,6 +240,7 @@ const props = defineProps({
    */
   engine: {
     type: String,
+    required: false,
   },
 })
 
--- a/packages/entities/entities-plugins/src/components/PluginEntityForm.vue
+++ b/packages/entities/entities-plugins/src/components/PluginEntityForm.vue
@@ -882,7 +882,7 @@ watch(() => props.schema, (newSchema, oldSchema) => {
 }, { immediate: true, deep: true })
 
 onBeforeMount(() => {
-  form.value = parseSchema(props.schema, undefined, undefined, props.engine as any)
+  form.value = parseSchema(props.schema, undefined, undefined, props.engine)
 
   Object.assign(formModel, form.value?.model || {})
   formSchema.value = form.value?.schema || {}
--- a/packages/entities/entities-plugins/src/components/PluginEntityForm.vue
+++ b/packages/entities/entities-plugins/src/components/PluginEntityForm.vue
@@ -232,6 +232,15 @@ const props = defineProps({
     type: Boolean,
     default: false,
   },
+
+  /**
+   * Force the engine type for the form.
+   * @enum
+   * 'vfg' | 'freeform'
+   */
+  engine: {
+    type: String,
+  },
 })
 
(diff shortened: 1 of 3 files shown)


packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -232,6 +232,15 @@ const props = defineProps({
...
   },
+
+  /**
+   * Force the engine type for the form.
+   * @enum
+   * 'vfg' | 'freeform'
+   */
+  engine: {
+    type: String,
+  },
```
jw-foss, line 243: **Question:** should we add `required: false` here?

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -852,7 +861,7 @@ watch(() => props.schema, (newSchema, oldSchema) => {
   if (objectsAreEqual(newSchema || {}, oldSchema || {})) {
     return
   }
-  const form: Record<string, any> = parseSchema(newSchema)
+  const form: Record<string, any> = parseSchema(newSchema, undefined, undefined, props.engine as any)
```
jw-foss, line 864: Not sure why we are casting it to any, maybe we should fix the typing in `parseSchema` instead?

2eha0: Because the type of `props.engine` is `string | undefined`

packages/entities/entities-plugins/src/composables/useSchemas.ts
```diff
@@ -315,7 +315,7 @@ export const useSchemas = (options?: UseSchemasOptions) => {
     formSchema._supported_redis_partial_type = currentSchema._supported_redis_partial_type
     formSchema._redis_partial_path = currentSchema._redis_partial_path
 
-    if (getSharedFormName(pluginName) || getFreeFormName(pluginName, experimentalFreeForms) || metadata?.useLegacyForm || options?.credential) {
+    if (getSharedFormName(pluginName) || getFreeFormName(pluginName, experimentalFreeForms) || metadata?.useLegacyForm || options?.credential || engine === 'freeform') {
```
hangrao, line 318: Will we remove experimentalFreeForms in future, it looks similar with "engine === 'freeform'"

2eha0: Definitely will

packages/entities/entities-plugins/src/components/PluginEntityForm.vue
```diff
@@ -863,14 +871,18 @@ watch(() => props.schema, (newSchema, oldSchema) => {
...
+      ? undefined
+      : getFreeFormName(form.model.name, experimentalFreeForms) || 'CommonForm'
   sharedFormName.value = getSharedFormName(form.model.name)
 
   initFormModel()
 }, { immediate: true, deep: true })
 
 onBeforeMount(() => {
-  form.value = parseSchema(props.schema)
+  form.value = parseSchema(props.schema, undefined, undefined, props.engine as any)
```
sumimakito, line 885: Would this work without `any` now?

```suggestion
  form.value = parseSchema(props.schema, undefined, undefined, props.engine)
```

sumimakito (commented):
Others LGTM!

PR #2259, opened 2025-07-29
feat(entities-plugins): datakit store, improve types

Justineo:
# Summary

[KM-1518](https://konghq.atlassian.net/browse/KM-1518)

### Add global editor store for Datakit

- Type updates: `UserNode` → `ConfigNode`, `NodeData` → `NodeInstance`
- Added edge-related types and state
- Provide global store via `provideEditorState` / `useEditorState`
- Expose core state: `nodes`, `edges`, lookup maps
- Support node/field/edge operations
- Undo/redo with tagged history
- Built-in validation: structure, cycles, phase rules
- VueFlow adapters and connection validation

[KM-1518]: https://konghq.atlassian.net/browse/KM-1518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/store/store.ts
+++ b/packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/store/store.ts
@@ -15,14 +15,19 @@ import type {
   NodeName,
   UINode,
 } from '../../types'
-import { createId, deepClone, findFieldById } from './helpers'
+import {
+  createId,
+  deepClone,
+  findFieldById,
+  generateNodeName,
+} from './helpers'
 import { isImplicitName, isImplicitType } from '../node/node'
 import { initEditorState, makeNodeInstance } from './init'
 import { useValidators } from './validation'
 import { useTaggedHistory } from './history'
 
 export const [provideEditorState, useEditorState] = createInjectionState(
-  function(configNodes: ConfigNode[], uiNodes: UINode[]) {
+  function createState(configNodes: ConfigNode[], uiNodes: UINode[]) {
     const state = ref<EditorState>(initEditorState(configNodes, uiNodes))
     const selection = ref<NodeId>()
     const history = useTaggedHistory(state, {
@@ -48,6 +53,11 @@ export const [provideEditorState, useEditorState] = createInjectionState(
         new Map<EdgeId, EdgeInstance>(
           state.value.edges.map((edge) => [edge.id, edge]),
         ),
+    )
+
+    // sets
+    const nodeNames = computed(
+      () => new Set(state.value.nodes.map((node) => node.name)),
     )
 
     // validators bound to current maps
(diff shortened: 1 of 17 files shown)


packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/store/store.ts
```diff
@@ -0,0 +1,545 @@
...
+  UINode,
+} from '../../types'
+import { createId, deepClone, findFieldById } from './helpers'
+import { isImplicitName, isImplicitType } from '../node/node'
+import { initEditorState, makeNodeInstance } from './init'
+import { useValidators } from './validation'
+import { useTaggedHistory } from './history'
+
+export const [provideEditorState, useEditorState] = createInjectionState(
+  function(configNodes: ConfigNode[], uiNodes: UINode[]) {
```
2eha0, line 25: nit: we can give a function name for a better error stack

packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/store/store.ts
```diff
@@ -0,0 +1,545 @@
...
+      },
+      commitNow = true,
+    ) {
+      if (isImplicitType(payload.type)) {
+        console.warn('[addNode] implicit nodes are fixed.')
+        return
+      }
+      const node = makeNodeInstance({
+        type: payload.type,
+        name: payload.name,
```
2eha0, line 99: We may need a helper function to create a unique name, something like:
```ts
name: generateUniqueName(payload.name),
```

2eha0: 1. Duplicate names will break the `nodeMapByName`.
2. Users might get validation error when saving, unless they set the names manually to ensure uniqueness.

sumimakito: Yes, it's nice to have a function like `generateUniqueName`, but would it be better to move the usage out of `addNode`, which respects the input payload and only checks for name duplication and fails if conflicted?

Justineo: > if this `addNode` function may also be used for re-constructing nodes from a config

`addNode` and `buildNodeInstance`(out of saved config/ui data) are currently sharing a common `makeNodeInstance` function under the hood.

Should we automatically name the newly created node as something like `${type}_${countByType}`? eg. `JQ_3`, `CALL_1`, etc.

sumimakito: I see. This sounds good to me

Justineo: Added a `generateNodeName` function which will be called inside `addNode` for non-implicit node types.

PR #2261, opened 2025-07-29
fix(querydataprovider): handle 403 and 408 errors manually [MA-4096]

jordojordo:
# Summary
Fix [MA-4096](https://konghq.atlassian.net/browse/MA-4096)

This will update the QueryDataProvider to properly catch and handle 403 and 408 status codes to display a message rather than causing a global error to be displayed.

[MA-4096]: https://konghq.atlassian.net/browse/MA-4096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics/dashboard-renderer/src/components/QueryDataProvider.vue
+++ b/packages/analytics/dashboard-renderer/src/components/QueryDataProvider.vue
@@ -154,7 +154,7 @@ const { data: v4Data, error, isValidating } = useSWRV(queryKey, async () => {
       errorMessage.value = e?.response?.data?.message || e?.message
     }
 
-    return errorMessage.value
+    throw e
   } finally {
     emit('queryComplete')
   }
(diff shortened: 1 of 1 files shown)


packages/analytics/dashboard-renderer/src/components/QueryDataProvider.vue
adorack, line 1: I think it's worth figuring out why these messages aren't showing up.  It looks like the empty state's message prop isn't getting filled in, maybe?

adorack (changes requested):

packages/analytics/dashboard-renderer/src/components/QueryDataProvider.vue
jordojordo: This was a silly issue. We were not awaiting the call to `queryBridge.queryFn()` which would not settle the promise and subsequently the `catch` would never run.

PR #2275, opened 2025-07-31
feat(dashboards): disable zoom if event handler not present [MA-4159]

adorack:
Dashboards aren't in charge of the time period they're querying, so unless the host app provides a zoom event handler, there's no point in allowing zoom.

# Summary

### DIFF
--- a/packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
+++ b/packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
@@ -42,15 +42,23 @@
 
 <script setup lang="ts">
 import type { DashboardRendererContext, DashboardRendererContextInternal, GridTile } from '../types'
-import type { AbsoluteTimeRangeV4, DashboardConfig, TileConfig, SlottableOptions, TileDefinition, AllFilters } from '@kong-ui-public/analytics-utilities'
+import type {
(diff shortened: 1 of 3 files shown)


filipgutica (dismissed):
Just one comment

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -185,12 +193,17 @@ const mergedContext = computed<DashboardRendererContextInternal>(() => {
     editable = false
   }
 
+  // Check if the host app has provided an event handler for zooming.
+  // If there's no handler, disable zooming -- it won't do anything.
+  const zoomable = !!getCurrentInstance()?.vnode?.props?.onZoomTimeRange
```
filipgutica, line 198: isn't the emit named `zoomTimeRange` ?

filipgutica (commented):
Just had a question

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
filipgutica: Heh, interesting, good to know!

filipgutica (approved):

PR #2280, opened 2025-08-04
chore(entities-shared): add supported entity type for event gateway

TT1228:
# Summary

Add supported entity type for event gateway


nekolab (approved):
LGTM

PR #2283, opened 2025-08-05
feat(*): filter out non-redis partials in redis partials [KM-1546]

2eha0:
# Summary

KM-1546


TT1228 (approved):
LGTM. Are there any other places we need to filter out other partials?

PR #2289, opened 2025-08-06
fix(entities-upstreams-targets): enable header in active health check 

hangrao:
[KM-1555,](https://konghq.atlassian.net/browse/KM-1555) enable header in active health check for Konnect.

Tested on Konnect, header can be displayed.
image

JIRA: KM-1555

# Summary


packages/entities/entities-upstreams-targets/src/components/UpstreamsFormActiveHealthCheck.cy.ts
```diff
@@ -280,7 +280,7 @@ describe('<UpstreamsFormActiveHealthCheck/>', { viewportHeight: 700, viewportWid
     })
   })
 
-  // TODO: remove when api will support `headers` property
+  // Konnect supports `headers` property now
```
Leopoldthecoder, line 283: I think we can merge this test with the following one:
1. Remove line 283 to line 295
2. Change line 296 to `it('Should bind headers data correctly', () => {`
3. Remove line 301
4. Then remove line 2

hangrao:
> The changes look good to me, but it seems the Konnect API still rejects the payload with `healthchecks.active.headers` while I was verifying this on the Gateway Manager. Let's hold off on merging this until we receive confirmation from the Koko team.

Thanks, resolved by using v2 validate endpoints.

Leopoldthecoder (approved):

PR #2292, opened 2025-08-06
fix(analytics): scope tile explore filters to datasource [MA-4174]

jordojordo:
# Summary

Fix [MA-4174](https://konghq.atlassian.net/browse/MA-4174)

This includes changes to the analytics-utilities and dashboard-renderer packages to allow for proper scoping of a tile's filters when the Jump to Explore action is used.

Now that we are adding multiple datasource filtering to tiles within a dashboard, we need to also ensure that the filters applied to the query of an explore link only includes the filters that apply to the tile's given datasource.

This will add a computed property within the DashboardTile component to derive the subset necessary for the Jump to Explore action. There is now a `getFieldDataSources` method to return the possible datasources for a filter's `field` property, based on this array we can determine which filters to apply to the query.

[MA-4174]: https://konghq.atlassian.net/browse/MA-4174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


adorack (changes requested):
I'd like to review.

packages/analytics/analytics-utilities/src/filters.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+  if ((filterableAiExploreDimensions as string[]).includes(dimension)) {
+    datasources.add('llm_usage')
+  }
+  if ((filterableRequestDimensions as string[]).includes(dimension) || (filterableRequestMetrics as string[]).includes(dimension) || (filterableRequestDimensions as string[]).includes(dimension)) {
+    datasources.add('requests')
+  }
+
+  return Array.from(datasources)
+}
```
Chalks, line 28: **question/preference - probably not a blocker**

This is a copy [of this function](https://github.com/Kong/konnect-ui-apps/blob/main/apps/analytics/src/utils/Filters.ts#L639-L656). Do you intend to go back and replace usage of `getFieldDatasources` in konnect with this one instead? I definitely prefer having just one implementation of this.

Chalks: discussed in chat

packages/analytics/analytics-utilities/src/filters.ts
```diff
@@ -0,0 +1,28 @@
...
+  filterableAiExploreDimensions,
+  filterableBasicExploreDimensions,
+  filterableExploreDimensions,
+  filterableRequestDimensions,
+  filterableRequestMetrics,
+} from './types'
+
+
+export const getFieldDataSources = (dimension: AllFilterableDimensionsAndMetrics): FilterDatasource[] => {
+  const datasources = new Set<FilterDatasource>()
```
adorack, line 12: Minor: this doesn't need to be a set.

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -301,6 +301,21 @@ const agedOutWarning = computed(() => {
...
+ *
+ * @returns Array of scoped filter objects to a datasource
+ */
+const datasourceScopedFilters = computed(() => {
+  const filters = [...props.context.filters, ...props.definition.query.filters ?? []]
+
+  return filters.filter(f => {
+    const possibleSources = getFieldDataSources(f.field)
+
+    return possibleSources.some((ds: FilterDatasource) => props.definition?.query?.datasource === ds)
```
adorack, line 315: It's not a safe assumption that `props.definition.query.datasource` is defined; if it's not, this will result in never including any filters.  (You can see this in some of the contextual previews in your version bump PR.)

I'd like to make it required, but it's some followup work that we haven't tackled yet -- https://konghq.atlassian.net/browse/MA-2987.

For now, I think you can assume the datasource defaults to `api_usage`.  I'll update MA-2987.

adorack (changes requested):

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
adorack, line 1: Ugh.  I'm sorry that I missed this in the earlier review, but it looks like we have some test coverage for the Explore link.  Could you please add a test to `DashboardTile.cy.ts` that passes the tile an irrelevant filter from the context and then makes sure the irrelevant dimension doesn't show up in the URL?  (See line 132 for where the jump to explore tests live.)

adorack (changes requested):

adorack (approved):

PR #2305, opened 2025-08-07
chore(documenation, plugins): clean up l2 empty state flag

afrzhussain:
**Description:**
This PR is dedicated to cleaning up the feature flag `khcp-14756-empty-states-m2`.

**Entities Cleaned Up:**
* **Documentation Content**
* **Plugins List**

### DIFF
--- a/packages/core/documentation/src/components/DocumentationContent.vue
+++ b/packages/core/documentation/src/components/DocumentationContent.vue
@@ -3,7 +3,6 @@
     <div v-if="documentList && !documentList.length">
       <KCard v-if="emptyStateCard">
         <EntityEmptyState
-          v-if="enableV2EmptyStates"
           :action-button-text="t('documentation.show.empty_state_v2.cta')"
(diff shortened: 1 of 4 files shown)


packages/core/documentation/src/components/DocumentationContent.vue
```diff
@@ -21,14 +20,9 @@
...
         </EntityEmptyState>
-        <DocumentationPageEmptyState
-          v-else
-          :can-edit="canEdit"
-          @create-documentation="handleAddClick"
-        />
       </KCard>
       <EntityEmptyState
-        v-else-if="enableV2EmptyStates"
+        v-else
```
mptap, line 25: @afrzhussain : seems this `v-else` shouldn't be here as the original condition was `v-else-if="enableV2EmptyStates"` and we're removing the `enableV2EmptyStates`

mptap (approved):
Lgtm 🚀

PR #2309, opened 2025-08-08
fix(forms): xss vulnerability in plugin schema

Leopoldthecoder:
# Summary

1. For field tooltip, sanitize the help string before passing it to `v-html`.
2. For field label/hint/error, use text Interpolation.

KM-1563


jw-foss (dismissed):

jw-foss (approved):
LGTM

PR #2316, opened 2025-08-11
feat(entities-upstreams-targets): add new field `failover`

jw-foss:
# Summary

- Add new field `failover` for upstreams targets.
- Introduce a flag for controlling it.

CleanShot 2025-08-11 at 15 23 07@2x

CleanShot 2025-08-11 at 17 32 14@2x

CleanShot 2025-08-11 at 17 31 28@2x

CleanShot 2025-08-12 at 11 38 42@2x

[KM-1573]

[KM-1573]: https://konghq.atlassian.net/browse/KM-1573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-upstreams-targets/docs/target-form.md
+++ b/packages/entities/entities-upstreams-targets/docs/target-form.md
@@ -100,7 +100,7 @@ Controls whether the failover feature is enabled for the target.
 - required: `false`
 - default: `false`
 
-Differs from `failoverEnabled`, when `failover` is `enabled`, this indicates whether the
+When `failoverEnabled` is `true`, this indicates whether the
 current upstream load balancing target supports `failover` or not.
 
 ### Slots
--- a/packages/entities/entities-upstreams-targets/src/locales/en.json
+++ b/packages/entities/entities-upstreams-targets/src/locales/en.json
@@ -372,7 +372,7 @@
           "label": "Failover Target",
           "load_balance_target": "Load balance",
           "failover_target": "Failover",
-          "help": "Whether to use this target only as backup or not.",
+          "help": "Whether to use this target only as backup.",
           "unsupported": "The current load balance algorithm does not support failover."
         }
       }


packages/entities/entities-upstreams-targets/src/components/TargetsList.vue
```diff
@@ -40,6 +40,19 @@
...
+
+      <template
+        v-if="failoverEnabled"
+        #failover="{ rowValue }"
+      >
+        <KBadge
+          appearance="info"
+          class="upstream-failover"
+        >
+          {{ rowValue ? 'Yes' : 'No' }}
```
jw-foss, line 52: Was considering if we should adding this into `en.json` or maybe simplify this with `Y` and `N`

packages/entities/entities-upstreams-targets/src/components/TargetForm.vue
```diff
@@ -104,6 +118,8 @@ const emit = defineEmits<{
   (e: 'cancel'): void
 }>()
 
+const failoverEnabled = inject(FAILOVER_INJECTION_KEY, true)
```
Leopoldthecoder, line 121: Are we using injection instead of a prop because the `TargetForm` component is not intended to be used standalone in the consuming app?

I think we could still make this a prop in both `TargetList` and `TargetForm`. In `TargetList`, the prop value would be determined by the consuming app, while in `TargetForm` it could be set [here](https://github.com/Kong/public-ui-components/blob/380f06f74148c26795da5302d3a24f58a8897186/packages/entities/entities-upstreams-targets/src/components/TargetsList.vue#L118) directly.

Keeping it as a prop and documenting it in the README would be more consistent with how we handle other switches.

packages/entities/entities-upstreams-targets/src/components/TargetsList.vue
Leopoldthecoder: Maybe we can rename this column to something like `Target type`, and the values can be `Failover target` and `Load balance target`?

jw-foss:
@Leopoldthecoder all comments resolved, PTAL again 🙇

PR #2341, opened 2025-08-15
feat(entities-plugins): datakit readonly mode for flow canvas

sumimakito:
## Features 🆕 

<details>
<summary>
Show readonly flows for previewing in the form page
</summary>
Screenshot 2025-08-21 at 18 24 20
Screenshot 2025-08-21 at 18 24 30
</details>

## Tweaks ⚙️ 

<details>
<summary>
Auto layout now prefers to add spaces between config nodes and implicit nodes to fill the viewport
</summary>

|Before|After|
|:-:|:-:|
|Screenshot 2025-08-20 at 03 35 52|Screenshot 2025-08-20 at 03 35 18|

</details>

<details>
<summary>
Make panel divider 1px high
</summary>
Screenshot 2025-08-21 at 18 26 48
</details>

KM-1604

### DIFF
--- a/packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/composables/useConflictConnectionConfirm.ts
+++ b/packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/composables/useConflictConnectionConfirm.ts
@@ -8,7 +8,7 @@ export function provideConfirmModal(open: OpenConfirm) {
 
 // For scenarios where the confirm modal is not needed
 export function useOptionalConfirm() {
-  return inject<OpenConfirm>(CONFIRM_MODAL_PROVIDE_KEY)
+  return inject<OpenConfirm | undefined>(CONFIRM_MODAL_PROVIDE_KEY, undefined)
 }
 
 export function useConfirm() {


packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/FlowCanvas.vue
```diff
@@ -37,36 +44,48 @@
...
+const { flowId, phase, editing = true } = defineProps<{
+  flowId: string
   phase: NodePhase
+
+  /**
+   * Whether the flow canvas is editable.
+   *
+   * @default true
+   */
+  editing?: boolean
```
Justineo, line 73: Shall we name it to `readonly` so that it can be default to `false`?

packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/FlowViewer.vue
Justineo, line 1: I think this should still be `FlowEditor` as it is by default an editor in a (partially) readonly state (you can still edit the `debug` config).

packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/FlowViewer.vue
```diff
@@ -0,0 +1,101 @@
...
+
+const flowPanels = useTemplateRef('flowPanels')
+
+function onChange(configNodes: ConfigNode[], uiNodes: UINode[]) {
+  emit('change',
+    {
+      ...config,
+      nodes: {
+        ...config?.nodes,
+        ...configNodes,
```
Justineo, line 61: q: why do we merge new nodes with previous nodes? (and Copilot is right, they are arrays)

sumimakito: I think this should be a mistake from the copy-paste of the previous code in `<FlowEditor />`. Let me fix it

packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/store/init.ts
```diff
@@ -27,6 +27,7 @@ import {
 export function initEditorState(
   configNodes: ConfigNode[],
   uiNodes: UINode[],
+  creationMode = false,
```
Justineo, line 30: Shall we stick with `isEditing` for consistency? Using different names for the same state (`isEditing` vs. `creationMode`) could add unnecessary friction.

packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/FlowPanels.vue
```diff
@@ -0,0 +1,321 @@
...
+    } else if (responseCanvasContainerRect.height.value < snappingHeight.value) {
+      requestHeight.value = '100%'
+      responseHeight.value = '0'
+    }
+  },
+})
+
+// 2-pass: Fit view and sync the zoom
+const fitView = () => {
+  // Hack. Next tick does not work as expected here
```
Justineo, line 175: What issues are you running into? I did some research while building a PoC and might be able to share my solution.

packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/store/init.ts
sumimakito: Yes. Let me update this

packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/node/FlowNode.vue
```diff
@@ -208,6 +209,7 @@ const { data } = defineProps<{
 
 const { t } = createI18n<typeof english>('en-us', english)
 
+const flowStore = useOptionalFlowStore()
```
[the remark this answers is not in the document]
sumimakito: This was because we call `provideFlowStore` in `<FlowCanvas />`, but we don't have that for `<FlowNode />`'s usage in `<NodePanel />`, `<EditorContent />`, and then `<EditorModal />`, which is the path used for DND previews. 

In the latter case, we neither provide nor need the store. I feel it should be okay to make the store getter optional here. WDYT?

packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/FlowPanels.vue
sumimakito: I previously found that although the two flows were fitted to view on init, upon a click on the "Fit View" button, the viewport moved slightly, which indicates the layout computation is not yet finalised by the point when the initial fit view happened.

However, since we no longer need to sync the zoom level between both flows. A simple `nextTick` worked here again.

Justineo (approved):
LGTM

PR #2354, opened 2025-08-19
fix(entities-plugins): request callout - KV default null [KM-1580]

2eha0:
# Summary

KM-1580


packages/entities/entities-plugins/src/components/free-form/shared/headless/useKeyValueField.ts
```diff
@@ -21,7 +21,7 @@ export interface KeyValueFieldProps<TKey extends string = string, TValue extends
 }
 
 export interface KeyValueFieldEmits<TKey extends string = string, TValue extends string = string> {
-  change: [Record<TKey, TValue>]
+  change: [Record<TKey, TValue> | null]
```
jw-foss, line 24: **Nit:** maybe we can make `Record<TKey, TValue> | null` a new type.

```ts
type KeyValueType = Record<TKey, TValue> | null
```

PR #2362, opened 2025-08-20
fix(analytics-chart): chart tooltip improvements

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-4133 

- use position:fixed rather than teleporting to body
- when computing the tooltip's absolute position use `useElementBounding()` to get reactive bounding rect of the parent. This allows offsetting if any parent is scrolled not just the document body.
- use `useDraggable` from vueuse in favour of custom drag handling - simplifies the code
- sync the drag position with the tooltip's absolute position when not dragging - this is what accounts for scroll offset and also allows the dragging state to start from the tooltip's last position,

https://github.com/user-attachments/assets/873fd890-afed-428c-bfe0-0b02c6da47cd

### DIFF
--- a/packages/analytics/analytics-chart/src/components/chart-plugins/ChartTooltip.vue
+++ b/packages/analytics/analytics-chart/src/components/chart-plugins/ChartTooltip.vue
@@ -5,12 +5,11 @@
     class="tooltip-container"
     :class="{ 'locked': state.interactionMode === 'interactive' }"
     :style="{
-      transform: isInteractive ? `translate(${dragPosition.left}, ${dragPosition.top})` : `translate(${absoluteLeft}, ${absoluteTop})`,
+      transform: isInteractive ? `translate(${dragLeft}, ${dragTop})` : `translate(${absoluteLeft}, ${absoluteTop})`,
       left: '0',
       top: '0',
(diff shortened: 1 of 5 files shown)


packages/analytics/analytics-chart/src/components/chart-plugins/ChartTooltip.vue
```diff
@@ -214,7 +150,7 @@ function handleMouseUp() {
   border-radius: var(--kui-border-radius-20, $kui-border-radius-20);
   box-shadow: 0 5px 15px rgba(0, 0, 0, 0.12), 0 5px 10px rgba(0, 0, 0, 0.24);
   max-width: 425px;
-  position: absolute;
+  position: fixed;
   transition: all 0.2s cubic-bezier(0.25, 0.8, 0.25, 1);
   z-index: 999;
```
Chalks, line 155: **probably a blocker**

With the change to fixed, I think the z-index needs changed to `1`.

**With z-index 999**

Screenshot 2025-08-21 at 10 05 57 AM

**With z-index 1**

Screenshot 2025-08-21 at 10 06 01 AM

filipgutica: good call

PR #2368, opened 2025-08-21
fix(dashboardrenderer): disable jump to request for llm datasource [MA-4246]

jordojordo:
# Summary

Fix [MA-4246](https://konghq.atlassian.net/browse/MA-4246)

This will disable the Jump to Requests action when zooming or selecting the action dropdown menu for a DashboardTile when the datasource is `llm_usage`. 

Added a computed property to clean up other sections of the component that use the same conditions.

[MA-4246]: https://konghq.atlassian.net/browse/MA-4246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
+++ b/packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
@@ -120,6 +120,7 @@ interface ChartProps {
   showAnnotations?: boolean
   timeseriesZoom?: boolean
   requestsLink?: ExternalLink
+  queryDatasource?: string
 }
 
 const emit = defineEmits<{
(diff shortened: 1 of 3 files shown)


packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
adorack, line 1: Is `AnalyticsChart` the right place to make a call on showing / hiding the link based on a datasource?  I'm kind of inclined to say no; we've pretty carefully avoided needing to give `AnalyticsChart` any information about the query that generated the data.

Maybe the component in the host app should decide whether to set `requestsLink`?  We tend to use `AnalyticsChart` as part of dashboards except in very special circumstances, so I feel like this would be fine.

adorack (changes requested):
Happy to discuss further on Slack if desired.

PR #2370, opened 2025-08-22
feat(entities-plugins): freeform - support array of array [KM-1521]

2eha0:
# Summary

KM-1521

image


packages/entities/entities-plugins/src/components/free-form/shared/ArrayField.vue
```diff
@@ -150,6 +155,8 @@ import { useField, useFieldAttrs, useFormShared, useItemKeys } from './composabl
 import useI18n from '../../../composables/useI18n'
 import * as utils from './utils'
 import Field from './Field.vue'
+import type { ArrayLikeFieldSchema } from '../../../types/plugins/form-schema'
+import TagField from './TagField.vue'
```
Leopoldthecoder, line 159: Non-blocking: since `TagField` is no longer used exclusively for tags, we could consider renaming this component in the future. Other field components follow the `[ValueType]Field` naming pattern, so perhaps `StringArrayField` would be appropriate.

packages/entities/entities-plugins/src/components/free-form/test/utils.ts
```diff
@@ -320,6 +320,27 @@ export function assertFormRendering(schema: FormSchema, options?: {
       })
   }
 
+  const assertStringArrayOfArrayField = ({
+    fieldKey,
+  }: AssertFieldOption<ArrayFieldSchema>) => {
+    debugger
```
Leopoldthecoder, line 326: Shall we remove this line?

2eha0, line 326:
```suggestion
```

Leopoldthecoder (approved):
LGTM. Thanks!

PR #2389, opened 2025-08-25
fix(entities-plugins): datakit auto layout conditions

sumimakito:
* Add a `needLayout` flag to `EditorState` to detect the possible out-of-sync between the layout (UI data) and the configurations
* Remove `isEditing` from `EditorState` as `needLayout` should be sufficient to cover the use case
* Leave some TODOs for future


packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/store/init.ts
```diff
@@ -141,7 +153,9 @@ export function initEditorState(
   // Mark all nodes that should be in 'request' phase
   markRequestNodes(['request', 'service_request'])
 
-  return { nodes, edges, isEditing }
+  // TODO(Makito): Should we clean up stale nodes from the UI data in case the name of a deleted node is reused in the future?
```
Justineo, line 156: I don't think we should "remember" the UI config of a deleted node. Once it's deleted in the config, we should delete everything related.

Justineo (approved):
LGTM

packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/store/init.ts
sumimakito: Agreed. Let's remove these dangling nodes from the UI data if they do not have corresponding ones in the config on init in the future.

PR #2390, opened 2025-08-25
fix(entities-plugins): datakit disconnect before update edges

sumimakito:
Fix the issue where the old edge is left untouched in some conditions while updating the edge


Justineo (dismissed):
LGTM

Justineo (approved):
LGTM

PR #2394, opened 2025-08-25
fix: use start and end from chart data for requests link

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-4219

ui-apps pr: https://github.com/Kong/konnect-ui-apps/pull/8086

In order to reflect the effective query time range, use `start_ms` and `end_ms` returned in the explore result meta when generating the request link for a dashboard tile.


adorack (changes requested):
Could you please add a test for this?

adorack (approved):

PR #2401, opened 2025-08-26
feat(entities-plugins): add 312 new plugin icon and metadata

hangrao:
# Summary
image

add icons and metadata for 3.12 new plugins, some of the new plugins schema are not submitted.


jw-foss (approved):

sumimakito (approved):
nit: Do we need to sort the keys in the i18n JSON?

PR #2402, opened 2025-08-26
fix(*): replace and deprecate entity empty state [KHCP-17718]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-17265

Replace EntityEmptyState usage with KEmptyState and deprecate it

### DIFF
--- a/packages/entities/entities-keys/src/components/KeyList.vue
+++ b/packages/entities/entities-keys/src/components/KeyList.vue
@@ -67,7 +67,7 @@
           :message="t('keys.list.empty_state_v2.description')"
           :title="t('keys.list.empty_state_v2.title')"
         >
-          <template #image>
+          <template #icon>
             <KeyIcon decorative />
           </template>
 


packages/entities/entities-key-sets/src/components/KeySetList.vue
```diff
@@ -61,33 +61,43 @@
...
           data-testid="key-sets-entity-empty-state"
-          :description="t('keySets.list.empty_state_v2.description')"
-          :learn-more="config.app === 'konnect'"
+          icon-background
+          :message="t('keySets.list.empty_state_v2.description')"
           :title="t('keySets.list.empty_state_v2.title')"
-          @click:create="handleCreate"
-          @click:learn-more="$emit('click:learn-more')"
         >
           <template #image>
```
Leopoldthecoder, line 70: We need to change this to the icon slot:
```suggestion
          <template #icon>
```

Otherwise the icon is aligned to the left:
image

portikM: Ah good catch, thanks!

kaiarrowood (approved):
LGTM! 🚀

Leopoldthecoder (approved):
LGTM. Thanks!

PR #2405, opened 2025-08-27
fix(entities-plugins): datakit expand handles on i/o fields update

sumimakito:
* Expand handles on I/O fields update (add, rename)
* Collapse handles when all I/O fields are removed
* Fix a `z-index` issue (the removal button is hard to interact with)

https://github.com/user-attachments/assets/86ee3311-50a9-471d-afef-0b04e51014eb

KM-1655


2eha0 (approved):
LGTM

PR #2408, opened 2025-08-27
fix(entities-plugins): datakit cannot clear ui data by null

sumimakito:
* Remove `null` from the parameter type on `handleConfigChange` as setting `__ui_data` to `null` does not clear the `__ui_data` in the entity record, which is by design of the API.
* Make `newUIData` handleFlowChange non-optional because it won't be `undefined`.


Justineo (commented):
LGTM

Justineo (approved):

PR #2409, opened 2025-08-27
refactor(entities-plugins): datakit - data flow

2eha0:
# Summary

### Before
The `flowEditor` and the `codeEditor` shared the local states from `DatakitForm`, then sync the local states to freeform, then freeform will send the data to the outside.
image

### After
The `flowEditor` and the `codeEditor` shared form data from freeform directly
image


Justineo (approved):
LGTM

PR #2411, opened 2025-08-27
feat: add explore link to zoom actions

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-4272

ui-apps pr: https://github.com/Kong/konnect-ui-apps/pull/8128

### Changes
- Add exploreLink prop to analytics chart and add explore zoom action item
- DashboardTile 
  - moved link generating logic to composable.
  - generate explore link on sele
- add unit test for new composable

https://github.com/user-attachments/assets/27fc6c41-beb4-4304-94c3-a326bb761c77

ct chart range event

### DIFF
--- a/packages/analytics/dashboard-renderer/src/components/DashboardTile.cy.ts
+++ b/packages/analytics/dashboard-renderer/src/components/DashboardTile.cy.ts
@@ -329,4 +329,68 @@ describe('<DashboardTile />', () => {
     // not the time range in the query definition.
     cy.getTestId('chart-jump-to-requests-1').invoke('attr', 'href').should('have.string', `"start":${start},"end":${end}`)
   })
+
+  it('should not show explore if datasource is unsupported', () => {
+    mount({
+      definition: {
+        chart: mockTileDefinition.chart,
+        query: {
+          datasource: 'random_unsupported_source',
+          metrics: [],
+          filters: [],
+        },
+      },
+    })
+
+    cy.getTestId('kebab-action-menu-1').click()
+    cy.getTestId('chart-jump-to-explore-1').should('not.exist')
+  })
+
+  it('should not show explore if exploreBaseUrl is undefined', () => {
+    const mockQueryProviderWithEmptyExplore = {
+      ...mockQueryProvider,
+      exploreBaseUrl: async () => undefined,
+    }
+    cy.mount(DashboardTile, {
+      props: {
+        definition: mockTileDefinition,
+        context: mockContext,
+        queryReady: true,
+        refreshCounter: 0,
+        tileId: '1',
+      },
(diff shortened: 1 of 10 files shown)


packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -190,15 +190,23 @@ const chartData = ref<ExploreResultV4>()
...
+  canGenerateRequestsLink,
+  buildExploreQuery,
+  buildExploreLink,
+  buildRequestLink,
+  buildRequestsQueryZoomActions,
+} = composables.useContextLinks({
+  queryBridge,
+  chartData: computed(() => chartData.value),
+  definition: computed(() => props.definition),
+  context: computed(() => props.context),
```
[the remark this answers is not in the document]
filipgutica: for the props it's better than using toRef

I agree the chart data can be passed directly, that was a mistake

packages/analytics/dashboard-renderer/src/composables/useContextLinks.ts
```diff
@@ -31,7 +31,7 @@ export default function useContextLinks(
   const canShowKebabMenu = computed(() => !['golden_signals', 'top_n', 'gauge'].includes(definition.value.chart.type))
 
   const canGenerateRequestsLink = computed(() => requestsBaseUrl.value && definition.value.query && definition.value.query.datasource !== 'llm_usage')
-
+  const canGenerateExploreLink = computed(() => exploreBaseUrl.value && definition.value.query && ['api_usage', 'llm_usage'].includes(definition.value.query.datasource))
```
adorack, line 34: Currently all datasources can jump to explore.  Notably, `basic` is missing from the above list; I think it's fine to either check for all three datasources if you want, but it's also OK to just check if `exploreBaseUrl.value` is truthy.

packages/analytics/dashboard-renderer/src/components/DashboardTile.cy.ts
```diff
@@ -329,4 +329,68 @@ describe('<DashboardTile />', () => {
...
+        query: {
+          datasource: 'random_unsupported_source',
+          metrics: [],
+          filters: [],
+        },
+      },
+    })
+
+    cy.getTestId('kebab-action-menu-1').click()
+    cy.getTestId('chart-jump-to-explore-1').should('not.exist')
```
adorack, line 346: The actual regression focused on `llm_usage` going to API requests, rather than a random unsupported source going to explore.  I think it'd be most helpful to test the actual regression.

adorack (changes requested):

adorack (approved):

PR #2416, opened 2025-08-28
feat(entities-plugins): datakit - avoid call and exit in response

2eha0:
# Summary

1. Avoid any nodes in the request phase to select a node from the response phase
2. Avoid the `call` node and the `exit` node to be added to the response phase

### DIFF
--- a/packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/FlowCanvas.vue
+++ b/packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/FlowCanvas.vue
@@ -48,23 +48,28 @@
         />
       </template>
     </VueFlow>
+
+    <div
+      v-if="disableDrop"
+      class="dk-flow-mask"
+    />
   </div>
 </template>
 
 <script setup lang="ts">
-import type { DragPayload, NodeId, NodePhase } from '../types'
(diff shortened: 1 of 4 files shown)


sumimakito (approved):

packages/entities/entities-plugins/src/components/free-form/Datakit/flow-editor/FlowCanvas.vue
```diff
@@ -80,8 +85,16 @@ const emit = defineEmits<{
   initialized: []
 }>()
 
+const { draggingNodePayload } = useEditorStore()
```
Justineo, line 88: I feel we can stick with `dataTransfer` for carrying the dragged node info. It naturally ties the data’s lifecycle to the drag session (so no need for manual reset), and also preserves interoperability with native DnD features like cursor feedback and potential external drag sources.

For showing forbidden states or reasons, we can still derive that during `dragover` without moving the payload into global store.

2eha0: The browser doesn't allow access to `dataTransfer` other than `dragstart ` and `drop` for security reasons.

Justineo (approved):
LGTM

Let's merge this first and let product/design to test out on Konnect.

PR #2437, opened 2025-09-03
fix(analytics): improve fallback axes titles [MA-4286]

jordojordo:
# Summary

Fix [MA-4286](https://konghq.atlassian.net/browse/MA-4286)

The returns for each analytics chart axis title type had a few inconsistencies. Namely:

- Returning localized strings that may not exist
- Returning `boolean`'s when a string is expected
- Not having proper fallbacks in place if a localized string didn't exist

This will add a few fallbacks to return a metric/dimension/granularity property when a localized string is not found; If neither is found, only then will `undefined` be returned.

[MA-4286]: https://konghq.atlassian.net/browse/MA-4286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
+++ b/packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
@@ -226,6 +226,10 @@ const tooltipMetricDisplay = computed<string | undefined>(() => {
 })
 
 const metricAxesTitle = computed<string | undefined>(() => {
+  if (props.chartOptions?.metricAxesTitle) {
+    return props.chartOptions?.metricAxesTitle
+  }
+
   if (!props.chartData?.meta.metric_names || !props.chartData?.meta.metric_units) {
     return undefined
   }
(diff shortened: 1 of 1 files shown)


packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
```diff
@@ -240,26 +243,44 @@ const metricAxesTitle = computed<string | undefined>(() => {
       return i18n.t('metricAxisTitles.size_in', { unit: i18n.t(`chartUnits.${metricUnit}`, { plural: 's' }) })
     }
   }
+
+  if (props.chartOptions?.metricAxesTitle) {
+    return props.chartOptions?.metricAxesTitle
+  }
```
Chalks, line 249: **question, potentially a blocker**

shouldn't this take priority above the i18n lookup? That is to say, if somebody added a `chartOptions.metricAxesTitle`, doesn't that imply they want that over the default? I suspect this block should move up to line 229ish.

jordojordo: Hmm I'm not sure, our current approach prioritizes the `metric_names` prop but perhaps it shouldn't?

Chalks (approved):

PR #2441, opened 2025-09-04
fix(entities-plugins): free-form scope is not correctly initialized

sumimakito:
Initialize the `scoped` ref in `<StandardLayout />` with `scopeIds` to fix an issue where a scoped plugin was displayed as a global plugin.

### Possible issue

* `scoped` is not reactive to `scopeIds` (so does `scopeId`), so it may be out of sync with `formModel`

KM-1701


2eha0 (approved):
LGTM

PR #2444, opened 2025-09-04
refactor(entities-plugins): datakit - simplify skip validation logic

2eha0:
# Summary
Replaced `newCreatedNodeId` with a boolean state `skipValidation` to make the logic for skipping validation on newly created nodes simpler.


Justineo (approved):
LGTM

PR #2454, opened 2025-09-05
fix: allow view requests and explore independent of zoom

filipgutica:
# Summary
Addresses https://konghq.atlassian.net/browse/MA-4288 

If any zoom option is provided allow brushing the chart to select an area.

## Changes
- Rename internal zoom prop in `TimeSeriesChart` to `brush` since `zoom` pertains to the specific zoom action. 
- Allow "brushing" if `timeseriesZoom` `exploreLink` or `requestsLink` is provided
- Add component test cases around interactivity


adorack (approved):
Thanks for adding the tests!

PR #2474, opened 2025-09-10
fix: add component test for no-select class

filipgutica:
# Summary

Add component test case called out by [ma-4263](https://konghq.atlassian.net/browse/MA-4263)


jordojordo (dismissed):
There should be some tests for these asks as well:
- The appropriate menu is displayed for actions.
- The correct time range and items appear in the menu.

PR #2490, opened 2025-09-15
fix(forms): scoped entities selected default value

jw-foss:
* Show `uuid` as selected value when certain scoped entity does not own a name
* Add test cases against the component `FieldAutoSuggestV2`.

[KM-1756]

[KM-1756]: https://konghq.atlassian.net/browse/KM-1756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/forms/src/components/fields/FieldAutoSuggestV2.vue
+++ b/packages/core/forms/src/components/fields/FieldAutoSuggestV2.vue
@@ -32,7 +32,7 @@
       <template #item="{ item }">
         <div class="entity-suggestion-item">
           <span class="entity-label">
-            {{ item.label ?? LABEL_PLACEHOLDER }}
+            {{ item.label ?? EMPTY_VALUE_PLACEHOLDER }}
           </span>
           <span class="entity-id">
             {{ item.id }}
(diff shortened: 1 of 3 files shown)


packages/core/forms/src/components/fields/FieldAutoSuggestV2.vue
```diff
@@ -42,7 +42,7 @@
 
       <template #selected-item="{ item }">
         <span class="selected-entity-item">
-          <span class="selected-entity-label">{{ item.label || '–' }}</span>
+          <span class="selected-entity-label">{{ item.label === LABEL_PLACEHOLDER ? item.value : item.label }}</span>
```
Justineo, line 45: Is `"-"` a valid name? If it is, I think we should display it as-is since the user explicitly chose that name.

jw-foss: `-` is a valid one, but if we do so, what should be shown for those with name set to `undefined` 🤔

jw-foss: Made a mistake, `-(short dash)` is different than `–(long dash)` , and `–(long dash)` is prohibited to be used as the entities' name. So I think the current implementation is ok for distinguishing entities' without name and user given `-(short dash)`

Justineo (approved):
LGTM

PR #2498, opened 2025-09-16
fix(entities-plugins): hide __ui_data in config view while editing

sumimakito:
Strip the `__ui_data` field (it will still be submitted via the UI) from the config displayed in the View Configuration slideout while editing the plugin. This applies to all plugins, as `__ui_data` should be considered a special field that should not be displayed for any plugin.

Screenshot 2025-09-16 at 14 21 03

KM-1759

### DIFF
--- a/packages/entities/entities-plugins/src/components/PluginForm.vue
+++ b/packages/entities/entities-plugins/src/components/PluginForm.vue
@@ -117,18 +117,18 @@
         <template #json>
           <JsonCodeBlock
             :config="config"
-            :entity-record="getRequestBody"
+            :entity-record="viewConfigurationRecord"
             :fetcher-url="submitUrl"
             :request-method="props.pluginId ? 'put' : 'post'"
           />
         </template>
         <template #yaml>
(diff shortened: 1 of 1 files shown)


packages/entities/entities-plugins/src/components/PluginForm.vue
```diff
@@ -1335,6 +1335,16 @@ const getRequestBody = computed((): Record<string, any> => {
   return requestBody
 })
 
+// Read-only record for the View Configuration feature
+const viewConfigurationRecord = computed(() => {
+  const record = { ...getRequestBody.value }
+
+  // Don't show UI data in the configuration view
+  delete record['__ui_data']
```
Justineo, line 1343:
```suggestion
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
  const { __ui_data, ...record } = getRequestBody.value
```

I think we can skip delete and use destructuring instead.

Justineo (approved):
LGTM

PR #2508, opened 2025-09-18
fix: improve plugin performance

filipgutica:
# Summary

Remove unnecessary `chart.update()` and `chart.draw()` calls that were firing at high frequency leading to performance degradation especially in charts with many data points.


adorack (changes requested):
I'd like to review.

adorack (approved):
That is a lot better, thanks!

PR #2511, opened 2025-09-18
fix(entities-plugins): remove unsupported scoping fields in detail view

sumimakito:
Remove unsupported scoping fields in the plugin detail view according to the plugin metadata. This does not affect Konnect, as their API does not return the unsupported fields in the first place.

1. If the scope definition is missing for a plugin, keep it as is, as it could be a new bundled plugin or a custom plugin with an incorrect schema.
2. If a scoping field is not empty, but the plugin metadata does not have that scope (unsupported if telling from the metadata), keep the scoping field in case we have incorrect metadata.

|Structured|JSON|
|:-|:-|
|Screenshot 2025-09-18 at 16 16 37|Screenshot 2025-09-18 at 16 16 22|

KM-1789

### DIFF
--- a/packages/entities/entities-plugins/src/components/PluginConfigCard.vue
+++ b/packages/entities/entities-plugins/src/components/PluginConfigCard.vue
@@ -144,30 +144,35 @@
 </template>
 
 <script setup lang="ts">
-import type { PropType } from 'vue'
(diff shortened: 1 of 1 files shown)


Leopoldthecoder (approved):
LGTM

PR #2513, opened 2025-09-18
feat (analytics): basic sparkline chart

Chalks:
# Summary

A sparkline chart!

Screenshot 2025-09-18 at 5 42 27 PM

### Description

This version of sparkline chart is a very simple one that ONLY accepts arrays of timestamps as data. We provide each rendered sparkline with the same range (start/end timestamps and a max count) so that every rendered datapoint is scaled the same way as all other sparklines that share those properties, regardless of number of timestamps given. It also accepts a `pointRenderCount` (default 24) which determines how many datapoints to draw. Effectively, the sparkline buckets all timestamps given it into `(maxStamp - minStamp) / pointRenderCount` datapoints. This is primarily so that it doesn't look silly when rendered with several thousand bits of data.

### Props

* `showLabel?: boolean` - whether to display a label when it has data, default false.
* `disableTooltip?: boolean` - whether to display a tooltip on hover or not. Default false
* `maxCount: number` - the maximum value to render. Difficult to determine before providing data, so there is an `@max` event that you can use to react to different sparklines with different max counts.
* `maxStamp?: number` - defaults to `Date.now()`, sets the upper bound on timestamps to render
* `minStamp: number` - sets the lower bound on timestamps to render
* `pointRenderCount?: number` - defaults to `24`, sets the number of points to draw (i.e. bucket your timestamps into ranges between minStamp and maxStamp)
* `datasets: SparklineDataset[]` - where `type SparklineDataset = { color?: string, label: string, timestamps: number[] }`. Timestamps do not have to be within the bounds of `minStamp` and `maxStamp`, nor do they have to be in order. It's sorted by default and anything outside bounds is ignored.
* `tooltipTitle?: string` - the title to display on the tooltip
* `type: 'bar' | 'line' | 'step'` - defaults to `'bar'` because that makes the most sense for counts, usually.

### Slots

* `<template #label="{ total, totalByDataset}" />` - replaces the label content if `showLabel` is true
* `<template #empty />` - replaces the empty state content

### Events

* `(e: 'max', max: number): void`, i.e. `@max` - Useful for watching what the max value each sparkline you render buckets its datasets into. When rendering multiple sparklines, you want them all to have the same `maxCount` because otherwise the scale is thrown off.

### Screenshots

**Multiple datasets**
Screenshot 2025-09-18 at 5 40 05 PM

**Single datasets at different sizes**
Screenshot 2025-09-18 at 5 39 04 PM

Screenshot 2025-09-18 at 5 39 42 PM

**Empty datasets**
Screenshot 2025-09-18 at 5 40 14 PM

**Lots of datasets rendered with random types**
Screenshot 2025-09-18 at 6 09 38 PM

**Label displayed**

Screenshot 2025-09-19 at 12 45 33 PM

**Label with too much data (note the title hover effect)**

Screenshot 2025-09-19 at 12 48 11 PM

**Tooltip hovering**

https://github.com/user-attachments/assets/63f922f1-b333-43e1-9ec7-ff2195c28da6

https://github.com/user-attachments/assets/1f954242-8e87-4373-8602-fc649f75479b

**Tooltip date offset**

When the gap between points is greater than 24 hours, it shows the full date range:
Screenshot 2025-09-19 at 12 08 20 PM

### How to test

I highly recommend you pull this branch down and run it locally. Open up the sandbox page and play with this:

Screenshot 2025-09-18 at 6 11 54 PM

### DIFF
--- a/packages/analytics/analytics-chart/src/components/SparklineChart.vue
+++ b/packages/analytics/analytics-chart/src/components/SparklineChart.vue
@@ -10,7 +10,7 @@
       </slot>
     </div>
     <Bar
-      v-else-if="type === 'bar'"
+      v-else-if="type === 'sparkline_bar'"
       :chart-id="chartId"
       :data="chartData"
       data-testId="sparkline-bar"
@@ -18,10 +18,10 @@
       :plugins="plugins"
     />
     <Line
-      v-else-if="type === 'line' || type === 'step'"
+      v-else-if="type === 'sparkline_line' || type === 'sparkline_step'"
       :chart-id="chartId"
       :data="chartData"
-      :data-testId="type === 'line' ? 'sparkline-line' : 'sparkline-step'"
+      :data-testId="type === 'sparkline_line' ? 'sparkline-line' : 'sparkline-step'"
       :options="(options as any)"
       :plugins="plugins"
     />
@@ -63,7 +63,7 @@ import { formatTime } from '@kong-ui-public/analytics-utilities'
 import composables from '../composables'
 import { lineChartTooltipBehavior } from '../utils'
 import bucketTimestamps from '../utils/bucketTimestamps'
-import type { ExternalTooltipContext, SparklineDataset, TooltipState } from '../types'
+import type { ExternalTooltipContext, SparklineDataset, SparklineType, TooltipState } from '../types'
 import { VerticalLinePlugin } from './chart-plugins/VerticalLinePlugin'
 import ToolTip from './chart-plugins/ChartTooltip.vue'
 
@@ -76,7 +76,7 @@ const {
   pointRenderCount = 24,
   showLabel = false,
   tooltipTitle,
-  type = 'bar',
+  type = 'sparkline_bar',
 } = defineProps<{
   datasets: SparklineDataset[]
   disableTooltip?: boolean
@@ -101,7 +101,7 @@ const {
   pointRenderCount?: number
   showLabel?: boolean
   tooltipTitle?: string
-  type: 'bar' | 'line' | 'step'
+  type: SparklineType
 }>()
 
 const chartId = crypto.randomUUID()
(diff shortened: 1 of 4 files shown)


packages/analytics/analytics-chart/src/utils/commonOptions.ts
```diff
@@ -22,10 +29,12 @@ export const lineChartTooltipBehavior = (tooltipData: TooltipState, context: Ext
...
-    tooltipData.tooltipContext = formatTooltipTimestampByGranularity({
-      tickValue: new Date(tooltip.dataPoints[0].parsed.x),
-      granularity,
-    })
+    tooltipData.tooltipContext = options?.contextFormatter
+      ? options.contextFormatter(tooltip.dataPoints[0].parsed.x, granularity)
+      : formatTooltipTimestampByGranularity({
+        tickValue: new Date(tooltip.dataPoints[0].parsed.x),
+        granularity,
+      })
```
Chalks, line 37: This is the only change that's not strictly related to sparkline. Shouldn't affect anything else as it's just adding an optional formatter, but I'd appreciate extra scrutiny here.

packages/analytics/analytics-chart/src/components/SparklineChart.vue
```diff
@@ -0,0 +1,384 @@
...
+  /**
+   * The number of datapoints to render. Will bucket timestamps into ranges that
+   * are `(maxStamp - minStamp) / pointRenderCount` apart. `pointRenderCount`
+   * must be greater than 1 (will default to 2 if a value less than 2 is
+   * provided) as you can't draw a line without at least 2 points rendered.
+   */
+  pointRenderCount?: number
+  showLabel?: boolean
+  tooltipTitle?: string
+  type: 'bar' | 'line' | 'step'
```
filipgutica, line 104: If this may become an "official chart type" supported by dashboards/reports, could update these types to be a bit more descriptive. 

EG `spark_line` `spark_bar` `spark_step` ? (kind of following the pattern of timeseries_bar/timeseries_line) 

Could also throw the types [here](https://github.com/Kong/public-ui-components/blob/main/packages/analytics/analytics-chart/src/types/chart-types.ts) where all the types are tracked?

filipgutica (changes requested):
just one comment

packages/analytics/analytics-chart/src/utils/commonOptions.ts
filipgutica: I think that's fine

PR #2524, opened 2025-09-25
feat: draw threshold intersection brush

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-4302

Detect when data crosses thresholds and brush the area

figure out where each datasource intersects the threshold, merge overlapping "intersections" so we don't draw overlapping brushes

image

image

### DIFF
--- a/packages/analytics/analytics-chart/src/components/chart-plugins/ThresholdPlugin.ts
+++ b/packages/analytics/analytics-chart/src/components/chart-plugins/ThresholdPlugin.ts
@@ -121,8 +121,7 @@ export class ThresholdPlugin implements Plugin {
 
   constructor(private i18n: ReturnType<typeof createI18n<typeof english>>) {}
 
-  beforeInit(chart: Chart, _: any, pluginOptions: ThresholdOptions): void {
-    const canvas = chart.canvas
+  private _syncThresholds(pluginOptions: ThresholdOptions) {
     for (const key of Object.keys(pluginOptions.threshold || {})) {
       const threshold = pluginOptions.threshold?.[key as ExploreAggregations]
       if (threshold) {
@@ -132,6 +131,17 @@ export class ThresholdPlugin implements Plugin {
         } as Record<ExploreAggregations, ThresholdExtra[]>
       }
     }
+  }
+
+  // Gets called when chart options are updated
+  beforeUpdate(chart: Chart, _: any, pluginOptions: ThresholdOptions): void {
+    this._syncThresholds(pluginOptions)
+  }
+
+
+  beforeInit(chart: Chart, _: any, pluginOptions: ThresholdOptions): void {
+    const canvas = chart.canvas
+    this._syncThresholds(pluginOptions)
 
     const onMouseMove = (event: MouseEvent) => {
       if (chart) {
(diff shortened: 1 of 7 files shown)


packages/analytics/analytics-chart/src/components/chart-plugins/ThresholdPlugin.ts
```diff
@@ -15,6 +17,102 @@ interface ThresholdOptions {
...
+const getExactIntersection = (p0: Point, p1: Point, targetY: number): number => {
+  const dy = p1.y - p0.y
+  if (dy === 0) {
+    return p1.x
+  }
+  const f = (targetY - p0.y) / dy
+  return p0.x + f * (p1.x - p0.x)
+}
+
+const getThresholdIntersections = (chart: Chart, thresholds: ThresholdExtra[]): ThresholdIntersection[] => {
```
Chalks, line 39: **not a blocker, but likely a good idea**

easy enough to export this and then we can unit test it quite simply. Probably worth doing since it's a bit of a tricky business

packages/analytics/analytics-chart/src/components/chart-plugins/ThresholdPlugin.ts
```diff
@@ -15,6 +17,102 @@ interface ThresholdOptions {
...
+  const intersections: ThresholdIntersection[] = []
+  chart.data.datasets.forEach((dataset) => {
+    if (dataset.hidden) {
+      return
+    }
```
Chalks, line 44: **maybe a blocker?**

When you have multiple datasets and they have overlapping areas, there's no distinction between them. I don't know what else you would draw here, but it feels kinda weird.

Screenshot 2025-09-26 at 4 05 22 PM

Worse (and this actually is a blocker I think) it doesn't seem to update when you show/hide series? Sort of? Sometimes? I would expect a gap in the places where it dips below and I don't see why it doesn't have a gap:

Screenshot 2025-09-26 at 4 03 28 PM

packages/analytics/analytics-chart/src/components/chart-plugins/ThresholdPlugin.ts
```diff
@@ -15,6 +17,102 @@ interface ThresholdOptions {
...
+
+    const dataPoints = dataset.data as Array<{ x: number, y: number }>
```
Chalks, line 46: **blocker**

For some reason when you hide all datasets it still shows a single highlight? It's odd.

Screenshot 2025-09-26 at 4 06 00 PM

packages/analytics/analytics-chart/src/components/chart-plugins/ThresholdPlugin.ts
```diff
@@ -15,6 +17,102 @@ interface ThresholdOptions {
...
+    return []
+  }
+
+  intersections.sort((a, b) => a.type.localeCompare(b.type) || a.start - b.start)
+
+  const merged: ThresholdIntersection[] = []
+
+  for (const curr of intersections) {
+    const last = merged[merged.length - 1]
+    if (last && last.type === curr.type && curr.start <= last.end) {
```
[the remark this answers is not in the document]
Chalks: For clarity:

Imagine you have a warn threshold and an error threshold. You get the following threshold intersections from all your datasets:

```
[
  {start: 1, end: 5, type: warn},
  {start: 2, end: 5, type: error},
  {start: 3, end: 5, type: warn},
  {start: 4, end: 8, type: warn},
]
```

when you merge those you would expect to get:

`[{start: 1, end: 8, type: warn}, {start: 2, end: 5, type: error}]`

but with the logic that's there now you would actually get

`[{start: 1, end: 5, type: warn}, {start: 2, end: 5, type: error}, {start: 3, end: 8, type: warn}]`

PR #2529, opened 2025-09-29
fix(entities-plugins): confluent auth mode is undefined [KM-1824]

2eha0:
# Summary

KM-1824


jw-foss (approved):
LGTM, should we add some test cases to cover this?

Leopoldthecoder (approved):
LGTM

PR #2536, opened 2025-10-02
fix(analytics): bump the limit override for geo map charts [MA-3698]

jordojordo:
# Summary

Fix [MA-3698](https://konghq.atlassian.net/browse/MA-3698)

This will add a `limit` bump for Geo maps within the `QueryDataProvider` component by accepting a new prop for the `chartType`, which is used to determine if the limit should be bumped prior to issuing the query. The `limit` will be bumped to the length of country codes we have within our `COUNTRIES` const.


adorack (changes requested):
I'd like to review.

adorack (changes requested):

adorack (approved):

PR #2539, opened 2025-10-03
fix(analytics): keep consistent metric formatting for geo maps [MA-4125]

jordojordo:
# Summary

Fix [MA-4125](https://konghq.atlassian.net/browse/MA-4125)
Contains chore from https://github.com/Kong/public-ui-components/pull/2532

This will add the ability to specify if a metric should be truncated within the GeoMap and also truncate them by default for non-latency type metrics in dashboards.

There are also a few updates to the sandbox to make things a bit cleaner, fix some typing, and add some functionality to the geo map sandbox.

[sandbox.webm](https://github.com/user-attachments/assets/79a7a190-698b-46db-a6f3-d4612f9f7373)

[MA-4125]: https://konghq.atlassian.net/browse/MA-4125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
+++ b/packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
@@ -79,6 +79,7 @@ const {
   showTooltipValue = true,
   fitToCountry = undefined,
   bounds = undefined,
+  truncatedMetric = true,
 } = defineProps<{
   countryMetrics: Record<string, number>
   metricUnit: MetricUnits
@@ -87,6 +88,7 @@ const {
   showTooltipValue?: boolean
   fitToCountry?: CountryISOA2 | undefined
   bounds?: Array<[number, number]>
+  truncatedMetric?: boolean
 }>()
 
 const emit = defineEmits<{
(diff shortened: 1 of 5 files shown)


packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
```diff
@@ -79,6 +79,7 @@ const {
   showTooltipValue = true,
   fitToCountry = undefined,
   bounds = undefined,
+  truncatedMetric = true,
```
filipgutica, line 82: Hmm, I'm not sure we want to do this through another prop. 

We'd have to update the dashboard schema as well and then everywhere a dashboard or geo map is used, we'd need logic for figuring out when to truncate or not. 

Is there some default behaviour we can implement internally to decide if the metric should be truncated, rather than introducing a new prop? I think analytics-chart does something like that

filipgutica (changes requested):
Just one comment

packages/analytics/dashboard-renderer/src/components/GeoMapRenderer.vue
```diff
@@ -60,6 +61,10 @@ const countryMetricUnit = computed(() => {
   return metricUnits && metricNames && metricUnits[metricNames[0]]
 })
 
+const truncatedMetric = computed(() => {
```
filipgutica, line 64: We could move this into the geo map component rather than adding the `truncatedMetric` prop

Rather than the truncated metric prop, I'd rather pass the metric through to the geo map component 🤔

filipgutica (approved):

PR #3055, opened 2026-03-25
feat(analytics): export dashboard to pdf [MA-5195]

jordojordo:
# Summary

Close [MA-5195](https://konghq.atlassian.net/browse/MA-5195)

This will add an available function to the DashboardRenderer for exporting a dashboard as a PDF.

It uses [snapdom](https://github.com/zumerlab/snapdom) to create a canvas of the dom, then provides the canvas to [jspdf](https://github.com/parallax/jsPDF) to generate a PDF from the canvas with calculation to ensure tiles aren't split between pages but attempts to retain the layout of a dashboard.

A bit of nuance surrounding map tiles requires a workaround due to the snapdom attempting to capture the draw buffer from WebGL which is not saved (available to capture) after writing to the frame buffer (what we see in the browser).

In order for those tiles to work, the draw buffer needs to be redrawn before the capture event happens. This happens by a `kapture:prepare` event that the `AnalyticsGeoMap` will listen for on mount (removed on unmount), when that event is caught the map is redrawn and temporarily converted into an `` element which gives snapdom something to capture. Once captured the `` is removed and the canvas is restored back to normal.

<details closed>
<summary>Sandbox demo</summary>

[export-pdf.webm](https://github.com/user-attachments/assets/e1090b4c-46f1-420f-b9b6-bd57834fab11)

</details>

Snapdom seemed the most capable and well maintained when comparing with [html2canvas](https://github.com/niklasvh/html2canvas). [jsPDF](https://github.com/parallax/jsPDF) is likely the best library for this job as it's actively maintained and fairly lightweight.

[MA-5195]: https://konghq.atlassian.net/browse/MA-5195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
+++ b/packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
@@ -51,8 +51,8 @@ import type { MapFeatureCollection, MetricUnits } from '../types'
 import type { Feature, MultiPolygon, Geometry, GeoJsonProperties, FeatureCollection } from 'geojson'
 import composables from '../composables'
 import 'maplibre-gl/dist/maplibre-gl.css'
-import { CAPTURE_PREPARE_EVENT, CAPTURE_RESTORE_EVENT } from '@kong-ui-public/analytics-utilities'
-import type { ExploreAggregations, CountryISOA2, CapturePrepareDetail } from '@kong-ui-public/analytics-utilities'
+import type { ExploreAggregations, CountryISOA2 } from '@kong-ui-public/analytics-utilities'
+import { registerWebglCapture } from '@kong-ui-public/analytics-utilities'
 import lakes from '../ne_110m_lakes.json'
 import * as geobuf from 'geobuf'
 import Pbf from 'pbf'
@@ -241,55 +241,18 @@ const emitBounds = () => {
 
 const debouncedEmitBounds = debounce(emitBounds, 300)
 
-// For PDF export the canvas will temporarily be an <img> to allow for
-// snapshotting the draw buffer of WebGL.
-let restoreSnapshot: (() => void) | null = null
-
-async function createCaptureSnapshot(): Promise<void> {
-  const glCanvas = map.value?.getCanvas()
-
-  if (!map.value || !glCanvas) {
-    return
-  }
-
-  // Recreates the drawing buffer for `toDataURL`
-  map.value.redraw()
-
-  const snapshot = new Image()
-  snapshot.src = glCanvas.toDataURL('image/png')
-
-  await snapshot.decode()
-
-  snapshot.style.cssText = `position:absolute;top:0;left:0;width:${glCanvas.clientWidth}px;height:${glCanvas.clientHeight}px;`
-  glCanvas.parentElement?.appendChild(snapshot)
-  glCanvas.style.visibility = 'hidden'
-
-  // Get rid of the <img> element and return the canvas to normal
-  restoreSnapshot = () => {
-    snapshot.remove()
-    glCanvas.style.visibility = ''
-  }
-}
-
-function onCapturePrepare(event: Event): void {
-  (event as CustomEvent<CapturePrepareDetail>).detail?.waitUntil(createCaptureSnapshot())
-}
-
-function onCaptureRestore(): void {
-  restoreSnapshot?.()
-  restoreSnapshot = null
-}
+// For PDF export the WebGL canvas is temporarily swapped for a static <img> so
+// its draw buffer can be captured
+const stopWebglCapture = registerWebglCapture(
+  () => map.value?.getCanvas(),
+  { redraw: () => map.value?.redraw() },
+)
 
 onUnmounted(() => {
-  window.removeEventListener(CAPTURE_PREPARE_EVENT, onCapturePrepare)
-  window.removeEventListener(CAPTURE_RESTORE_EVENT, onCaptureRestore)
-  onCaptureRestore()
+  stopWebglCapture()
 })
 
 onMounted(async () => {
-  window.addEventListener(CAPTURE_PREPARE_EVENT, onCapturePrepare)
-  window.addEventListener(CAPTURE_RESTORE_EVENT, onCaptureRestore)
-
   try {
     const countriesPbfUrl = await countriesPbfUrlPromise
     const response = await fetch(countriesPbfUrl)
--- a/packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
+++ b/packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
@@ -44,14 +44,15 @@
 </template>
 
 <script setup lang="ts">
-import { ref, onMounted, watch, computed, useId, toRef } from 'vue'
+import { ref, onMounted, onUnmounted, watch, computed, useId, toRef } from 'vue'
 import { Map } from 'maplibre-gl'
 import type { ColorSpecification, DataDrivenPropertyValueSpecification, ExpressionSpecification, LngLatBoundsLike, MapOptions } from 'maplibre-gl'
 import type { MapFeatureCollection, MetricUnits } from '../types'
 import type { Feature, MultiPolygon, Geometry, GeoJsonProperties, FeatureCollection } from 'geojson'
 import composables from '../composables'
 import 'maplibre-gl/dist/maplibre-gl.css'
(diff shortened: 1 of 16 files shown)


packages/analytics/dashboard-renderer/sandbox/pages/EditableDashboardDemo.vue
```diff
@@ -316,6 +323,13 @@ const addTile = () => {
   })
 }
 
+const exportPdf = () => {
+  dashboardRendererRef.value?.exportPdf({
+    title: 'Sandbox Editable Dashboard',
+    dashboardUrl: 'https://cloud.konghq.com/analytics/dashboards?ref=pdf',
```
Chalks: **nit**

prefer `utm_source` over `ref` unless there's a kong-specific reason for this.

packages/analytics/dashboard-renderer/sandbox/pages/RendererDemo.vue
```diff
@@ -462,6 +469,13 @@ const refresh = () => {
 const toggleFullscreen = () => {
   dashboardRendererRef.value?.toggleFullscreen()
 }
+
+const exportPdf = () => {
+  dashboardRendererRef.value?.exportPdf({
+    title: 'Sandbox Dashboard',
+    dashboardUrl: 'https://cloud.konghq.com/analytics/dashboards?ref=pdf',
```
Chalks: **nit**

`utm_source` vs `ref` again

packages/analytics/dashboard-renderer/src/composables/useExportPdf.ts
```diff
@@ -0,0 +1,325 @@
...
+    .replace(/^-+|-+$/g, '')
+}
+
+export function buildFilename(filename: string | undefined, title: string | undefined, now: Date): string {
+  const base = filename || (title && slugify(title)) || 'dashboard-export'
+  const yyyy = now.getFullYear()
+  const mm = String(now.getMonth() + 1).padStart(2, '0')
+  const dd = String(now.getDate()).padStart(2, '0')
+
+  return `${base}-${yyyy}-${mm}-${dd}.pdf`
```
Chalks: **nit, existing library handles this**

`date-fns` makes this super easy with [format](https://date-fns.org/v4.4.0/docs/format), e.g.


```suggestion
  const base = filename || (title && slugify(title)) || 'dashboard-export'
  const dateString = format(now, 'YYYY-MM-dd')

  return `${base}-${dateString}.pdf`
```

hmmmm, I guess we don't have date-fns in this package but I feel like it would be worth it for just this. shrug.jpg.

packages/analytics/dashboard-renderer/src/types/renderer-types.ts
```diff
@@ -25,6 +25,46 @@ export interface DashboardRendererContextInternal extends Required<DashboardRend
...
+   * Defaults to a slug of `title`, falling back to 'dashboard-export'.
+   */
+  filename?: string
+  /** Dashboard name used in the header */
+  title?: string
+  /** Second header line */
+  subtitle?: string
+  /** URL to the dashboard in the footer. */
+  dashboardUrl?: string
+  /** Page orientation. Defaults to 'landscape'. */
```
Chalks, line 40: **nit, maybe blocker**

Why default to landscape?

packages/analytics/dashboard-renderer/src/composables/useExportPdf.ts
```diff
@@ -0,0 +1,325 @@
...
+        beforeClone: () => prepareForCapture(),
+        afterClone: () => restoreAfterCapture(),
+      }
+
+      const canvas = await snapdom.toCanvas(element, { scale, exclude, plugins: [webglSnapshotPlugin] })
+
+      exportState.value = { status: 'generating' }
+
+      const now = new Date()
+      const pdf = new jsPDF({ orientation, unit: 'mm', format: 'a4' })
```
Chalks: **nit, potential blocker**

Why A4 size? I think the majority of our customers are US based, so A4 would be rather awkward for a decent chunk of them. Also why isn't page size an argument that can be modified like `margin`?

I think it would be nice to have this be:

1. a prop that can be modified (I suspect eventually a pdf export will have options for how to export it)
2. have it default to Letter or A4 based on your geo (which we should have access to)

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
```diff
@@ -240,7 +241,55 @@ const emitBounds = () => {
...
+function onCaptureRestore(): void {
+  restoreSnapshot?.()
+  restoreSnapshot = null
+}
+
+onUnmounted(() => {
+  window.removeEventListener(CAPTURE_PREPARE_EVENT, onCapturePrepare)
+  window.removeEventListener(CAPTURE_RESTORE_EVENT, onCaptureRestore)
+  onCaptureRestore()
+})
```
Chalks, line 253: **question, potential blocker**

I take it that the geo map has renders in a way that requires something custom to access the data that none of the other visualizations have?

I'm not sure I love the idea of including this in the `AnalyticsGeoMap` component directly. Is there any way we could make this more external? I'm primarily thinking "well, what if we have another FooRenderer that also uses webgl, will we have to do the same thing there?" It seems messy.

packages/analytics/analytics-utilities/src/captureProtocol.ts
```diff
@@ -0,0 +1,27 @@
+// Protocol for preparing WebGL content for PDF export.
+//
+// MapLibre renders into a WebGL canvas with `preserveDrawingBuffer: false`,
+// so capture libraries that read canvases using `toDataURL()` get a blank image.
+// This will temporarily swap the canvas for a static <img> while capturing.
+
+export const CAPTURE_PREPARE_EVENT = 'kapture:prepare'
```
Chalks, line 3: > kapture

:suspicious-fry:

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
Chalks: I'm also kind of thinking about slottable tiles. Could we create some kind of contract that requires a slottable (or at least gives the slottable developer hints on how to implement) to have something that allows us to include it in our pdf?

Chalks (commented):
some questions. Overall this is looking pretty solid.

packages/analytics/dashboard-renderer/sandbox/pages/EditableDashboardDemo.vue
jordojordo: It was mentioned in the PRD, but I'm not partial to it. I'll change it and if someone complains we can go back to `ref`.

packages/analytics/dashboard-renderer/src/composables/useExportPdf.ts
jordojordo: Ah yeah that is a bit nicer, I don't think it would hurt to have that package honestly.

packages/analytics/dashboard-renderer/src/types/renderer-types.ts
jordojordo: PRD mentions "orientation (landscape likely)", I could see the reasoning behind this as the window tends to be wider than it is tall, but I'm not opposed to hearing other opinions.

packages/analytics/dashboard-renderer/src/composables/useExportPdf.ts
jordojordo: I went with the first option mentioned in the PRD, but you bring up a good point. I'll add a page size prop and default to letter

packages/analytics/analytics-geo-map/src/components/AnalyticsGeoMap.vue
jordojordo: I'll have to give this some thought, I was solely focused on the map component working and didn't consider slottable tiles or the possibility of having other webgl things.

packages/analytics/analytics-utilities/src/captureProtocol.ts
jordojordo: :trollface: 

`kCapture` is just plain wrong.

Chalks (dismissed):

Chalks (approved):

PR #3058, opened 2026-03-26
feat(entities-plugins): add metering-and-billing-plugin form

TT1228:
# Summary

Add custom metering-and-plugin form


packages/entities/entities-plugins/src/components/free-form/shared/ArrayField.vue
```diff
@@ -23,23 +23,38 @@
...
-        class="ff-array-field-label"
-        v-bind="fieldAttrs"
-        :data-testid="`ff-label-${field.path.value}`"
-        :tooltip-attributes="fieldAttrs.labelAttributes.tooltipAttributes"
-      >
-        {{ fieldAttrs.label }}
-        <template
-          v-if="fieldAttrs.labelAttributes?.info"
-          #tooltip
+      <div class="ff-array-field-header-toggle">
```
2eha0: Should we implement the collapsible by composing `CollapsibleSection` and `ArrayField` in the template instead of introducing a new feature in the `ArrayField`, for example:
```vue
<CollapsibleSection>
  <ArrayField hide-label name="xxx" />
</CollapsibleSection>
```

packages/entities/entities-plugins/src/components/free-form/plugins/metering-and-billing/index.ts
```diff
@@ -0,0 +1,102 @@
...
+      match: ({ path }) => path.startsWith('config.attributes.') && path.endsWith('.look_up_value_in'),
+      component: StringField,
+      propsOverrides: { placeholder: t('plugins.free-form.metering-and-billing.fields.attributes.look_up_value_in.placeholder') },
+    },
+    {
+      match: ({ path }) => path.startsWith('config.attributes.') && path.endsWith('.event_property_name'),
+      component: StringField,
+      propsOverrides: { placeholder: t('plugins.free-form.metering-and-billing.fields.attributes.event_property_name.placeholder') },
+    },
+  ],
```
2eha0: I suggest overriding these fields in the template directly, and keep the config file simple:
```ts
export default definePluginConfig({
  experimental: true,
  component: MeteringAndBillingForm,
})
```

2eha0: Otherwise, we have to jump between files to understand which fields are customized

packages/entities/entities-plugins/src/components/free-form/shared/CollapsibleSection.vue
```diff
@@ -0,0 +1,141 @@
+<template>
```
2eha0, line 1: We should refactor `ObjectField` to reuse this component of the collapsibility, but not blocking in this PR

packages/entities/entities-plugins/src/components/free-form/plugins/metering-and-billing/index.ts
TT1228: Well it was in template before, i moved them into the config file after merging 🤣

2eha0: I think we should use either the config file or the template, we shouldn't use both of them

2eha0 (commented):
LGTM

2eha0 (approved):
LGTM

hangrao (approved):

PR #3108, opened 2026-04-09
feat(workspace): add workspace support to shared and gateway service

nekolab:
Expose the `workspace` prop to Konnect.

The `workspace` prop used to be available only in Kong Manager. We’re now introducing it to Konnect as well, so users can specify which `(control plane, workspace)` they want to work with when using Konnect.

Because this is a big change that affects all core entities, we’re first rolling it out in the `entities-shared` and `entities-gateway-services` packages, and then propagating it to the rest of the codebase. It's safe because the `workspace` prop is still optional and will not take any effect when not specified.

Tests were added to cover the new workspace prop in both entities-shared and entities-gateway-services.

JIRA: KM-2445

### DIFF
--- a/packages/entities/entities-gateway-services/docs/gateway-service-config-card.md
+++ b/packages/entities/entities-gateway-services/docs/gateway-service-config-card.md
@@ -53,7 +53,7 @@ A config card component for gateway services.
     - type: `string`
     - required: `true`
     - default: `undefined`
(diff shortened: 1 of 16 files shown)


Leopoldthecoder (dismissed):
LGTM

hangrao (approved):

PR #3122, opened 2026-04-10
fix(entities-plugins): convert existing consumer_claim to consumer_claims

TT1228:
# Summary

On form init, if config-consumer_claim has a value and config-consumer_claims is empty, automatically populate consumer_claims with the deprecated value wrapped in an outer array (string[] → string[][])
The original consumer_claim value is not passed in payload because it's in shorthand_fields and marked deprecated
Adds unit tests covering array migration, string normalization, no-overwrite, and null cases

# Why
consumer_claim is deprecated in favor of consumer_claims but existing users may have data stored under the old field. Without this migration the UI would render an empty consumer_claims field and discard consumer_claim field even when data exists.

### DIFF
--- a/packages/core/forms/src/components/forms/OIDCForm.vue
+++ b/packages/core/forms/src/components/forms/OIDCForm.vue
@@ -368,6 +368,18 @@ export default {
             }
           }
         }
+
+        // Migrate deprecated consumer_claim → consumer_claims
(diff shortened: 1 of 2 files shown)


Leopoldthecoder (approved):
LGTM

PR #3133, opened 2026-04-15
fix(plugins-endpoints): adjust called url [KOKO-3398]

LorenzoScebba:
# Summary

This PR changes the following endpoints from:

- /v2/control-planes/{controlPlaneId}/**schemas/core-entities**/plugins/{plugin} -> /v2/control-planes/{controlPlaneId}/**core-entities/schemas**/plugins/{plugin}
- /v2/control-planes/{controlPlaneId}/**schemas/core-entities**/{plugin} -> /v2/control-planes/{controlPlaneId}/**core-entities/schemas**/{plugin}

First change ref:  https://developer.konghq.com/api/konnect/control-planes-config/v2/#/operations/fetch-plugin-schema (Get plugin schema)

image

Second change: This endpoint seems to never get called, this endpoint might be called if `treatAsCredential` is true, which would require `props.config.entityId` to be provided, but entityId is always commented:

- https://github.com/Kong/public-ui-components/blob/74ca9d142fc0db70a90485003e84576c66385df4/packages/entities/entities-plugins/sandbox/pages/PluginFormPage.vue#L102
- https://github.com/Kong/public-ui-components/blob/74ca9d142fc0db70a90485003e84576c66385df4/packages/entities/entities-plugins/sandbox/pages/PluginFormPage.vue#L120
- https://github.com/Kong/public-ui-components/blob/73d690d7be1d19fab7cfbadf28a325cb3eb9eeb4/packages/entities/entities-plugins/sandbox/pages/PluginFormPlayground.vue#L160
- https://github.com/Kong/public-ui-components/blob/73d690d7be1d19fab7cfbadf28a325cb3eb9eeb4/packages/entities/entities-plugins/sandbox/pages/PluginFormPlayground.vue#L175

### DIFF
--- a/packages/entities/entities-plugins/src/components/PluginForm.cy.ts
+++ b/packages/entities/entities-plugins/src/components/PluginForm.cy.ts
@@ -1036,7 +1036,7 @@ describe('<PluginForm />', () => {
       cy.intercept(
         {
           method: 'GET',
-          url: `${baseConfigKonnect.apiBaseUrl}/v2/control-planes/${baseConfigKonnect.controlPlaneId}/schemas/core-entities/plugins/*`,
+          url: `${baseConfigKonnect.apiBaseUrl}/v2/control-planes/${baseConfigKonnect.controlPlaneId}/core-entities/schemas/plugins/*`,
         },
         {
           statusCode: 200,
@@ -1764,7 +1764,7 @@ describe('<PluginForm />', () => {
       cy.intercept(
         {
           method: 'GET',
-          url: `${baseConfigKonnect.apiBaseUrl}/v2/control-planes/${baseConfigKonnect.controlPlaneId}/schemas/core-entities/plugins/cors`,
+          url: `${baseConfigKonnect.apiBaseUrl}/v2/control-planes/${baseConfigKonnect.controlPlaneId}/core-entities/schemas/plugins/cors`,
         },
         {
           statusCode: 500,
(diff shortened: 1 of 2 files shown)


Leopoldthecoder:
Hi @LorenzoScebba, it seems `/schemas/core-entities/plugins/{plugin}` and `/core-entities/schemas/plugins/{plugin}` both exists, otherwise the Konnect UI would break. Could you share some context of why we want to change this now?

LorenzoScebba:
> Hi @LorenzoScebba, it seems `/schemas/core-entities/plugins/{plugin}` and `/core-entities/schemas/plugins/{plugin}` both exists, otherwise the Konnect UI would break. Could you share some context of why we want to change this now?

Good catch — both do work today, but that's actually an accident of how the gateway routes these calls (it strips out core-entities from the path wherever it appears, so the segment order around it doesn't matter). /core-entities/schemas/plugins/{plugin} is the one that's officially documented, so this just makes the frontend call the real contract instead of relying on that routing quirk — no functional change, just future-proofing in case that gateway behavior ever changes.

https://github.com/kong-konnect/konnect-gateways-helm/blob/117c30d5afd66c9bcb643813dd09d84c5f5d2880/charts/kgateway/deck-configs/konnect-api.yaml#L109

|         | path                                                          |
|---------|---------------------------------------------------------------|
| correct | /v2/control-planes/abc-123/core-entities/schemas/plugins/cors |
| wrong   | /v2/control-planes/abc-123/schemas/core-entities/plugins/cors |

- correct: rgid=abc-123, path=schemas/plugins/cors
- wrong: rgid=abc-123/schemas, path=plugins/cors

Both rewrite to: /api/runtime_groups/abc-123/schemas/plugins/cors

CC @mayurm88 as original reported if any more detail is needed (and for fact-checking before merging)

Leopoldthecoder:
@LorenzoScebba Thank you for the details, makes great sense. I'm merging this PR.

Leopoldthecoder (approved):

PR #3137, opened 2026-04-16
feat(entities-plugins): custom plugin form API integration

2eha0:
# Summary
KM-2486


packages/entities/entities-plugins/src/components/CustomPluginForm.vue
TT1228, line 902: When i preview the form, `.kong-ui-entity-form-block .header-content` has a `padding-top: 4px` so it's not strictly aligned in center.

Image

TT1228 (commented):
A detail about texts here: some of them ends with period while others don't. Should "Inherit(s)" align with "Follows"?
Image

TT1228 (approved):
Tested and all the other part LGTM

PR #3139, opened 2026-04-16
fix: add preview flag to dashboards

filipgutica:
# Summary

Add a preview flag to force a dashboard to be readonly and have no "interactions"

### DIFF
--- a/packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
+++ b/packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
@@ -41,7 +41,7 @@
             :key="isFullscreen ? `${tile.id}-tile` : `${tile.id}-tile-fullscreen`"
             v-model:refresh-counter="refreshCounter"
             class="tile-container"
-            :context="internalContext"
+            :context="tileContext"
             :definition="tile.meta"
             :height="tile.layout.size.rows * (model.tile_height || DEFAULT_TILE_HEIGHT) + parseInt(KUI_SPACE_70, 10)"
             :hide-actions="preview"
@@ -145,12 +145,7 @@ const queryReady = computed<boolean>(() => {
   return !!context.timeSpec || !configStore.loading
 })
 
-const rendererContext = computed<DashboardRendererContext>(() => ({
-  ...context,
-  editable: preview ? false : context.editable,
-}))
-
-const isEditable = computed<boolean>(() => !!rendererContext.value.editable)
+const isEditable = computed<boolean>(() => !preview && !!context.editable)
 
 const tileSortFn = (a: TileConfig, b: TileConfig) => {
   const rowDiff = a.layout.position.row - b.layout.position.row
@@ -313,9 +308,15 @@ const globalFilters = computed<AllFilters[]>(() => {
 
 const { internalContext } = composables.useDashboardInternalContext({
   globalFilters,
-  context: rendererContext,
+  context: computed(() => context),
   isFullscreen,
 })
+
+const tileContext = computed(() => ({
+  ...internalContext.value,
+  editable: preview ? false : internalContext.value.editable,
+  zoomable: preview ? false : internalContext.value.zoomable,
+}))
 
 defineExpose({
   refresh: refreshTiles,
--- a/packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
+++ b/packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
@@ -22,7 +22,7 @@
         {{ i18n.t('renderer.noQueryBridge') }}
       </KAlert>
       <component
-        :is="isEditable && !isFullscreen ? DraggableGridLayout : GridLayout"
+        :is="internalContext.editable && !isFullscreen ? DraggableGridLayout : GridLayout"
         v-else
         ref="gridLayoutRef"
         :tile-height="model.tile_height"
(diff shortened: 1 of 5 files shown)


packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -120,8 +125,8 @@ composables.useRequestQueue()
 const configStore = useAnalyticsConfigStore()
 
 const timeSpec = computed<TimeRangeV4>(() => {
-  if (props.context.timeSpec) {
-    return props.context.timeSpec
+  if (context.timeSpec) {
```
Chalks: **blocker**

all references to `context` should be references to `internalContext` instead. Granted, this is an issue unrelated to your PR, but since you're touching them all... might as well.

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -137,9 +142,11 @@ const queryReady = computed<boolean>(() => {
   // We're ready to issue queries if we know the time spec.
   // We know the timespec if we were given the timespec, or if the config store has loaded the org's retention
   // and we're able to calculate a timespec.
-  return !!props.context.timeSpec || !configStore.loading
+  return !!context.timeSpec || !configStore.loading
 })
 
+const isEditable = computed<boolean>(() => !preview && !!context.editable)
```
Chalks: **blocker**

This should not exist. Use `internalContext.editable` instead. We want a single source of truth.

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -301,10 +308,16 @@ const globalFilters = computed<AllFilters[]>(() => {
 
 const { internalContext } = composables.useDashboardInternalContext({
   globalFilters,
-  context: toRef(props, 'context'),
+  context: computed(() => context),
   isFullscreen,
 })
```
Chalks: **blocker**

pass `preview` to this composable, and modify `editable` there instead of managing it in this file.

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -301,10 +308,16 @@ const globalFilters = computed<AllFilters[]>(() => {
...
 
+const tileContext = computed(() => ({
+  ...internalContext.value,
+  editable: preview ? false : internalContext.value.editable,
+  zoomable: preview ? false : internalContext.value.zoomable,
+}))
+
```
Chalks: **blocker**

This is a third context. :( This should not exist. Modify `internalContext` instead as suggested above.

Chalks (changes requested):
All of my requested changes below boil down to the same thing:

**There should be one source of truth for context.** We've already decided that `props.context` is not the source of truth because we introduced `internalContext`. So let's actually lean on `internalContext` to do the work.

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -119,25 +123,16 @@ composables.useRequestQueue()
 
 const configStore = useAnalyticsConfigStore()
 
-const timeSpec = computed<TimeRangeV4>(() => {
```
filipgutica, line 122: Removing this cause it's a duplicate. `useDashboardInternalContext` already has it: https://github.com/Kong/public-ui-components/blob/main/packages/analytics/dashboard-renderer/src/composables/useDashboardInternalContext.ts#L27

and exposes it as `internalContext.value.timeSpec`

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
filipgutica: done

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
filipgutica: done

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
filipgutica: done

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
filipgutica: done

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
Chalks: nice, I thought there might be something like that hanging out once I realized how much `useDashboardInternalContext` was doing. I'm glad you saw it.

packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -119,25 +123,16 @@ composables.useRequestQueue()
...
   // In the future, this will need to be determined on a per-tile basis to support pipelining.
   // For now, it's fine for it to only be global.
 
   // We're ready to issue queries if we know the time spec.
   // We know the timespec if we were given the timespec, or if the config store has loaded the org's retention
   // and we're able to calculate a timespec.
-  return !!props.context.timeSpec || !configStore.loading
+  // Note: this checks the *caller-supplied* timeSpec, not internalContext.timeSpec, because the latter
+  // always has a fallback value even before configStore finishes loading.
+  return !!context.timeSpec || !configStore.loading
```
Chalks: **question**

Appreciate this comment (and it's sufficient for me to not block on this), but could we either:

1. move the entire `queryReady` into `internalContext` or
2. change this logic to something like `configStore.loading ? !!context.timeSpec : internalContext.timeSpec` (I think that's the same idea, but I'm not 100% sure)

I'm not sure which makes more sense, especially with the goal of not blowing up the scope on this PR too much.

Chalks (commented):
just one non-blocking question left.

Chalks (approved):

PR #3147, opened 2026-04-18
chore(entities-plugins): metering&billing improvements

TT1228:
# Summary

- Support array of expected values in render-rules dependencies (['query', 'header'] = any-of). Introduce a new Symbol `RENDER_RULE_EXACT_MATCH` for explicitly stating that an array in render-rules is to be exactly matched 
- Prefill ingest_endpoint from geoApiServerUrl region on create; fall back to https://{{region}}.api.konghq.com/v3/openmeter/events
- Hide subject field unless look_up_value_in is query or header
- Fix "Metering&Billing" → "Metering & Billing" in locale
- Add component tests for ingest_endpoint prefill behavior

### DIFF
--- a/packages/entities/entities-plugins/src/components/free-form/plugins/metering-and-billing/MeteringAndBillingForm.vue
+++ b/packages/entities/entities-plugins/src/components/free-form/plugins/metering-and-billing/MeteringAndBillingForm.vue
@@ -1,6 +1,6 @@
 <template>
   <StandardLayout
-    v-bind="props"
+    v-bind="{ ...props, model: modelWithDefaults }"
     :plugin-config-description="t('plugins.free-form.metering-and-billing.sections.plugin_config.description')"
   >
     <template #field-renderers>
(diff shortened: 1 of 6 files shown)


packages/entities/entities-plugins/src/components/free-form/plugins/metering-and-billing/MeteringAndBillingForm.vue
```diff
@@ -140,6 +140,11 @@
       <ObjectField
         :label="t('plugins.free-form.metering-and-billing.sections.subject.title')"
         name="config.subject"
+        :render-rules="{
+          dependencies: {
+            field: ['look_up_value_in', ['query', 'header']],
```
2eha0, line 145: I am not sure if this is straightforward enough.
Let's say we have an array field, and we want to set a rule for it when it equals another array,
In that case we might need to design a new interface, but that case is pretty rare honestly

2eha0 (approved):
LGTM

PR #3156, opened 2026-04-21
fix(entities-plugins): prefill default values with prepareFormData

TT1228:
# Summary

Replace `modelWithDefaults` with `formConfig` containing `prepareFormData` hook to correctly initialize metering&billing form initial value and inject geo based `ingest_endpoint` properly. `modelWithDefaults` bypasses `getDefaultFromSchema `


2eha0 (approved):
LGTM

PR #3173, opened 2026-04-23
fix(entities-plugins): set allow/deny to null when toggling

TT1228:
# Summary

Set `allow`/`deny` to `null` on toggle so that for `PATCH` request, `allow`/`deny` can be unset.


Leopoldthecoder (approved):
LGTM

sumimakito (approved):

hangrao (approved):

PR #3174, opened 2026-04-23
feat(analytics-chart): enhance donut chart visuals and add center metric

sichvoge:
Focus for this PR is to improve our current donut chart visualization. The PR adds the following modifications:
- Add showCenterMetric prop to display summed total in donut chart center
- Only renders for summable metric units (count, requests, usd, token count)
- Add isSummableMetricUnit() utility and SUMMABLE_METRIC_UNITS set
- Improve donut segment styling: explicit 65% cutout, white border, hover offset
- Add DonutChart.cy.ts component tests
- Align center metric label format with SingleValue for consistency

Before

image

After

image


packages/analytics/analytics-chart/src/components/chart-types/DonutChart.vue
```diff
@@ -121,15 +136,38 @@ const formattedDataset = computed<DonutChartData[]>(() => {
...
+    hoverBorderColor: [],
+    hoverBorderWidth: 3,
+    data: [],
+    hoverOffset: 10,
+  })
 
   return [formatted]
 })
 
+const { formatUnit } = unitFormatter({ i18n })
```
filipgutica, line 157: I think we should use our common unit formatter from `analytics-utilities` 

See https://github.com/Kong/public-ui-components/blob/main/packages/analytics/analytics-utilities/src/utils/formatUnit.ts

filipgutica (changes requested):
One minor comment lgtm otherwise

packages/analytics/analytics-chart/src/components/chart-types/DonutChart.vue
sichvoge: We are already using unitFormatter from @kong-ui-public/analytics-utilities, same pattern as SingleValue or TopNTable.

Example from [SingeValue L90](https://github.com/Kong/public-ui-components/blob/main/packages/analytics/analytics-chart/src/components/chart-types/SingleValue.vue#L90)

Is there anything I am missing?

filipgutica (approved):

PR #3194, opened 2026-04-27
fix: bumping actions [KHCP-20416]

ValeryG:
# Summary

last bunch of actions with node20 warnings

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -354,7 +354,7 @@ jobs:
           path: packages/
 
       - name: Run Component Tests
-        uses: cypress-io/github-action@248bde77443c376edc45906ede03a1aba9da0462 # v5.8.4
+        uses: cypress-io/github-action@783cb3f07983868532cabaedaa1e6c00ff4786a8 #v7.1.9
         with:
           install: false
           command: pnpm run test:component:ci --spec ${{ matrix.container }}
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-on-close.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Remove preview consumption comment
-        uses: marocchino/sticky-pull-request-comment@0ea0beb66eb9baf113663a64ec522f60e49231c0 # v3.0.4
+        uses: Kong/sticky-pull-request-comment@ae583cebedbfd32256b349cce8ddf5772e77636e # v3.0.4
```
johncowen, line 15: Q: How come we are changing this to a fork? Is this intentional?

ValeryG: yes it was intentional. Since we have the forks - we should use it :)

adamdehaven (approved):

.github/workflows/ci.yaml
```diff
@@ -413,7 +413,7 @@ jobs:
 
       - name: Download build artifacts
         if: ${{ github.event_name == 'push' }}
-        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
```
johncowen, line 416: I noticed this is downgrading, is this intentional? If so could you let me know the reason for the downgrade?

PR #3202, opened 2026-04-29
fix(entities-plugins): filter scoped plugins in featured plugin list

TT1228:
# Summary

For `PluginCatalog` under a certain scope, add filtering for `Featured` category so that it filter out plugin unavailable under this scope.

### DIFF
--- a/packages/entities/entities-plugins/src/components/PluginCatalog.vue
+++ b/packages/entities/entities-plugins/src/components/PluginCatalog.vue
@@ -417,11 +417,13 @@ const buildPluginList = (): PluginCardList => {
       return list
     }, {})
   // Pick highlighted plugin objects from pluginMetaData and assign to 'Featured'
+  // Only include plugins that passed scope filtering (i.e., already present in the list)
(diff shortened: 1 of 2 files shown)


packages/entities/entities-plugins/src/components/PluginCatalog.vue
```diff
@@ -431,7 +433,10 @@ const buildPluginList = (): PluginCardList => {
           group: meta.group || PluginGroup.CUSTOM_PLUGINS,
         }
       })
-      .filter(Boolean) // remove undefined if any id not found in meta
+      .filter(Boolean)
```
kingyue737, line 436: This line seems to be dead code. elements in the array returned from the previous `flatMap` are always truthy.

kingyue737 (approved):
Manually tested and LGTM

packages/entities/entities-plugins/src/components/PluginCatalog.vue
TT1228: Agreed, i will remove it in future PR

PR #3217, opened 2026-05-04
feat: remove MCP types in favor of Agentic

alxgrant:
# Summary


adorack (approved):
This seems fine, but please mark it as a breaking change (since you're removing an export).

erichsend (approved):

PR #3198, opened 2026-04-28
fix: vue-router semver range

schogges:
I think we were a little too fast upgrading to `vue-router@>5`. [Since the current version of `kongponent` didn't upgrade the peer of `vue-router` to `>5` ](https://github.com/Kong/kongponents/blob/7b1ff958b085f3390219111ac98ed2a5f6c9a74e/package.json#L80) the bump here introduces a lot more unmet peers. Additionally dependents are now forced to upgrade to `vue-router@^5.0.6` otherwise they can't upgrade the versions of their dependencies sourced from this repo without introducing conflicts.

I've kept `vue-router@^5.0.6` in the peer dependency semver range because as of https://router.vuejs.org/guide/migration/v4-to-v5.html#For-Vue-Router-4-Users-without-file-based-routing-
>No breaking changes.

So keeping the new major in the range allows smooth upgrading and as soon as `kongponents` includes `v5` in the peer semver range the dev version here can be bumped to `v5` too.

```
$ pnpm install --resolution-only

...

 WARN  Issues with peer dependencies found
.
├─┬ @kong/kongponents 9.52.11
│ └── ✕ unmet peer vue-router@^4.6.4: found 5.0.6
└─┬ stylelint-config-recommended-vue 1.6.1
  └─┬ stylelint-config-recommended 14.0.1
    └── ✕ unmet peer stylelint@^16.1.0: found 17.8.0

packages/analytics/analytics-chart
└─┬ @kong/kongponents 9.52.11
  └── ✕ unmet peer vue-router@^4.6.4: found 5.0.6

packages/analytics/analytics-geo-map
└─┬ @kong/kongponents 9.52.11
  └── ✕ unmet peer vue-router@^4.6.4: found 5.0.6

packages/analytics/analytics-metric-provider
└─┬ @kong/kongponents 9.52.11
  └── ✕ unmet peer vue-router@^4.6.4: found 5.0.6

packages/core/app-layout
├─┬ @kong/kongponents 9.52.11
│ └── ✕ unmet peer vue-router@^4.6.4: found 5.0.6
└─┬ focus-trap-vue 4.1.0
  └── ✕ unmet peer focus-trap@^7.0.0: found 8.1.0

packages/core/documentation
└─┬ @kong/kongponents 9.52.11
  └── ✕ unmet peer vue-router@^4.6.4: found 5.0.6

packages/core/entities-config-editor
└─┬ @kong/kongponents 9.52.11
[...]

PR #3209, opened 2026-04-30
feat: add table-data-grid

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-5001
https://konghq.atlassian.net/browse/MA-5084

image

Adds `@kong-ui-public/table-data-grid`, a Vue 3 wrapper around AG Grid for reusable table-style data grids in public UI components.

The package gives host apps a typed boundary for row fetching, column-key slots, persisted table configuration, filters/search, selection, table state, and toolbar composition without requiring each consumer to wire AG Grid directly.

## What changed

- Adds the `packages/core/table-data-grid` package with `TableDataGrid`, `TableDataGridColumnVisibilityMenu`, public types, table preference interop helpers, sandbox, README, and architecture documentation.
- Supports server-driven pagination and infinite loading through one fetcher contract, including sort, filter, search, page/block size, cursor, known-total, and unknown-total flows.
- Adds controlled or internal `tableConfig` state for column order, visibility, widths, pinning, sorting, and page size, with conversion helpers for Kongponents table preferences.
- Adds column definition, sizing, fit-to-width, resize, pin, move, visibility, and layout-sync behavior while preserving host/user-defined widths.
- Adds row click, cell click, single-row selection, multi-row selection, selected-row bulk action slots, and disabled row-click columns for action cells.
- Adds header-driven `KFilterGroup` filters, host-owned custom filter slots, opt-in debounced search, toolbar slots, outside search/filter/action targets, and column visibility controls.
- Adds row and cell attribute hooks for test ids, accessibility, and integration-specific metadata.
- Adds loading, empty, and error state surfaces plus the `state` event for host integrations.
- Adds AG Grid Community/Vue dependencies as package dependencies so consumers only need the documented peer dependencies.

## Reviewer notes

- Start with `packages/core/table-data-grid/docs/architecture.md` for ownership boundaries, dataflow, fetch lifecycle, config lifecycle, renderer context, and verification focus.
- Use `packages/core/table-data-grid/README.md` for public API review: examples, props, models, events, slots, exposed methods, and exported types.
- The main review boundaries are:
  - public API and examples: `src/types/index.ts`, `TableDataGrid.vue`, README
  - fetch behavior: `useTableDataGridPaginationFetch.ts`, `useTableDataGridInfiniteFetch.ts`, `src/utils/fetchers.ts`
[...]

PR #3220, opened 2026-05-06
chore(linting): add eslint-package-constraints to lint dependency configuration

johncowen:
**DO NOT MERGE!!!111** (we need to decide how to publish/consume the lint plugin included possibly temporarily here)

# Summary

- Adds Kuma's `eslint-package-constraints` from https://github.com/kumahq/kuma-gui/tree/master/packages/eslint-package-constraints. Via "poor mans publish" for the moment (potentially to be published normally before merge)
- Uses `pnpm run lint` to lint the majority of `package.json` files
- In using the linter, we fix some consequential/incorrect peerDependency/kongponents configuration which previously was using exact dependency constraints, and now is changed to use caret `^` range constraints.

## What / Why

I've had several experiences where our team has been blocked do to un-installable packages. Very recently in https://github.com/Kong/public-ui-components/pull/3127 and relatedly in https://github.com/Kong/public-ui-components/pull/3198.

Packages were un-installable due to incorrect configuration of dependencies (i.e. exact constraint instead of a range/caret constraint). These are generally just little mistakes people naturally make all the time when add a dependency via `pnpm`. Depending on several things sometimes this is just a CI failure over in our application followed by a PR to correct the issue, in these cases everyone agrees its just a little mistake that needs amending. This is the perfect use-case for a linter. Let's use a linter to "police" this so humans don't have to.

`eslint-package-constraints` uses standard JSON schema validation to validation the versions used in package.json files to ensure they are using recommended ranges (depending on the type of dependency, this is mostly caret/`^` based). If someone mistakenly doesn't use a caret range, then the linter will tell them before they even submit a PR. This makes it easy to do the right thing, and hard to do the wrong thing.

Obviously sometimes we cannot do the "ideally right" thing, and pragmatically we need to do the "wrong thing" but being pragmatic this end up being the "right thing" in these cases (yeah my brain is exploding right now also 😅 but hopefully you know what I mean).

In these cases, usually you use eslint ignores that can be validated by the reviewer to ensure we are ignoring for valid edge cases. Thing is, for JSON files, there is no way to ignore things inline in the JSON file.

`eslint-package-constraints` includes a centralized configuration that includes a method of ignoring/changing certain packages from its default constraints. Handily, this PR includes an example: I've allowed the `swagger-ui-web-component` package to use exact-pinned react related packages as I'm aware that this is a bit of a can of worms to sort out.

### Further Notes

- I found that linting does not consider anything in the root of the monorepo
- I found that the `swagger-ui-web-component` package has no linting, I've included a lint script here
- I've (perhaps temporarily) included the source for Kuma's `eslint-package-constraints` here. This lets reviewers see what's happening and also helps me start a discussion on what people would prefer for consuming this code. Kuma doesn't _yet_ publish this to a registry.


packages/analytics/analytics-config-store/package.json
```diff
@@ -40,7 +40,7 @@
   },
   "devDependencies": {
     "@kong-ui-public/analytics-utilities": "workspace:^",
-    "pinia": ">= 3.0.4 < 4",
+    "pinia": "^3.0.4",
```
johncowen, line 43: I found this to be a little weird, to my understanding `>= 3.0.4 < 4` is just `^3.0.4` I think? Maybe I missed something. I took a look back through the history of this change and couldn't find a specific reason why this less obvious range was used, so I updated it to use `^`.

Let me know if I missed something.

package.json
```diff
@@ -64,7 +64,7 @@
     "tslib": "^2.8.1",
     "typedoc": "^0.28.19",
     "typedoc-plugin-markdown": "^4.11.0",
-    "typescript": "~5.9.3",
+    "typescript": "^5.9.3",
```
johncowen, line 67: Again this one was a little weird but less so that the one I commented on above.

Again, I took a look through the history and I cannot find any reason why this was added as `~` and not `^`. I think this one was just a little mistake, and I'm glad we have a linter for this.

I updated it to use `^` but let me know if I missed something.

adamdehaven (changes requested):
T hi is would be better handled by Renovate config if we wanted to implement

adamdehaven:
> I've had several experiences where our team has been blocked do to un-installable packages

This is due to subjective settings and enforcement on the consuming repository. 

npm, yarn, and pnpm all handle this properly, but the upstream repository in your case has chosen to enforce specific dependency rule sets. 

This repository _should_ lean on version ranges for **most** dependencies, but should not be precluded from utilizing a pinned dependency when warranted. 

It would be better for us to update our Renovate settings to prefer version ranges over pinned dependencies if we wanted to go this route. 

There have been discussions from the security team around preferring pinned dependencies; however, as you show here that likely doesn't make sense for shared component libraries as it offers no benefit around vulnerabilities from transitive dependencies in most cases anyway.

adamdehaven:
One more piece of info from Renovate directly:

> Browser or dual browser/node.js libraries that are consumed/`required()`'d by others should keep using SemVer ranges for `dependencies` but can use pinned dependencies for `devDependencies` ([source](https://docs.renovatebot.com/dependency-pinning/#so-whats-best))

adamdehaven:
@ValeryG my suggestion would be to evaluate enabling `pinonlydevdependencies` via our Renovate config https://docs.renovatebot.com/presets-default/#pinonlydevdependencies

via https://github.com/Kong/public-shared-renovate/pull/76

johncowen:
Hey @adamdehaven , thanks for the response!

Note: I think the issues here are not just regarding `devDependencies`, so shall we put `devDependencies` aside from the discussion for the moment? Are you ok just concentrating on incorrectly configured `dependencies` for now? (but glad you are thinking of devDeps also!)

> T hi is would be better handled by Renovate config if we wanted to implement

Q: Would Renovate prevent people from merging (and subsequently releasing on merge) incorrectly configured dependency version constraints, that then block consuming repositories from installing? 

> This is due to subjective settings and enforcement on the consuming repository.

In your first comment you seem to agree that we should "handle" "something" with Renovate? So I'm not sure if that's an agreement that it is an issue we can solve in _this_ repository and _not_ in the consuming repository. If you think there is something we can do in _this_ repository it would be great if we could work together to improve things here.

Q: Could you clarify whether you think there is an issue we can solve here in _this_ repository?

adamdehaven:
> Could you clarify whether you think there is an issue we can solve here in _this_ repository?

I'm open to the Renovate config change linked above (this would apply to all of our FE repositories). This would not prevent explicit dependency changes (e.g. if someone pushes a PR) but this is by design -- there are valid cases for pinning `dependencies` and this (and other) repositories should not be prevented from doing so.

packages/analytics/analytics-config-store/package.json
Chalks: [it was originally added here](https://github.com/Kong/public-ui-components/pull/1208/changes#diff-7a9c79dc13712962a176113be00d7481a2b5c072c4984cb186e78b490a1b949b) and as far as I can tell your conclusion is correct.

johncowen:
Ok so, just a one off comment/confirmation so we can move on.

---

It sounds like we all agree there is a problem of incorrect dependency configurations being made that have negative downstream repercussions? Specifically: people accidentally using exact constraints (rather than range constraints)  in repositories containing consumable dependencies - these can then block downstream applications from being installed correctly.

Additionally, it sounds like we all agree that the problem should be solved in the repositories that _provide_ the dependencies rather than the applications that are _consuming_ the dependencies?

Q: Do we agree on the above points?

In turn, I agree that there are times when people do actually need to exact pin dependencies and they should not be prevented from doing so if they absolutely need to.

packages/analytics/analytics-config-store/package.json
johncowen: hey there👋 , thanks for the confirmation and that link 👍

adamdehaven:
> It sounds like we all agree there is a problem of incorrect dependency configurations being made that have negative downstream repercussions? Specifically: people accidentally using exact constraints (rather than range constraints) in repositories containing consumable dependencies - these can then block downstream applications from being installed correctly.

Again, no, this isn't a blanket "correct" response. There are valid use cases for pinning exact dependencies, e.g. if we depend on an exact version of `X` because of features, interfaces, security, etc.

> Additionally, it sounds like we all agree that the problem should be solved in the repositories that _provide_ the dependencies rather than the applications that are _consuming_ the dependencies?

It depends. Again, a downstream repository may have its own rules that are too strict, and obviously these are unknown to the component library as an example.

johncowen:
@adamdehaven would you be able to join me on a zoom early next week to discuss further?

adamdehaven:
> @adamdehaven would you be able to join me on a zoom early next week to discuss further?

I'll be out of office until the week of May 26

johncowen:
I'm happy to wait until then, could we organise a call for when you are back?

johncowen:
Hey @adamdehaven I'd like to pick this PR back up again if possible.

Just to summarise what this PR is doing to refresh our memories and clear up any potential misunderstanding:

- Uses a eslint lint rule (via a JSON schema rule) to enforce "correct" version constraints _before_ merging or even pushing a PR and consequently publishing a package with "incorrect" version constraints.
- Uses _range constraints_ (specifically caret `^`) for `dependencies` and `devDependencies`, allows wider more varied ranges for `peerDependencies`. **Importantly, this PR _prevents_ people from _exact pinning_ _any_ dependencies.**, whether they are `dependencies`, `devDependencies` or `peerDependencies`.
- Allows people to override these "default _range constraints_" per dependency _when warranted_. This can be done in a place where it is possible to leave inline code comments explaining The Why (i.e. this config is not a JSON file)

Please let me know if there are any questions or if anything isn't clear. I'm not sure if its still necessary, but I'm happy to jump on a zoom call if any of the above isn't clear or you disagree with the approach here. Just let me know!

I'd like to get this approach merged as soon as possible, so if you have any concerns over the "range constraints everywhere" approach then please let me know so we can discuss.

If we can agree on the above "range constraints everywhere" then we just need to discuss how to provide the `eslint-package-constraints` plugin that is currently just copy/pasta'ed into this PR.

Lmk!

ValeryG:
1. developers should not be editing package.json - everybody needs to use pnpm command to add/update e.t.c packages.
2. Assuming 1 is valid - this lint is auditing the result of pnpm command..

Q: How are we sure that lint is smarter than pnpm command itself?  :)

Q: Are there any recommendations / blogposts or anything in the internet that recommends or discusses package.json lining? I never saw anything like that, would be curious.

johncowen:
Hey @ValeryG 👋 

Good questions! Let me see if I can answer them briefly!

> developers should not be editing package.json - everybody needs to use pnpm command to add/update e.t.c packages.

I'm not sure how exact/pinned versions are getting into the package.json file, I'm only aware of the result and I think we can maybe agree judging by the amount of changes made to package.json files here in this PR (and the PRs I mention in the description) that they _are_ getting into the package.json files.

Is that something we can agree on? That "incorrect"" constraints _are_ finding their way into package.json files?

---

> Assuming 1 is valid - this lint is auditing the result of pnpm command..
> Q: How are we sure that lint is smarter than pnpm command itself? :)

I guess "Assuming 1 is valid" would not explain why there are a lot of exact pinned dependencies being changed in this PR and PRs like the ones mentioned in the description happen.

But, lets assume for the moment that pnpm _is_ used and the resulting package.json _is_ correct. In this case the linter does nothing, everything is correct and the PR is created, nobody even notices. The linter (just like _all_ other linters we currently use) only kicks in when somebody makes a mistake to help them correct it. We don't not have linters because they only occasionally kick in.

But, lets not lose sight of the fact that this assumption still doesn't explain why we often encounter issues with incorrect constraints being use, the changes in this PR nor the PRs linked in the description of this PR.

---

> Q: How are we sure that lint is smarter than pnpm command itself? :)

I guess I'm not saying that lint is smarter than pnpm itself! :)

What I am saying is that exact pinned versions of dependencies are being added probably by accident. This linter catches those when necessary, and for whatever reason the current approach (whatever that is) doesn't.

---

> Q: Are there any recommendations / blogposts or anything in the internet that recommends or discusses package.json lining? I never saw anything like that, would be curious.

This plugin includes `eslint-plugin-json-schema-validator` which seems pretty supported and popular and this plugin is specifically for linting json files.

As for specifically linting `package.json` files themselves, then I guess most people don't have such a large problem of accidentally specifying "incorrect" version constraints. When they do encounter it, they are just corrected with no additional ceremony. Personally, this is the only time (working at Kong) that I've experienced this being a problem to such a large extent.

For example here's one I saw only today in a pretty popular package: https://github.com/vitejs/vite-plugin-vue/pull/776

The PR is changing an exact pinned dependency to a caret range. Note the PR descriptions says:

> So that the package manager can dedupe

The issue was spotted, the correction was made the PR was merged. Personally, as I've seen this here to such a large extent I'd rather we had a linter so people can "police" this themselves rather than only my two man team doing it which is what has been happening in the past.

Maybe you are referring to the versioning approach? I'm aware of a few articles around that do agree with the versioning approach used here in this PR (use caret ranges by default), and very many many repositories. Maybe to use your pnpm question as an example:

As far as I'm aware `pnpm`, `npm` and `yarn` all default to using a caret range constraint when adding dependencies if you don't specify otherwise. 

---

Ok, those where good questions and as always I'm glad you are asking, so thankyou. Let me know if you still have questions or I've not explained sufficiently.

In return could I ask one additional question of my own:

> Uses range constraints (specifically caret ^) for dependencies and devDependencies, allows wider more varied ranges for peerDependencies. Importantly, this PR prevents people from exact pinning any dependencies., whether they are dependencies, devDependencies or peerDependencies.

Do you have any problems with us following the above approach (i.e. use caret ranges for everything by default, like pnpm)? If you do, could you explain what they are so we can work through them/any issues you have. Thanks!

adamdehaven:
Also, please see my comments further up in this discussion. There are valid reasons for pinning dependencies and this repository, just like any other, should not be prevented from doing so.

johncowen:
> There are valid reasons for pinning dependencies and this repository, just like any other, should not be prevented from doing so.

Agree 👍 . This PR allows exceptions.

The default should be caret range, like `pnpm`.

johncowen:
👋 @adamdehaven @ValeryG just checking in to see if there is any update here? It sounds like we might be on the same page?

- We should always use pnpm (so it defaults to a caret `^` range) for adding new dependencies.
- People shouldn't be prevented from changing this to a different pinning approach when warranted.

Does that sound right?

adamdehaven:
I can take another look at the implementation when I'm back but at this time I'm still not in favor of implementing this.

johncowen:
Just linking another example here of where using a linter to enforce caret ranges would have been an advantage (thanks for picking up on that in the PR @adamdehaven 👍 ).

https://github.com/Kong/public-ui-components/pull/3209#discussion_r3195718661

Screenshot 2026-06-09 at 10 58 18

I guess person submitting the PR didn't use `pnpm` to add this dependency?

P.S.

Ah another in the same PR:

https://github.com/Kong/public-ui-components/pull/3209#discussion_r3327130681

Screenshot 2026-06-09 at 11 02 26

---

I'm kinda assuming that we all agree on these considering the above screengrabs?

> We should always use pnpm (so it defaults to a caret ^ range) for adding new dependencies.
> People shouldn't be prevented from changing this to a different pinning approach when warranted.

I'm assuming that the issue folks have with my PR here is just that it adds an existing 3rd party linter that nudges people to do this automatically?

Is that that right assumption to make here?

(btw: if anyones interested, author of above PR asked folks to review his PR in Slack this morning, hence me noticing the above this morning)

johncowen:
hi @adamdehaven, in relation to https://github.com/Kong/public-ui-components/pull/3220#issuecomment-4633726646 from 2 weeks back. Have you had any more thoughts on this?

PR #3227, opened 2026-05-07
Feat/donut chart center metric default

sichvoge:
# Summary

The donut chart's center metric was previously hidden by default and only          supported a limited set of unit-based checks to determine whether a total could be displayed. 

This PR addresses both issues.

Changes:
  - showCenterMetric now defaults to true                                            
  - Adds isSummableMetricName to handle metrics where the unit alone isn't
  sufficient to determine summability — specifically size metrics (request_size_sum, response_size_sum, etc.) where all aggregations (sum, avg, percentiles) share the same bytes unit. Supports both _sum (Druid) and :sum (GOAP) naming conventions
  - Expands SUMMABLE_METRIC_UNITS with platform entity count units (service_count, route_count, etc.) which use their metric name as the unit string
  - Removes the dead requests entry from SUMMABLE_METRIC_UNITS — the API never returns this string as a unit; it was an i18n display artifact from old mock data
  - Passes the metric name alongside the metric unit from AnalyticsChart down to DonutChart to enable the name-based summability check

### DIFF
--- a/packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
+++ b/packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
@@ -110,7 +110,7 @@ import DonutChart from './chart-types/DonutChart.vue'
 import { computed, provide, toRef } from 'vue'
 import { msToGranularity } from '@kong-ui-public/analytics-utilities'
 import type { AbsoluteTimeRangeV4, ExploreAggregations, ExploreResultV4, GranularityValues } from '@kong-ui-public/analytics-utilities'
-import { hasMillisecondTimestamps, defaultStatusCodeColors, isUnitlessMetricUnit } from '../utils'
+import { hasMillisecondTimestamps, defaultStatusCodeColors, isNoSuffixMetric } from '../utils'
 import TimeSeriesChart from './chart-types/TimeSeriesChart.vue'
 import { KUI_COLOR_TEXT_WARNING, KUI_ICON_SIZE_40 } from '@kong/design-tokens'
 import { WarningIcon } from '@kong/icons'
@@ -248,7 +248,7 @@ const tooltipMetricDisplay = computed<string | undefined>(() => {
   }
 
   // @ts-ignore - dynamic i18n key
-  if (isUnitlessMetricUnit(metricUnit) && i18n.te(`metricAxisTitles.${metricName}`)) {
+  if (isNoSuffixMetric(metricUnit) && i18n.te(`metricAxisTitles.${metricName}`)) {
     // @ts-ignore - dynamic i18n key
     return i18n.t(`metricAxisTitles.${metricName}`) || undefined
   }
@@ -281,8 +281,8 @@ const metricAxesTitle = computed<string | undefined>(() => {
   }
 
   // @ts-ignore - dynamic i18n key
-  if (i18n.te(`metricAxisTitles.${metricName}`) && (isUnitlessMetricUnit(metricUnit) || i18n.te(`chartUnits.${metricUnit}`))) {
-    if (isUnitlessMetricUnit(metricUnit)) {
+  if (i18n.te(`metricAxisTitles.${metricName}`) && (isNoSuffixMetric(metricUnit) || i18n.te(`chartUnits.${metricUnit}`))) {
+    if (isNoSuffixMetric(metricUnit)) {
       // @ts-ignore - dynamic i18n key
       return i18n.t(`metricAxisTitles.${metricName}`) || undefined
     }
--- a/packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
+++ b/packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
@@ -89,6 +89,7 @@
         :dataset-colors="chartOptions.chartDatasetColors || defaultStatusCodeColors"
         :legend-position="legendPosition"
         :legend-values="legendValues"
+        :metric-name="computedMetricName"
         :metric-unit="computedMetricUnit"
         :show-center-metric="chartOptions.showCenterMetric"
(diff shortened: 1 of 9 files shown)


packages/analytics/analytics-chart/src/utils/metric-units.ts
```diff
@@ -14,11 +14,22 @@ export const isUnitlessMetricUnit = (unit: string): boolean => {
...
+  'consumer_count',
+  'plugin_count',
+  'node_count',
 ])
 
 export const isSummableMetricUnit = (unit: string): boolean => {
   return SUMMABLE_METRIC_UNITS.has(unit)
 }
+
+// Unit-based checks don't work for size metrics since all aggregations (sum, avg, percentiles) share
```
sichvoge, line 6: @filipgutica I wasn't sure if a comment is usually allowed because I don't see it very often but thought that it is not obvious why this function actually exists. Hence, I added it just in case. Let me know if thats ok.

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
```diff
@@ -89,6 +89,7 @@
         :dataset-colors="chartOptions.chartDatasetColors || defaultStatusCodeColors"
         :legend-position="legendPosition"
         :legend-values="legendValues"
+        :metric-name="computedMetricName"
         :metric-unit="computedMetricUnit"
         :show-center-metric="chartOptions.showCenterMetric"
```
filipgutica, line 94: This was missed in the review when the center metric was first introduced. Need to add the showCenterMetric property to the chart options interface.

https://github.com/Kong/public-ui-components/blob/main/packages/analytics/analytics-chart/src/types/chart-data.ts#L73

packages/analytics/analytics-chart/src/utils/metric-units.ts
```diff
@@ -14,11 +14,22 @@ export const isUnitlessMetricUnit = (unit: string): boolean => {
...
+// the same 'bytes' unit. Check the metric name instead — both _sum (Druid) and :sum (GOAP) notations.
+export const isSummableMetricName = (name: string): boolean => {
+  return name.endsWith('_sum') || name.endsWith(':sum')
+}
```
filipgutica: I think we can just collapse these into one function called `isSummableMetric(name, unit)`

filipgutica (changes requested):

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
sichvoge: It's already there: https://github.com/Kong/public-ui-components/blob/main/packages/analytics/analytics-chart/src/types/chart-data.ts#L118 or do you mean something different?

packages/analytics/analytics-chart/src/utils/metric-units.ts
sichvoge: Thanks for the suggestion! I looked into collapsing these into a single `isSummableMetric(name, unit)` but there's a constraint that makes it tricky. [Line 165](https://github.com/Kong/public-ui-components/blob/main/packages/analytics/analytics-chart/src/components/chart-types/DonutChart.vue#L165) uses a unit-only check inside the `translateUnit` callback (which only receives `unit`, not the metric name) to suppress the unit label from the grand total. This keeps the display consistent for all count-type metrics, e.g.         showing "1.2K" instead of "1.2K requests". If I used the combined function there, size sum metrics would match via the `_sum` check and incorrectly suppress "1.2   MB". So I'd still need `SUMMABLE_METRIC_UNITS` accessible at that call site anyway. Feels cleaner to just keep two small focused functions than to export an   internal set alongside a combined one.

Also noticed that `count` and `token count` aren't in `UNITLESS_METRICS`, which makes sense because we do want "requests" and "tokens" showing up in tooltips and legend values. But it does mean the definition of "unitless" is a bit inconsistent right now. I think it's worth revisiting how we define unitless vs. how we control tooltip label display more broadly, but that feels like a separate PR.

filipgutica: Ahh I don't think that you need that isSummableMetric check, the translateUnit composable should do this work internally.

I would push to making it more consistent. This is my suggestion: 

- Remove the `isSummableMetric` check on [L165](https://github.com/Kong/public-ui-components/blob/main/packages/analytics/analytics-chart/src/components/chart-types/DonutChart.vue#L165) 
- add more metrics such as `request_count` to `UNITLESS_METRICS` since we already display the metric in the tooltip and don't need to show `n requests` repeatedly
- Consolidate to one `isSummableMetric` function
Image


Let me know if you think we can make this change, I know it's more broad but it keeps the logic simpler and more centralized

packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
filipgutica: Oh my bad. It was late, I was looking at old code when comparing, disregard my comment 😅

packages/analytics/analytics-chart/src/utils/metric-units.ts
sichvoge: Should we also add token counts to unitless?

filipgutica: Yeah I think we can add all those to unitless metrics . And may be we name it  "no unit suffix metrics" or something so it's more clear?

sichvoge: As discussed in Slack. We changed it as per Filips suggestion. I had to also fix a couple tests now that we renamed and moved a few functions.

filipgutica (approved):

PR #3235, opened 2026-05-08
feat(entities-plugins): move editor mode switch to StandardLayout

sumimakito:
Moving the editor mode (form vs. code) to StandardLayout so that the switch can be enabled to more plugins.

KM-2537


2eha0 (approved):
LGTM

jw-foss (approved):

hangrao (approved):

PR #3240, opened 2026-05-11
feat(entities-shared): add a secret advice to the decK customization editor

sumimakito:
Adding a secret advice to the decK command editor, as it may contain sensitive fields:

Screenshot 2026-05-11 at 14 55 25

### DIFF
--- a/packages/entities/entities-shared/src/components/common/DeckCodeBlockInternal.vue
+++ b/packages/entities/entities-shared/src/components/common/DeckCodeBlockInternal.vue
@@ -87,23 +87,27 @@
       @code-block-render="highlightCodeBlock"
     />
 
-    <component
-      :is="DeckCommandEditor"
-      v-else-if="props.entityRecord && props.isCustomizing && DeckCommandEditor"
-      v-model="unredactedDeckCommand"
-      :language="shell"
-    />
+    <template v-else-if="props.entityRecord && props.isCustomizing && DeckCommandEditor">
+      <div class="customization-secret-advise">
+        {{ t('deckCodeBlock.customization.secret_advise') }}
+      </div>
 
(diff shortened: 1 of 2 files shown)


kingyue737 (approved):

Leopoldthecoder:
Before this change, the second `KAlert` will show when `props.entityRecord` or `DeckCommandEditor` is falsy: https://github.com/Kong/public-ui-components/blob/f66019ae33d57a2ac1814a546998bdd53389cf82/packages/entities/entities-shared/src/components/common/DeckCodeBlockInternal.vue#L90-L106

After this change, the second `KAlert` will only show when `props.entityRecord` and `DeckCommandEditor` are both truthy. Is it intentional?

sumimakito:
@Leopoldthecoder Yes, the 2nd KAlert ("Changes made here…") is for customization only. Previously, it was controlled separately from the DeckCommandEditor component. When the `isCustomizing` is truthy but the DeckCommandEditor is missing, the KAlert will be displayed without the editor or the KCodeBlock, which does not make much sense. 

The latest change makes the secret advice, editor, and the footer KAlert controlled by the same condition, which ensures they appear or not at the same time.

hangrao (approved):

Leopoldthecoder (approved):

sumimakito:
@Leopoldthecoder Although we can also show an error message when `isCustomizing` is truthy but the DeckCommandEditor is missing, I think the `console.warn` in the below part would suffice, as users will have nothing to do to recover from this; the developers have to ensure the DeckCommandEditor is provided if the customization is enabled. 

Maybe we can use a KTextArea as a fallback if DeckCommandEditor is missing to ensure the minimum usability if this happens.

PR #3248, opened 2026-05-12
fix(ci): block merge when component tests timeout

2eha0:
## Problem
I recently found a bug that PR can not be blocked when the test is timed out.
- The problem PR: https://github.com/Kong/public-ui-components/pull/3241
- The action logs: https://github.com/Kong/public-ui-components/actions/runs/25655877320/job/75304758976?pr=3241

Root cause:

1. `finish-test-and-publish` had `continue-on-error: true` — GitHub marks the job as succeeded even when it runs `exit 1`, so branch protection rules are never triggered.
2. The failure gate only checked for `'failure'`, but a job that hits `timeout-minutes` gets status `'cancelled'` — so the gate ran `exit 0` and the workflow passed.

## Fix

- Remove `continue-on-error: true` from `finish-test-and-publish`
- Add `|| contains(needs.*.result, 'cancelled')` to the failure condition

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -383,7 +383,6 @@ jobs:
     name: Collect Test Results And Publish
     needs: [build, component-tests]
     runs-on: ubuntu-latest
-    continue-on-error: true
     timeout-minutes: 20
     permissions:
       id-token: write # This is required for requesting the JWT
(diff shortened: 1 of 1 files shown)


TT1228:
Could we have a minimal PR that shows this change blocks cancelled job?

2eha0:
> Could we have a minimal PR that shows this change blocks cancelled job?

https://github.com/Kong/public-ui-components/pull/3249

TT1228 (approved):

merrywhether:
Natural experiment currently in main, suite/check is being cancelled because a test is hanging right now. From another branch open right now (#3854, unrelated to issue):
image

https://github.com/Kong/public-ui-components/actions/runs/35657475867/job/106526004899?pr=3854
image

With same main issue merged into this branch, "Collect Test Results" fails appropriately:
image

merrywhether:
#3855 has merged so this PR will pass CI once its branch is updated (leaving it as-is for now for demo purposes)

.github/workflows/ci.yaml
```diff
@@ -377,7 +377,6 @@ jobs:
     name: Collect Test Results And Publish
     needs: [build, component-tests]
     runs-on: ubuntu-latest
-    continue-on-error: true
```
ValeryG, line 380: not sure if removing this is a right move. without this line we would never send a slack notification on failure ....

ValeryG (changes requested):

PR #3253, opened 2026-05-13
feat(entities-plugins): reset cleared array enums to null

2eha0:
## Summary
- reset optional ArrayField values to null when the last item is removed
- reset optional multiselect EnumField values to null when cleared
- remove the now-obsolete rate-limiting-advanced workaround and add focused field specs

### DIFF
--- a/packages/entities/entities-plugins/src/components/free-form/shared/EnumField.vue
+++ b/packages/entities/entities-plugins/src/components/free-form/shared/EnumField.vue
@@ -83,6 +83,11 @@ const fieldModel = defineModel({
   // If multiple and value is null/undefined, return empty array for v-model
   get: () => isMultiple.value && fieldValue?.value == null ? [] : fieldValue!.value,
   set: (val) => {
+    if (isMultiple.value && Array.isArray(val) && val.length === 0 && !field.schema?.value?.required) {
+      fieldValue!.value = null as any
+      return
+    }
+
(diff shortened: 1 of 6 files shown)


packages/entities/entities-plugins/src/components/free-form/shared/EnumField.vue
```diff
@@ -75,20 +76,27 @@ const {
...
+function normalizeValue(value: EnumValue): EnumValue {
+  if (isMultiple.value && Array.isArray(value) && value.length === 0 && !fieldAttrs.value.required) {
+    return null
+  }
+
+  return value
+}
 
 const fieldModel = defineModel({
   // If multiple and value is null/undefined, return empty array for v-model
```
TT1228, line 93: I'm curious what's the purpose of translating `null` to empty array in editing?

TT1228: The normalization happens when field data changes, imagine in edit some enum field gets `[]` for v-model and remain untouched in edit, the form will pass `[]` in payload. Is this expected?

2eha0: The normalization only affects the model, form data won't be modified util user change its value

TT1228 (approved):
LGTM

PR #3259, opened 2026-05-14
fix(page-layout): shortcuts support [KHCP-20629]

portikM:
# Summary

Addresses:
- https://konghq.atlassian.net/browse/KHCP-20629
- https://konghq.atlassian.net/browse/KHCP-20410

Add `pageShortcutData` prop and integration with host app shortcuts functionality. The integration has two halves:

1. The page declares itself as a shortcut-eligible entity via the `pageShortcutData` prop.
2. The host app provides a `PageShortcutsContext` via Vue `provide()` under the `app:pageShortcutsContext` injection key. This context is reactive and tells `PageLayout` whether the current page is currently favorited, and gives it callbacks to invoke when the user favorites a page or when an entity page is visited.

The favorite star button is only rendered when **both** halves of the contract are satisfied:

- `pageShortcutData` is provided with non-empty `entityType`, `path`, and `label`, **and**
- An `app:pageShortcutsContext` is injected and exposes an `onFavoriteToggle` function.

If either side is missing, the button is hidden entirely (no DOM is rendered for it). This means a page can opt-in unconditionally by declaring `pageShortcutData`, and the host app controls global availability of the feature by deciding whether to `provide()` the context.


packages/core/page-layout/docs/page-layout.md
```diff
@@ -293,6 +309,95 @@ interface PageLayoutTab {
...
+
+### `pageShortcutData` prop
+
+Provide this prop to mark the page as an entity page. The `path` field is the URL the shortcut will navigate to when clicked from the host's shortcut list; `label` is the display text; `entityType` is a string the host uses to group / categorize shortcuts (e.g. `"gateway-service"`, `"route"`). An optional `parentLabel` can disambiguate entities with the same label under different parents.
+
+### `app:pageShortcutsContext` injection
+
+The host application must `provide('app:pageShortcutsContext', ctx)` a reactive object (typically created via `reactive()`). `PageLayout` interacts with it as follows:
+
+- **`isFavorite`** — reactive boolean. When `true`, the star button renders as a filled star (and its aria-label switches to "Remove page from shortcuts"). The host is responsible for keeping this in sync with the user's favorites list whenever the route changes.
```
ValeryG, line 334: I am a little confused here. app:pageShortcutsContext is injected once  (I guess in child-app), and it is to provide methods to add to the list of recents, and toggle favorite.  
But why is `isFavorite` part of this - it looks like it is not a method,  why it how come it's mixed up with methods? where is the data to calculate  isFavorite value?

portikM: `isFavorite` is needed in order to know which state to render the star in. The logic related to managing `isFavorite` state lives in the app-root.

ValeryG (approved):

PR #3274, opened 2026-05-20
feat(entities-plugins): add oidc principals support

TT1228:
# Summary

KM-2599
Disable principal rendering for `OIDC` in KM and add render template for Konnect.

### DIFF
--- a/packages/core/forms/src/components/forms/OIDCForm.vue
+++ b/packages/core/forms/src/components/forms/OIDCForm.vue
@@ -23,6 +23,14 @@
             Parameters for enabling the OpenID Connect plugin. Set these parameters before adding authorization,
             authentication, or other advanced configuration details.
           </div>
+
+          <OIDCPrincipals
+            v-if="isKonnect && hasPrincipalsFields"
+            :form-model="formModel"
+            :form-options="formOptions"
+            :form-schema="formSchema"
+            :on-model-updated="onModelUpdated"
+          />
           <VueFormGenerator
             v-if="displayForm"
             :model="formModel"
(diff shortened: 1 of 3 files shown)


packages/core/forms/src/components/forms/OIDCForm.vue
```diff
@@ -215,6 +240,14 @@ export default {
...
       return (this.formModel.id && this.isEditing) || !this.isEditing
     },
+    hasPrincipalsFields() {
+      return this.formSchema.fields?.some(
+        (field) => PRINCIPALS_FIELD_MODELS.has(field.model),
+      )
+    },
+    isKonnect() {
+      return this.formsConfig?.app === 'konnect'
+    },
```
jw-foss, line 250: Can these be merged into one as `showPrincipalsFields` since `isKonnect` was not used elsewhere

packages/core/forms/src/components/forms/OIDCPrincipals.vue
```diff
@@ -0,0 +1,94 @@
+<template>
+  <div
+    class="oidc-principals-section"
+    data-testid="oidc-principals-section"
+  >
+    <div class="principals-header">
+      Who should manage authentication and credentials?
```
jw-foss, line 7: Should the following lines be translated entries?

L#7
L#10
L#14

jw-foss (approved):

packages/core/forms/src/components/forms/OIDCPrincipals.vue
2eha0: All copies in plugin form are not translated. I think these are fine to hard-code

TT1228: This file will be reimplemented after design finalized

packages/core/forms/src/components/forms/OIDCForm.vue
TT1228: I will update it in next PR

PR #3369, opened 2026-05-28
fix(page-layout): debounce entity page visit callback

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-20675

Changes:
- add `title` slot in PageLayout
- debounce `onEntityPageVisit` call in watcher


ValeryG (approved):
as long as it is not breaking e2e tests against new nav. :)

PR #3374, opened 2026-05-28
fix(analytics-chart): add i18n messages for event-gateway charts

kingyue737:
[KM-2638]

Add missing i18n messages for Event Gateway metrics charts in analytics-chart.

[KM-2638]: https://konghq.atlassian.net/browse/KM-2638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/analytics/analytics-chart/src/locales/en.json
```diff
@@ -202,6 +202,8 @@
     "mcp_response_body_size_sum": "MCP Response Size (sum)",
     "control_plane_count": "Control plane count",
     "service_count": "Gateway service count",
+    "record_bytes:sum": "Bytes",
+    "record_count:sum": "Count",
```
adorack: We usually recommend that there's more context here -- e.g., something like 'Record size (sum)' rather than just 'Bytes'.  The 'bytes' may be misleading in some cases; e.g., here's an example of a similar record where the sum is in KB rather than B:

Image

Do you have guidance that you're following about what to put here?

kingyue737: image

I came up with these two i18n messages based on Anthony's sketch, but I think your point makes sense. I'll sync with Anthony.

adorack (approved):

PR #3401, opened 2026-06-01
feat(analytics-chart): support multiple dimensions in TopN table

filipgutica:
# Summary

https://konghq.atlassian.net/browse/MA-5040
https://konghq.atlassian.net/browse/MA-5089

- Support up to 3 dimensions in TopN table

## 1 Dimension
image

## 2 Dimensions
image

## 3 Dimensions
image

### DIFF
--- a/packages/analytics/analytics-chart/src/components/TopNTable.vue
+++ b/packages/analytics/analytics-chart/src/components/TopNTable.vue
@@ -62,14 +62,20 @@
       class="top-n-table"
       data-testid="top-n-table"
     >
-      <table class="top-n-table-table">
+      <table
+        class="top-n-table-table"
+        :class="{ 'top-n-table-table--multi-dimension': hasMultipleDimensions }"
+      >
         <thead data-testid="top-n-table-header">
           <tr class="top-n-table-header-row">
             <th
               v-for="header in tableHeaders"
               :key="header.key"
               class="top-n-table-header-cell"
-              :class="{ 'top-n-table-header-cell-metric': header.key !== 'name' }"
+              :class="{
+                'top-n-table-header-cell-metric': header.type === 'metric',
+                'top-n-table-header-cell-dimension-compact': isCompactDimensionHeader(header),
+              }"
               data-testid="top-n-table-header-column"
             >
               <span class="table-header-label">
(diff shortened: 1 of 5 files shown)


packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -82,11 +88,14 @@
...
+            :key="row.rowKey"
             class="top-n-table-row"
           >
-            <!-- Name / dimension column -->
-            <td class="top-n-table-cell top-n-table-cell--name">
+            <!-- Primary dimension column -->
+            <td
+              class="top-n-table-cell top-n-table-cell--name"
+              :class="{ 'top-n-table-cell-dimension-compact': hasMultipleDimensions }"
+            >
```
Chalks, line 98: **question**

why is the primary dimension special? Couldn't this just be part of the `v-for` loop for all the other dimensions too?

packages/analytics/analytics-chart/src/components/TopNTable.vue
```diff
@@ -177,12 +209,24 @@ const { formatUnit } = unitFormatter({ i18n })
...
 
-const displayKey = computed((): string => {
+const displayKeys = computed((): string[] => {
   if (!props.data.meta?.display) {
-    return ''
+    return []
   }
 
-  return Object.keys(props.data.meta.display)?.[0] || ''
+  return Object.keys(props.data.meta.display).slice(0, MAX_DIMENSIONS)
```
Chalks, line 222: **question, possibly a blocker**

If there are more than `MAX_DIMENSIONS`, doesn't this present extremely misleading data and lead to duplicate entries? Like if you had four dimensions and three metrics with values `[a, b, c, d, 2, 3, 4], [a, b, c, z, 7, 8, 9]`, you would display two rows with `a b c 2 3 4` and `a b c 7 8 9`?

packages/analytics/analytics-chart/src/components/TopNTable.vue
Chalks: discussed in chat. we're preserving existing behavior.

Chalks (approved):

packages/analytics/analytics-chart/src/components/TopNTable.vue
Chalks: discussed in chat, restricted to 3 because that's what our api supports.

PR #3407, opened 2026-06-03
feat(entities-vaults): support ai gateway vault

Leopoldthecoder:
# Summary

This pull request adds support for Kong AI Gateway vault providers across the vaults entity packages, including configuration, forms, lists, and secret picker integrations. It ensures correct field translation between API Gateway and AI Gateway vault shapes.


kingyue737 (approved):
LGTM

Q: will we support terraform? The resource type may need be updated in the future:

Image

Leopoldthecoder:
> LGTM
> 
> Q: will we support terraform? The resource type may need be updated in the future:
> 
> Image

Good question. It is already on my questions-to-be-checked-with-pm list.

PR #3418, opened 2026-06-05
fix(page-layout): improve favorite button state logic

portikM:
# Summary

Improve favorite button state logic

### DIFF
--- a/packages/core/page-layout/src/components/PageLayout.vue
+++ b/packages/core/page-layout/src/components/PageLayout.vue
@@ -50,6 +50,7 @@
             </span>
             <div
               v-if="showFavoriteButton"
+              :key="favoriteButtonKey"
               class="favorite-button-container"
             >
               <KTooltip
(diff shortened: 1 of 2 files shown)


packages/core/page-layout/src/components/PageLayout.vue
```diff
@@ -143,7 +144,20 @@ const isBackToString = computed((): boolean => typeof backTo === 'string')
...
+const favoriteButtonKey = ref<number>(0)
+/**
+ * pageShortcutsContext has both isFavorite and isPageFavorite properties.
+ * isFavorite is a boolean that indicates if the page is favorited.
+ * isPageFavorite is a function that returns a boolean that indicates if the page is favorited.
+ * We need to check both properties.
+ */
+const isFavorite = computed((): boolean =>
+  !!pageShortcutsContext &&
+  (('isFavorite' in pageShortcutsContext &&
```
adamdehaven, line 156: You're checking a few of these conditions multiple times. Can you refactor so they are only each evaluated once?

portikM: This is the only place in this component where we check for `isFavorite` and  `isPageFavorite`. This computed property is literally the one place that handles all those checks so I don't need to perform them elsewhere.

adamdehaven (approved):

PR #3423, opened 2026-06-08
feat: initial PR to scaffold table-data-grid

filipgutica:
# Summary

Break up: https://github.com/Kong/public-ui-components/pull/3209 into smaller PRs

https://konghq.atlassian.net/browse/MA-5115

Adds the initial `@kong-ui-public/table-data-grid` package scaffold under `packages/core/table-data-grid`.

This PR establishes the public `TableDataGrid` component contract, wires up basic AG Grid rendering, adds initial Kong/AG Grid styling, and includes a minimal sandbox plus a happy-path component test. The component accepts a basic fetcher and header configuration, calls the fetcher for the first page, and renders returned static row data in AG Grid.

## Scope

- Adds the new `packages/core/table-data-grid` package.

- Exports `TableDataGrid` and initial `TableDataGrid*` public types.

- Defines the initial fetcher contract:

  - `TableDataGridFetcherParams`

  - `TableDataGridFetcherResult`

  - `TableDataGridFetcher`

- Renders AG Grid columns from `headers`.

- Fetches and renders row data from the provided `fetcher`.

- Emits `grid:ready` with the AG Grid API.

- Adds initial component styles using AG Grid theme variables and Kong tokens.

- Adds README usage docs.

- Adds minimal sandbox scaffolding.

- Adds a component test covering headers, data rendering, and fetcher params.


packages/core/table-data-grid/src/components/TableDataGrid.vue
```diff
@@ -0,0 +1,132 @@
...
+    minWidth: header.minWidth,
+    valueGetter: params => params.data?.[header.key],
+    width: header.width,
+  }
+
+  return columnDef
+}))
+
+const fetchRows = async () => {
+  const fetchId = latestFetchId.value + 1
```
adorack, line 64: When does the table issue multiple fetches such that you need to squash outdated ones?  Is it possible to trigger this in the current setup?

(It might help to describe the situation in a comment.)

packages/core/table-data-grid/src/components/TableDataGrid.vue
```diff
@@ -0,0 +1,132 @@
...
+  emit('grid:ready', event.api)
+}
+
+watch(
+  [
+    toRef(() => fetcher),
+    toRef(() => pageSize),
+  ],
+  () => {
+    void fetchRows()
```
adorack: `watch` can handle an async effect, so I don't think you need the `void` here.  (I'd never seen that construction before; I think in almost all cases we want to try to avoid discarding the promise.)

packages/core/table-data-grid/README.md
```diff
@@ -0,0 +1,103 @@
+# TableDataGrid
+
+Reusable Vue wrapper around AG Grid for Kong table data grids.
+
+This initial package version supports a basic first-page fetcher and AG Grid rendering. Features such as pagination controls, infinite scroll, row selection, toolbar controls, filtering, search, sorting, column visibility, slottable cells, persistent state, and movable or pinnable columns are intentionally out of scope for the first PR.
```
adorack: You probably don't need to call out the "first PR" in the README.

package.json
```diff
@@ -95,7 +95,8 @@
       "lodash-es@<=4.17.23": ">=4.18.0",
       "lodash@<=4.17.23": ">=4.18.0",
       "axios@<1.15.0": ">=1.16.1",
-      "tmp@<0.2.6": ">=0.2.6"
+      "tmp@<0.2.6": ">=0.2.6",
+      "shell-quote@>=1.1.0 <=1.8.3": ">=1.8.4"
```
adorack, line 99: Huh -- out of curiosity, what's `shell-quote` getting you?

adorack (changes requested):

package.json
filipgutica: This was due to pnpm audit issues

packages/core/table-data-grid/src/components/TableDataGrid.vue
filipgutica: done

packages/core/table-data-grid/README.md
filipgutica: done

package.json
filipgutica: It's in the `fix: pnpm audit` commit

packages/core/table-data-grid/src/components/TableDataGrid.vue
filipgutica: Example race condition:

1. fetch A starts with pageSize = 25
2. pageSize changes, fetch B starts with pageSize = 50
3. fetch B resolves first and sets correct data
4. fetch A resolves later
5. The guard prevents A from replacing B’s data

Let me know if this is being overly defensive.
I can add a unit test in the next PR that adds fetch handling composable/utilities

filipgutica: Or, just more broadly... if request A starts then request B starts, B resolves first, and A resolves later, A is ignored instead of overwriting B’s newer rows. 

In platform analytics it could be... load page initially with a table explore query for control planes... then you quickly switch to routes... if the result from the control planes query resolves after, the routes one, you don't want that one to overwrite it, showing you a list of control planes when you want to be looking at a list of routes

filipgutica: Perhaps this can also be addressed with an abort controller 🤔 

But the host app would handle that in the fetcher implementation? So this is safe to leave as an internal defensive check?

johncowen: Just an opinion, but this feels like a strange design choice if we are hitting race issues and then having to fix them with defensive coding here.

Q: Would it be a better design to pass the data in from The Outside rather than have this race-y fetcher approach? (a bit like what it seems like AGGrid already accepts in `:row-data`)

johncowen: P.S. Sorry, we crossed responses a little here 😆 , comment was in reference to this one https://github.com/Kong/public-ui-components/pull/3423/changes#r3389912357, I've not read the others yet!

packages/core/table-data-grid/src/types/index.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+      mode: 'infinite',
+      pageSize: 25,
+    } satisfies TableDataGridFetcherParams
+
+    expect(params).toEqual({
+      cursor: 'next-cursor',
+      mode: 'infinite',
+      pageSize: 25,
+    })
+  })
```
johncowen: What's the intention here?

packages/core/table-data-grid/src/components/TableDataGrid.vue
```diff
@@ -0,0 +1,132 @@
...
+  --ag-header-column-border: 1px solid var(--kui-color-border, #{$kui-color-border});
+  --ag-header-column-resize-handle-color: transparent;
+  --ag-wrapper-border: none;
+  --ag-wrapper-border-radius: 0;
+  /* stylelint-enable custom-property-pattern */
+
+  flex: 1 1 360px;
+  min-height: 360px;
+  width: 100%;
+}
```
johncowen, line 126: Could you let me know what this CSS is doing (the layout stuff not the coloring stuff).

Also, maybe a nit but I'm interested why:

How come we use the `var(--kui-color-border, $kui-color-border);` form in one place but the `var(--kui-color-background, #{$kui-color-background});` form in another?

lmk!

packages/core/table-data-grid/src/types/index.spec.ts
filipgutica: Sorry, unnecessary test, removed it

packages/core/table-data-grid/src/components/TableDataGrid.vue
adorack: > In platform analytics it could be... load page initially with a table explore query for control planes... then you quickly switch to routes... if the result from the control planes query resolves after, the routes one, you don't want that one to overwrite it, showing you a list of control planes when you want to be looking at a list of routes

I was wondering about that, because right now, this part of the query is handled in the fetcher -- at least from what I saw, the table component doesn't know anything about what's happening in the fetcher.  I think something must end up resetting the table state to handle changes in what the table is trying to render?

However, `pageSize` changing I guess is something that I could picture.  Hmmm.  I'm not sure that an abort controller would solve it unless you add additional requirements to the type of the `fetcher` -- we do this for Analytics components, but I'm not sure if other areas in the app do it.

I guess given that, I'm fine with this approach.

> Q: Would it be a better design to pass the data in from The Outside rather than have this race-y fetcher approach?

I think the requirements of this component are that it handles fetching.

packages/core/table-data-grid/src/components/TableDataGrid.vue
filipgutica: We’re overriding AG Grid theme variables here to align the grid with Kong tokens. 

Because these are custom property declarations, Sass requires interpolation for token fallbacks, so we use #{$...}. Without it, Sass emits the raw $... string instead of the resolved token value.

https://sass-lang.com/documentation/style-rules/declarations/#custom-properties
https://sass-lang.com/documentation/breaking-changes/css-vars/

adorack (approved):

adorack (changes requested):
Actually -- could you please update CODEOWNERS as well?

packages/core/table-data-grid/src/components/TableDataGrid.vue
filipgutica: Yeah the requirements of this component is that it follows a similar pattern to KTableData in terms of accepting a fetcherFn as a prop so it needs some of this handling internally.

packages/core/table-data-grid/src/components/TableDataGrid.vue
johncowen: > Because these are custom property declarations, Sass requires interpolation for token fallbacks, so we use #{$...}. Without it, Sass emits the raw $... string instead of the resolved token value.

Ah of course! Thanks 👍 

> We’re overriding AG Grid theme variables here to align the grid with Kong tokens.

Ah sorry I meant the layout/width/flex-y stuff

packages/core/table-data-grid/src/components/TableDataGrid.vue
johncowen: Oh gotcha, I think I have a link to the doc let me re-read that. Just weird that from what I understand most of the issues with `KTableData` (and previous iterations) have been down to the race-y `fetcher` approach. Weird that we've chosen to continue this pattern here 🤷

adorack (approved):

adamdehaven (approved):
Unblocking based on @adorack's review

PR #3425, opened 2026-06-09
feat(entities-shared): add SentitiveInput

Leopoldthecoder:
# Summary

This pull request introduces a new reusable component, `SensitiveInput`, to the shared entities package. `SensitiveInput` is designed for securely handling sensitive values such as API keys or tokens, providing features like masking, visibility toggling, key generation, and a one-time hint for newly generated keys.

##

#### Create mode
Kapture 2026-06-09 at 16 21 42

##

#### Edit mode
Kapture 2026-06-09 at 16 23 02

##

#### Generate key(optional)
Kapture 2026-06-09 at 16 24 15

##

#### Copy hint(optional)
Kapture 2026-06-09 at 16 25 21


sumimakito (dismissed):
LGTM. Just one thought: Maybe we can make the UI texts customizable via props so that this component can also be applied to scenarios with other types of credentials, like passwords, tokens, etc.

Leopoldthecoder:
> LGTM. Just one thought: Maybe we can make the UI texts customizable via props so that this component can also be applied to scenarios with other types of credentials, like passwords, tokens, etc.

Do you mean the button text, labels, and placeholders? Take tokens for example, do we want to show "Rotate tokens", "Generate tokens", etc.?

packages/entities/entities-shared/src/components/common/SensitiveInput.vue
```diff
@@ -0,0 +1,309 @@
...
+  }
+}
+
+// KInput measures the `after` slot width only once on mount to compute the
+// native input's `padding-right`. Since our `after` slot content changes
+// (Rotate key → Generate key + eye → eye only), that one-time measurement
+// becomes stale. We observe the rendered after-content wrapper and drive the
+// input's `padding-right` ourselves (via the `inputPaddingRight` CSS binding
+// below), reusing KInput's own formula: `calc(space-50 + after-width + space-40)`.
+const rootElement = ref<HTMLElement>()
```
kingyue737: nit: we can use `useTemplateRef`

sumimakito:
> > LGTM. Just one thought: Maybe we can make the UI texts customizable via props so that this component can also be applied to scenarios with other types of credentials, like passwords, tokens, etc.
> 
> Do you mean the button text, labels, and placeholders? Take tokens for example, do we want to show "Rotate tokens", "Generate tokens", etc.?

Yes, I think this can be an improvement, but it should not block this PR from being merged.

packages/entities/entities-shared/src/components/common/SensitiveInput.vue
```diff
@@ -0,0 +1,309 @@
...
+  }
+  syncInputPadding()
+})
+
+onBeforeUnmount(() => {
+  resizeObserver?.disconnect()
+  if (rafId !== undefined) {
+    window.cancelAnimationFrame(rafId)
+  }
+})
```
kingyue737, line 267: Non-blocking suggestion: 
If we introduce `@vueuse/core` to `entites-shared`,  the manual `ResizeObserver` + `requestAnimationFrame` + cleanup logic could be replaced with something like:
```ts
 const afterContentRef = ref<HTMLElement>()
 
 onMounted(() => {
   afterContentRef.value = rootElement.value?.querySelector<HTMLElement>('.after-content-wrapper')
 })
 
 const { width: afterWidth } = useElementSize(afterContentRef)
 
 const inputPaddingRight = computed(() =>
   `calc(${KUI_SPACE_50} + ${afterWidth.value}px + ${KUI_SPACE_40})`
 )
```

Ref: https://vueuse.org/core/useElementSize/

kingyue737 (dismissed):
LGTM

Leopoldthecoder:
> > > LGTM. Just one thought: Maybe we can make the UI texts customizable via props so that this component can also be applied to scenarios with other types of credentials, like passwords, tokens, etc.
> > 
> > 
> > Do you mean the button text, labels, and placeholders? Take tokens for example, do we want to show "Rotate tokens", "Generate tokens", etc.?
> 
> Yes, I think this can be an improvement, but it should not block this PR from being merged.

Yes, I had actually been thinking about this improvement as well, but it slipped my mind. I've just added a new labels prop to support it.

packages/entities/entities-shared/src/components/common/SensitiveInput.vue
Leopoldthecoder: Updated.

packages/entities/entities-shared/src/components/common/SensitiveInput.vue
Leopoldthecoder: Good idea. But I'm not sure if this justifies adding `@vueuse/core` as a dependency. Maybe we should wait until we have a stronger use case?

kingyue737 (approved):

sumimakito (approved):
LGTM. Thanks!

PR #3426, opened 2026-06-09
fix(entities-plugins): show inline realm error

TT1228:
# Summary

**Changes**

- Introduced a `BEFORE_SAVE_KEY` injection key so child components can register before-save guards. `PluginForm.vue` provides it and runs all guards before submitting; submission is silently blocked if any guard returns false.
- `ConfigFormContent.vue` registers a guard that blocks submission if no realm is selected, and only reveals the inline error after the user has touched the field (click or value change) or attempted to submit. The error auto-clears once a valid realm is selected.
- Moved the `identity_realms` field out of the advanced section and inline below the mode selector when in `centrally-managed` mode.
- Fixed `detectInitialMode` for Konnect: only treats existing data as `centrally-managed` when there is at least one non-placeholder realm (scope !== `'cp'` with a non-null id).
- Added `isLoadingRealms` state passed to `KongIdentityField` to show a loading skeleton while realms are being fetched.
- Fixed an edit-mode bug where removing all realms wouldn't trigger validation, because `realmTouchReady` was never initialized when the component mounted already in `centrally-managed` mode.
- Updated the `identity_realms_required` i18n string to clarify the requirement.
- Added Cypress tests for `error_on_miss`, `Required principals behavior`, and the new realm validation behavior.

https://github.com/user-attachments/assets/bcfe6612-053c-41e9-a80f-f7fdc4ae4c5d

### DIFF
--- a/packages/entities/entities-plugins/src/components/PluginForm.vue
+++ b/packages/entities/entities-plugins/src/components/PluginForm.vue
@@ -217,7 +217,7 @@ import {
 import PluginEntityForm from './PluginEntityForm.vue'
 import PluginFormActionsWrapper from './PluginFormActionsWrapper.vue'
 import unset from 'lodash-es/unset'
-import { REDIS_PARTIAL_INFO } from '../components/free-form/shared/const'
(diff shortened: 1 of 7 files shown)


2eha0 (approved):
LGTM

PR #3443, opened 2026-06-13
Feat: Add table-data-grid infinite scroll support

filipgutica:
## Summary

This branch adds cursor-first infinite scrolling support to `@kong-ui-public/table-data-grid` and updates the surrounding component behavior, docs, and sandbox examples.

- define the public cursor-first infinite fetcher contract
- add `useFetchInfinite` for AG Grid infinite datasource wiring, cursor chaining, refresh invalidation, and stale request guards
- add fetch lifecycle state derivation and public `state` emits, including later-block loading and success transitions
- wire `TableDataGrid` to AG Grid infinite row model with `pageSize`, `refreshKey`, host error handling, empty state handling, and `grid:ready`
- add default and slotted empty/error presentation states backed by Kongponents
- document usage, fetcher behavior, refresh invalidation, internal fetch state, state emits, slots, column sizing, and host setup expectations
- rebuild the sandbox into a reference-style infinite-fetch playground with dataset controls, fetch delay, host error, refresh, state logs, and fetch history
- make unconstrained columns fill available table width by default while preserving explicit width and max-width constraints
- declare Kongponents as a package peer/dev dependency

## Cursor-based infinite fetching state model

AG Grid’s infinite row model is range-based: it calls the datasource with row windows like `0-25`, `25-50`, and `50-75`. TableDataGrid translates those ranges into cursor-backed blocks while keeping `startRow` / `endRow` internal.

State flow:

1. AG Grid requests a range through `getRows({ startRow, endRow })`.
2. TableDataGrid derives:
   - `pageSize = endRow - startRow`
   - `blockIndex = startRow / pageSize`
3. Block `0` calls the public fetcher with `cursor: undefined`.
4. The fetcher response returns `{ data, cursor, total?, hasMore? }`.
5. TableDataGrid passes `data` back to AG Grid with `successCallback`.
6. If the response includes a cursor, TableDataGrid stores it in an internal `cursorMap` keyed by the block that produced it.
7. When AG Grid later requests the next range, TableDataGrid derives the next `blockIndex`.
8. Before fetching that next block, TableDataGrid reads the cursor stored for `blockIndex - 1`.
9. The next fetcher call receives that cursor:
   - block `0` -> `cursor: undefined`
   - block `1` -> cursor returned by block `0`
   - block `2` -> cursor returned by block `1`
10. Response metadata is translated back into AG Grid’s `lastRow` signal so AG Grid knows whether to keep requesting more blocks.

## Notes

This slice stays scoped to the current infinite-fetch surface. Pagination, filters, selection, configurable/resizable columns, toolbar behavior, and table-config persistence are left for later work.

### DIFF
--- a/packages/core/table-data-grid/src/components/TableDataGrid.vue
+++ b/packages/core/table-data-grid/src/components/TableDataGrid.vue
@@ -62,7 +62,8 @@ import {
   ModuleRegistry,
   themeQuartz,
 } from 'ag-grid-community'
-import { computed, watch } from 'vue'
+import { computed } from 'vue'
+import { useEmitState } from '../composables/useEmitState'
 import { useFetchInfinite } from '../composables/useFetchInfinite'
 import useFetchState from '../composables/useFetchState'
 
@@ -130,51 +131,11 @@ const shouldShowEmptyState = computed<boolean>(() => (
   && !hasData.value
 ))
 
-watch(
-  () => ({
-    hasData: hasData.value,
-    hostError,
-    state: fetchLifecycleState.value,
-  }),
-  ({ hasData, hostError, state }) => {
-    if (hostError) {
-      emit('state', {
-        hasData,
-        state: 'error',
-      })
-
-      return
-    }
-
-    if (state === fetchState.PENDING) {
-      return
-    }
-
-    if (state === fetchState.LOADING) {
-      emit('state', {
-        hasData,
-        state: 'loading',
-      })
-
-      return
-    }
-
-    if (state === fetchState.ERROR) {
-      emit('state', {
-        hasData,
-        state: 'error',
-      })
-
-      return
-    }
-
-    emit('state', {
-      hasData,
-      state: 'success',
-    })
-  },
-  { immediate: true },
-)
+useEmitState({
+  emitState: payload => emit('state', payload),
+  fetchLifecycleState,
+  hasData,
+})
 
 const onGridReady = (event: GridReadyEvent<Row>) => {
   emit('grid:ready', event.api)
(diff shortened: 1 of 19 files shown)


packages/core/table-data-grid/src/components/TableDataGrid.vue
```diff
@@ -3,11 +3,43 @@
...
+    </div>
+
     <AgGridVue
+      v-else
+      :cache-block-size="pageSize"
       class="table-data-grid-grid"
       :column-defs="columnDefs"
-      :grid-options="agGridOptions"
-      :row-data="rowData"
+      :datasource="datasource"
```
filipgutica, line 38: datasource is a ag grid thing, don't confuse with our concept of query datasource.

packages/core/table-data-grid/src/composables/useFetchInfinite.ts
```diff
@@ -0,0 +1,367 @@
...
+      rejectBlockCompletion(blockIndex, currentBlockCompletion)
+      return 'stale'
+    }
+
+    if (!previousBlockCompleted) {
+      rejectBlockCompletion(blockIndex, currentBlockCompletion)
+      return 'failed'
+    }
+
+    return 'ready'
```
Chalks, line 158: **logic question**

Can you get here? Should we fail closed instead of open? It seems like if we didn't hit the 'ready' case on line 135, we should be returning `'failed'` here.

Chalks: discussed in chat. all failure cases are handled, the initial case on line 135 is for just the first entry, not the common/every case.

Chalks (dismissed):

Chalks (approved):

PR #3447, opened 2026-06-16
feat(debugger): Initial changes to show datakit in debugger, with mode=inspect

yugs16:
# Summary

As we want to show Datakit flow inside debugger, we need to Export Datakit FlowCanvas from here, and make it read-only canvas. With some functionalities similar to mode=edit.

More details - [here](https://docs.google.com/document/d/189uEOtyzesldORzbaoRjg4Uk_Ds0Zs2p5Kg-g8OcImU/)
Jira - [OLLY-1022](https://konghq.atlassian.net/browse/OLLY-1022) , [olly-970](https://konghq.atlassian.net/browse/OLLY-970)

## Approach in this PR
Export `FlowCanvas` and `provideEditorStore` directly to consumers.

We added a `mode='inspect'`, for a read-only canvas with click, zoom-in/out, selected-node css, etc. Based on this mode, we keep functionalities same as `mode=edit,` with drag disabled, allowing zoom-in/out, zoom-pinch, css class override and  custom node-click emit. 

### `FlowCanvas` 

- Added `inspect` as a valid `mode` value alongside `edit | view | preview`
- Inspect mode enables trackpad pan/zoom (like edit), keeps nodes non-draggable and
  non-connectable (like view), and emits `node-click` with the `NodeInstance` payload
- Active node border now holds after click in inspect mode — fixed by including
  `mode === 'inspect'` in `elements-selectable`
- Nodes show `cursor: pointer` in debugger mode; other readonly modes keep
  `cursor: default`
- Auto-layout + fit-view runs automatically after nodes initialize in inspect mode
- Added `node-before-handlers`, `node-after-handlers` and `node-actions` named slots, forwarded to `FlowNode`

### `FlowNode`

- Added `#actions` slot for latency placement
- Added  `#node-before-handlers` slot in the node body (for per-node status indicators, e.g. ✓ ok)
- Added  `#node-after-handlers` for future design support.

### Public exports 

- Exports in (`src/index.ts`)
- Export `provideEditorStore`, `useEditorStore` and `FlowCanvas` from here
- NodeBadge`, renders a node type badge (icon + label) based on `node.type`, this we will need to show icon in Input/Output detail view. 

### Mocks

**Current-** 
image

**Target in consumer app-**
image

[OLLY-1022]: https://konghq.atlassian.net/browse/OLLY-1022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/entities/entities-plugins/src/components/free-form/plugins/datakit/DatakitReadOnlyCanvas.vue
```diff
@@ -0,0 +1,53 @@
+<template>
```
yugs16: This component is Optional, and we might not go with it. We can either expose this wrapper on `FlowCanvas`, and use `provideStore` here, or we can directly export `FlowCanvas` and `provideStore`, and let consumer app decide how to use it.

packages/entities/entities-plugins/src/index.ts
```diff
@@ -23,41 +24,20 @@ export {
...
-
-export type {
-  ResolvedPluginFormConfig,
-} from './components/free-form/shared/plugin-registry'
-
 export * from './types'
 
 export { pluginEndpoints }
 
+export { provideEditorStore, useEditorStore } from './components/free-form/plugins/datakit/composables'
```
yugs16: exporting `useEditorStore`, in case we want to update pluginDatat on the same page. Useful for sandbox purpose in consumer apps too.

packages/entities/entities-plugins/src/components/free-form/plugins/datakit/flow-editor/node/FlowNode.vue
```diff
@@ -542,6 +556,12 @@ $one-over-sqrt2-px: math.pow(2, -0.5) * 1px;
     }
   }
 
+  .footer {
+    border-top: 1px solid var(--kui-color-border-neutral-weak, $kui-color-border-neutral-weak);
+    margin-top: var(--kui-space-40, $kui-space-40);
+    padding: var(--kui-space-30, $kui-space-30) var(--kui-space-40, $kui-space-40) 0;
+  }
```
yugs16: This will be removed.

packages/entities/entities-plugins/src/components/free-form/plugins/datakit/flow-editor/node/FlowNode.vue
```diff
@@ -329,6 +344,8 @@ const {
   branchGroups,
 } = useEditorStore()
 
+const slots = useSlots()
```
Justineo: Let's use `defineSlots` for better typing.

packages/entities/entities-plugins/src/components/free-form/plugins/datakit/flow-editor/node/FlowNode.vue
```diff
@@ -274,13 +279,23 @@
...
+
+    <div
+      v-if="slots.latency"
+      class="footer"
+    >
+      <slot
+        :data="data"
+        name="latency"
+      />
+    </div>
```
Justineo: I think we should expose a more generic slot name like `footer` to decouple `FlowNode` itself from `debugger`.

Justineo (commented):
I think we should always auto-layout the canvas, since node positions are saved relative to node sizes, which can vary.

I’m fine with adding a `debugger` mode to `FlowCanvas`, but I don’t think `FlowNode` should be coupled to debugger-specific behavior.

packages/entities/entities-plugins/src/components/free-form/plugins/datakit/flow-editor/node/FlowNode.vue
yugs16: I have used `before-handles` and `after-handles` for naming now, as per our design, we will be using only `before-handles ` and `actions` slots for now, but keeping`after-handles` also for future ref here.

Let me know if hese names are not good enough.

packages/entities/entities-plugins/src/components/free-form/plugins/datakit/flow-editor/node/FlowNode.vue
yugs16: Done

yugs16:
>I think we should always auto-layout the canvas, since node positions are saved relative to node sizes, which can vary.

yes from consumer app, for mode='inspect', we will always do auto-layout. For other modes it will work it used to work right now.

> I’m fine with adding a debugger mode to FlowCanvas, but I don’t think FlowNode should be coupled to debugger-specific behavior.

We need some slot support in debugger, I have added 3 generic slots, let me know if that is also a concern.

packages/entities/entities-plugins/src/components/free-form/plugins/datakit/flow-editor/FlowCanvas.vue
```diff
@@ -6,24 +6,20 @@
     <VueFlow
       :id="flowId"
       class="flow"
-      :class="{ readonly: mode !== 'edit' }"
+      :class="{ readonly: mode !== 'edit', inspect: mode === 'inspect' }"
+      v-bind="interactiveViewportProps"
       :connect-on-click="false"
-      :elements-selectable="mode === 'edit'"
+      :elements-selectable="mode === 'edit' || mode === 'inspect'"
```
Justineo, line 12: [Codex] Inspect mode is meant to be read-only, but enabling selection here makes VueFlow's existing Delete/Backspace handler able to remove selected nodes or edges because `deleteKeyCode` remains active and nodes/edges are not marked `deletable: false` in readonly mode. Please disable deletion outside edit mode or make readonly elements non-deletable.

packages/entities/entities-plugins/src/components/free-form/plugins/datakit/flow-editor/FlowCanvas.vue
```diff
@@ -154,6 +198,12 @@ function getDraggedNodeType(event: DragEvent): string | undefined {
 }
 
 function onNodeClick(event: NodeMouseEvent) {
+  if (mode === 'inspect') {
+    // Emit the full node-data for inspect mode, used in consumer apps
+    emit('node-click', event.node.data as NodeInstance)
```
Justineo, line 205: [Codex] This emits before filtering out group nodes, so clicking a branch group in inspect mode can send `FlowGroupNodeData` while the event is typed as `NodeInstance`. Consumers expecting node fields like `name`, `type`, or `config` can break; please move the group guard before this emit or narrow the emitted payload to real flow nodes.

Justineo (commented):
[Codex] Inline review comments.

packages/entities/entities-plugins/src/components/free-form/plugins/datakit/flow-editor/FlowCanvas.vue
yugs16: Fixed in flow.ts. Disabling delete for readonly nodes.

packages/entities/entities-plugins/src/components/free-form/plugins/datakit/flow-editor/FlowCanvas.vue
yugs16: Done, added return guard for group-node.

Justineo (approved):
LGTM

PR #3464, opened 2026-06-18
feat(analytics): support single result in trend queries [MA-5139]

Chalks:
# Summary

We want single value chart to display trends. However, its current implementation was unique from how analytics-metric-provider did it. So, we're updating single value chart to implement trends the same way. To do this, we had to:

1. refactor analytics metric provider so that `buildDeltaMapping` was exported. This is the function that does all the calculation given a query result for determining if a value is the "previous" or "current" value.
2. update single value chart to use `buildDeltaMapping`
3. update the tests to reflect the new (old) logic

I also ended up needing to fix a minor build issue due to a missing dependency in the analytics chart package.json.

### Screenshots

**show trend, but no previous value exists**
Screenshot 2026-06-22 at 5 56 46 PM

**show trend, but no current value exists**
Screenshot 2026-06-22 at 5 56 32 PM

### DIFF
--- a/packages/analytics/analytics-chart/package.json
+++ b/packages/analytics/analytics-chart/package.json
@@ -76,6 +76,7 @@
     "@types/lodash.pick": "^4.4.9",
     "@types/lodash.pickby": "^4.6.9",
     "@types/papaparse": "^5.5.2",
+    "@vueuse/core": "^14.3.0",
     "approximate-number": "^3.0.1",
     "chart.js": "^4.5.1",
     "chartjs-adapter-date-fns": "^3.0.0",
--- a/packages/analytics/analytics-chart/src/components/chart-types/SingleValue.cy.ts
+++ b/packages/analytics/analytics-chart/src/components/chart-types/SingleValue.cy.ts
@@ -131,9 +131,9 @@ describe('<SingleValue />', () => {
       .should('be.visible')
       .contains('-')
 
-    // cy.getTestId('single-value-trend').should('be.visible')
-    // cy.get('.trend-change').should('have.class', 'neutral').and('have.text', 'Not available')
-    // cy.get('.trend-change .trend-up-icon').should('exist')
+    cy.getTestId('single-value-trend').should('be.visible')
+    cy.get('.trend-change').should('have.class', 'negative').and('have.text', '100.00%')
+    cy.get('.trend-change .trend-down-icon').should('exist')
   })
 
   it('treats a non-numeric value as empty and shows the error state', () => {
--- a/packages/analytics/analytics-chart/src/components/chart-types/SingleValue.cy.ts
+++ b/packages/analytics/analytics-chart/src/components/chart-types/SingleValue.cy.ts
@@ -82,6 +82,60 @@ describe('<SingleValue />', () => {
     cy.getTestId('single-value-trend').should('be.visible')
   })
 
+  it('when showing trend and only have a current result, displays that value and show indeterminate change', () => {
+    const exploreResult = buildExploreResult({ previous: 100, current: 250 })
+    const resultOneValue = {
+      ...exploreResult,
+      data: [
+        exploreResult.data[1], // the current data
+      ],
+    }
+
+    cy.mount(SingleValue, {
+      props: {
+        data: resultOneValue,
+        showTrend: true,
+      },
+    })
+
+    const expected = parseFloat(
+      exploreResult.data[1].event.request_per_minute.toFixed(2),
+    )
+
+    cy.getTestId('single-value-chart')
+      .should('be.visible')
+      .contains(expected)
+
+    cy.getTestId('single-value-trend').should('be.visible')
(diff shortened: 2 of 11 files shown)


adorack (changes requested):
We should talk about this further.

packages/analytics/analytics-chart/package.json
```diff
@@ -76,6 +76,7 @@
     "@types/lodash.pick": "^4.4.9",
     "@types/lodash.pickby": "^4.6.9",
     "@types/papaparse": "^5.5.2",
+    "@vueuse/core": "^14.3.0",
```
Chalks, line 79: this was just... missing? It complains when you `pnpm build` but I guess you can get away with due to workspace dependencies.

packages/analytics/analytics-metric-provider/src/utilities/trend-display.ts
```diff
@@ -71,3 +76,68 @@ export const defineIcon = (polarity: number, thisIsBad: boolean = false): typeof
       ? TrendDownIcon
       : IndeterminateSmallIcon
 }
+
+const setMetric = (result: ChronologicalMappedMetrics, time: 'previous' | 'current', topLevelKey: string | typeof DEFAULT_KEY, secondLevelKey: string | typeof DEFAULT_KEY, metricValue: number) => {
```
Chalks, line 80: all of the additions/changes in the `analytics-metric-provider` are just moving `setMetric` and `buildDeltaMapping` out of the `useMetricFetcher` (and test) into a utility so that it's exported.

packages/analytics/analytics-chart/src/components/chart-types/SingleValue.cy.ts
```diff
@@ -82,6 +82,60 @@ describe('<SingleValue />', () => {
...
+      },
+    })
+
+    cy.getTestId('single-value-chart')
+      .should('be.visible')
+      .contains('-')
+
+    // cy.getTestId('single-value-trend').should('be.visible')
+    // cy.get('.trend-change').should('have.class', 'neutral').and('have.text', 'Not available')
+    // cy.get('.trend-change .trend-up-icon').should('exist')
```
adorack: I think this is intended to check the trend display?

adorack (changes requested):

packages/analytics/analytics-chart/src/components/chart-types/SingleValue.vue
```diff
@@ -137,13 +151,18 @@ const metricUnit = computed((): string | undefined => {
...
   || !!metricName.value?.includes('_latency_')
   || metricName.value === 'error_rate')
 
 const previousValue = computed<number | null>(() => {
-  if (!props.showTrend || records.value.length < 2 || !metricName.value) {
+  if (props.showTrend) {
+    return trendPrevious.value ?? null
+  }
+
+  if (records.value.length < 2 || !metricName.value) {
```
adorack: If you're not showing trend (`props.showTrend` is `false`), shouldn't it not be possible to have a previous value?  Why do you check for 2 values?

(I might not be following the flow correctly, but it seems like if you're not expecting a previous value, because `showTrend == false`, shouldn't this always just return `null`?)

packages/analytics/analytics-chart/src/components/chart-types/SingleValue.vue
```diff
@@ -246,8 +295,9 @@ const colorAttribute = (polarityNum: number) => {
 onMounted(() => {
   if (!props.showTrend && props.data?.data?.length > 1) {
     console.warn('SingleValue chart should only be used with a single data point. Data length:', props.data.data.length)
-  } else if (props.showTrend && props.data?.data?.length !== 2) {
-    console.warn('SingleValue with trend expects exactly 2 data points. Data length:', props.data.data.length)
+  } else if (props.showTrend && props.data?.data?.length > 2) {
+    // because 1 data point is when we don't have any previous data to compare against
+    console.warn('SingleValue with trend expects at least 1 data point, usually 2. Data length:', props.data.data.length)
```
adorack: This warning should probably focus on "got too many data points" rather than "expects at least 1" -- it's fine if it gets 0, but it's not fine if it gets 3.

packages/analytics/analytics-chart/src/components/chart-types/SingleValue.cy.ts
Chalks: whoops, debugging cruft. will fix.

packages/analytics/analytics-chart/src/components/chart-types/SingleValue.vue
Chalks: ah yeah, you're right. I had preserved the previous logic without even thinking about when it's used. I'll clean it up.

adorack (dismissed):

adorack (approved):

PR #3470, opened 2026-06-19
feat(page-layout): add tab slot and tweak styling [KHCP-20703]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-20703
Figma: https://www.figma.com/design/Yze0SWXl5nKjR0rFdilljK/Components?m=auto&node-id=11765-17817&t=hI0vI3LBc5QmMxj4-1

Add tab slot in PageLayout and tweak styling of the overflow menu trigger according to Figma

Screenshot 2026-06-19 at 12 57 25 PM
Screenshot 2026-06-19 at 12 58 34 PM


packages/core/page-layout/src/components/PageLayout.vue
```diff
@@ -92,7 +92,17 @@
...
+        <template
+          v-for="tab in tabs"
+          #[`tab-${tab.key}`]="slotProps"
+        >
+          <slot
+            :name="`tab-${tab.key}`"
+            v-bind="slotProps"
+          />
+        </template>
+      </PageLayoutTabs>
```
johncowen, line 104: q: seeing as we are letting people slot things, what do you think of requiring folks to use a `RouterLink` in the slot to wrap their `label`?

Then we'd get all the benefits of native links (including native accessibility by default e.g. command-click to open in a new Browser Tab) without needing any javascript (which means less stuff to maintain/fix)

The following is a bit pseudo-code-y, but I guess something like this:

```vue
<PageLayoutTabs>
  <template>
      v-for="{ label, ...tab } in tabs"
      :key="label"
      #[`${label}-tab`]
    >
      <RouterLink
        :aria-current="tab.name === currentRoute.name"
        :to="tab"
      >
        {{ label }}
    </RouterLink>
  </template>
</PageLayoutTabs>
```

wdyt?

---

FYI: Here's our KTabs replacement we did for kuma recently which works the same way, is very accessible and has almost no javascript to maintain:

https://github.com/kumahq/kuma-gui/blob/667c78d7fee567f4ff49aba905b8dc6287fb6197/packages/x/src/components/x-tabs/XTabs.vue

Usage (XAction is our version of RouterLink):

https://github.com/kumahq/kuma-gui/blob/667c78d7fee567f4ff49aba905b8dc6287fb6197/packages/kuma-gui/src/app/meshes/views/MeshDetailTabsView.vue#L37-L52

portikM: > Then we'd get all the benefits of native links (including native accessibility by default) without needing any javascript (which means less stuff to maintain/fix)

That's opposite of what we want in certain scenarios:
- when route object is passed through the `to` parameter - the `router-link` component is rendered - native and optimal navigation handling
- when a string is passed - anchor element is used but instead of native navigation we're using injected `app:navigateTo` method which allows navigating to other MFEs without page reload - which is the desired behaviour

As for accessibility, the elements rendered by PageLayout tabs should be accessible - can you specify any issues if you know something I don't?

johncowen: > when route object is passed through the to parameter - the router-link component is rendered - native and optimal navigation handling

Is this the opposite scenario of my example? It sounds the same to me.

The route object is passed, the router-link component is rendered - native and optimal navigation handling. Its also far clearer to the engineer how things work and no extra code is required.

Functionally, it renders the same thing but with less custom Javascript that will need fixing occasionally and generally maintaining. I would argue using a composition pattern to pass the route object straight to a RouterLink is a better approach.

> when a string is passed - anchor element is used but instead of native navigation we're using injected app:navigateTo method which allows navigating to other MFEs without page reload - which is the desired behaviour

Curious to understand the usecases/instances where clicking a tab navigates you to a different MFE? Can you only ever link to another MFE from a PageLayoutTab? Thats sounds strange to me. I'm curious to understand this more. Let me know, I'd be interested to hear!

> As for accessibility, the elements rendered by PageLayout tabs should be accessible - can you specify any issues if you know something I don't?

One example is: "e.g. command-click to open in a new Browser Tab)" - if you hold down the command key and click the tab should open the link in a new browser tab. I personally use this quite a lot. Unless I'm mistaken, this functionality is missing from PageLayoutTabs.

There are other instances where not letting people add links via slots is problematic. i.e. people adding Tooltips (interactive content) into a slot which (with your current approach) ends up _inside_ the link (interactive content inside a link is explicitly forbidden https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#technical_summary). The approach makes it easy to do the wrong thing and hard to do the right thing. I've seen quite a few instances of this happening with the older KTabs in our apps.

There are probably other things, but I think the biggest point is, if you use the browsers native functionality where possible then things will Just Work, for present and future functionality.

If we were to require people to compose the RouterLink into the slot itself, then they could add the Tooltip to the anchor itself. Its very clear how things work (just use a link) and all native browser functionality/accessibility works by default, there is nothing to fix and maintain, and we make it easy to do the right thing and hard to do the wrong thing.

Have a think and let me know, if you have any other opposing arguments I'd also be curious to understand them 👍

johncowen: P.S. I also just noticed that PageLayoutTabs uses a `space` to "activate" a link/tab.

https://github.com/Kong/public-ui-components/blob/e7b1fcb4cb66017a05b98d50f9533a8761b4f6a9/packages/core/page-layout/src/components/PageLayoutTabs.vue#L25

The correct behaviour for tabbing to a link and hitting `space` on a link is to scroll the page, not activate the link.

portikM: > Curious to understand the usecases/instances where clicking a tab navigates you to a different MFE?

Organization area - _Labs_ and _Plan and usage_ L2s are standalone MFEs so basically PageLayoutTabs are linking to a different MFE within organization MFE.

> One example is: "e.g. command-click to open in a new Browser Tab)" - if you hold down the command key and click the tab should open the link in a new browser tab. I personally use this quite a lot. Unless I'm mistaken, this functionality is missing from PageLayoutTabs.

Good call, I'll make a ticket to look into fixing that.

> people adding Tooltips (interactive content) into a slot

AFAIK there currently is no use-cases for tooltips in PageLayout tabs. The only reason why we need the slot today is for passing the _New_ or _Beta_ badges in some rare scenarios.

The biggest opposing argument I have is I see no value in exposing the slot for folks to provide their own link elements that would justify the effort. We'd need to rewrite the styles to make sure the links are styled correctly when a custom link is slotted (which maybe isn't such a big deal) but ultimately we'd be exposing ourselves to possibility of some tabs in some area to look / work inconsistently from others. One of the main goals PageLayout and PageLayoutTabs components were created to tackle is ensuring consistency in common elements (page title, breadcrumbs, second level tabs navigation) across all pages and areas in Konnect. Until there is no clear use case that would demand allowing slotting custom link elements, I am uncomfortable with giving up control over links looks and behaviour we have with current setup.

johncowen: > Organization area - Labs and Plan and usage L2s are standalone MFEs so basically PageLayoutTabs are linking to a different MFE within organization MFE.

Ok let me take a look at that. I think I found where it is. Is this the only place where we link from one MFE to another? Is there anywhere else in the entirety of Konnect where we link to another MFE but _not_ from a tab?

packages/core/page-layout/src/components/PageLayout.cy.ts
```diff
@@ -106,6 +106,31 @@ describe('<PageLayout />', () => {
...
+    })
+
+    cy.getTestId('page-layout-tab-overview')
+      .findTestId('custom-overview-tab')
+      .should('be.visible')
+      .and('contain.text', 'Custom Overview')
+    cy.getTestId('page-layout-tab-settings')
+      .should('be.visible')
+      .and('contain.text', 'Settings')
+  })
```
kaiarrowood, line 134: nits: 
- preferred pattern is store strings that you assert on (`Custom Overview`, `Settings`) in variables. Variables are used in both prop/slot and assertion
- don't chain assertions, avoid loss of the target elem between assertions (can't remember why this is a problem, just that Cypress flakiness increases when you use `.and`)

packages/core/page-layout/src/components/PageLayoutTabs.cy.ts
```diff
@@ -179,6 +179,46 @@ describe('<PageLayoutTabs />', () => {
     cy.getTestId('tabs-overflow-dropdown-button').should('not.exist')
   })
 
+  it('renders custom content via the dynamic `tab-${key}` slot', () => {
```
kaiarrowood, line 183: Same nits as above

packages/core/page-layout/src/components/PageLayout.vue
johncowen: To come back to you on your other comments:

> Good call, I'll make a ticket to look into fixing that.

I think this is my point. If people could use `RouterLink` within the slot then we wouldn't need another ticket and it wouldn't need fixing, because it just the uses the native link's default behaviour. The way its currently done is creating work for us from now into the future. Do you see my point?

---

What are your thoughts on the "space no longer activating a scroll when a tab is focussed"? 

---

> The biggest opposing argument I have is I see no value in exposing the slot for folks to provide their own link elements that would justify the effort.

I'm not sure I understand this, it feels like it has taken more effort (and is continuing to require more effort, see above issues for examples) by trying to reinvent accessible HTML links ourselves with javascript?

> We'd need to rewrite the styles to make sure the links are styled correctly when a custom link is slotted (which maybe isn't such a big deal)

Agree its not a big deal (and I don't think its a huge change either). But I think the largest benefit is not needing to "fix" things that don't work like native links.

> but ultimately we'd be exposing ourselves to possibility of some tabs in some area to look / work inconsistently from others.

Curious how this would happen, could you explain further?

I would imagine the styling would stay inside `PageLayoutTabs` and the HTML structure would be exactly the same, the only difference would be that the link would be slotted which I don't think is a major problem? Maybe I'm missing something?

portikM: > I think this is my point. If people could use RouterLink within the slot then we wouldn't need another ticket and it wouldn't need fixing, because it just the uses the native link's default behaviour. The way its currently done is creating work for us from now into the future. Do you see my point?

I see your point but I still think we can overcome challenges like this without exposing slots we don't want to expose in order to maintain control over those elements.

> What are your thoughts on the "space no longer activating a scroll when a tab is focussed"?

I don't even remember where the requirement to navigate on space came from but I guess we can discard this in order to match default element behaviour (already made the change).

Regarding the slotted links conversation, how about we move it to Slack just not to further stretch this PR discussion? Ultimately even if we decide to introduce those slots I won't be tackling that in this PR as I want to keep this PR scoped to the ticket. I think it turns into a bigger discussion and I don't want it to be a blocker for this PR.

packages/core/page-layout/src/components/PageLayout.cy.ts
portikM: Updated

packages/core/page-layout/src/components/PageLayoutTabs.cy.ts
portikM: Updated

packages/core/page-layout/src/components/PageLayout.vue
johncowen: > I see your point but I still think we can overcome challenges like this without exposing slots we don't want to expose in order to maintain control over those elements.

Awesome, thanks!

> how about we move it to Slack just not to further stretch this PR discussion? Ultimately even if we decide to introduce those slots I won't be tackling that in this PR as I want to keep this PR scoped to the ticket. I think it turns into a bigger discussion and I don't want it to be a blocker for this PR.

Yes totally agree! Again thanks for listening 👍 ❤️ 

All in all I see lots of problems that we have to work around (or just accept the issues as something we can't solve) because of a general avoidance of slot usage, which I don't understand. I hear your (or someone else's?) reasoning, but I don't think it holds.

Slots are a native standard feature of VueJS and are essential for building webapps. This is why Vue supports them. I don't understand why we aren't using them, and in turn creating ourselves lots of extra work all for the reasoning of needing to maintain control over the styling of elements (an argument which doesn't hold IMO)

But yes, lets move this to a Slack conversation I think the my argument is more than this PR. I'll make a thread in our channel and re-ask my questions from above and we can keep chatting there.

Just to be clear: I personally don't have approval/change-request power here anyway, but don't let this conversation be a blocker for merging this.

portikM: Perfect, thank you!

kaiarrowood (approved):
LGTM! 🚀

PR #3472, opened 2026-06-22
feat(entities-plugins): add Kong Identity principals creation guide

TT1228:
# Summary

Adds a **Principals Creation Guide** to the Kong Identity section of the plugin config form (Konnect only). When a plugin is set to use Kong Identity (principal-based auth) but no principals exist yet, the form surfaces an empty-state panel guiding the user to create their first principal, and emits events so the consuming app owns navigation.

## What's new

**`PrincipalsCreationGuide.vue`** (new field) — presentational empty-state panel shown when Kong Identity mode is selected and the directory has no principals. Shows a loading skeleton during lookup, then a "Create principal" + "Learn more" panel. Pure props-in / events-out:
- props: `loading`, `showPanel`
- emits: `click:create-principal`, `click:learn-more`

**Directory & principals lookup (`ConfigFormContent.vue`)** — on entering Kong Identity mode, a single `GET /v2/directories?pageSize=1` call:
- stores the resolved directory **name** on `config.principals.directory`,
- uses the directory **id** to check `GET /v2/directories/{id}/principals` for existing principals,
- shows the guide when none exist, hides it once principals are confirmed,
- disables the `error_on_miss` radios while the guide is visible (nothing to configure yet).
- On edit-load the stored directory is preserved; it's only adopted from the lookup on a genuine user switch into the mode.

**Event propagation** — `click:create-principal` / `click:learn-more` bubble `PrincipalsCreationGuide → ConfigFormContent → ConfigFormWithIdentity → PluginForm`/`PluginEntityForm`, so the host app handles navigation.

**i18n** — new strings: `configure_title`, `configure_description`, `create_principal`, `configure_learn_more`.

## Consuming app integration

```vue
<PluginForm
  :config="config"
  @click:create-principal="() => router.push(/* your principal-create route */)"
  @click:learn-more="(entity) => /* open docs for `entity`, e.g. 'kong-identity' */"
/>
```

Both events are optional; bind `@click:create-principal` wherever the guide can appear so the button is actionable.

## Tests

`ConfigFormContent.cy.ts` (+134): guide shown/hidden by principals existence, loading skeleton, `error_on_miss` disabled while the guide is visible, and `click:create-principal` emission.

### DIFF
--- a/packages/entities/entities-plugins/src/components/PluginForm.vue
+++ b/packages/entities/entities-plugins/src/components/PluginForm.vue
@@ -56,7 +56,7 @@
         :raw-schema="loadedSchema"
         :record="record"
         :schema="finalSchema"
-        @click:create-principal="() => $emit('click:create-principal')"
+        @click:create-entity="(payload: EntityCreateEvent) => $emit('click:create-entity', payload)"
         @click:learn-more="(entity: string) => $emit('click:learn-more', entity)"
         @global-action="(name: GlobalAction, payload: any) => $emit('globalAction', name, payload)"
         @loading="(val: boolean) => formLoading = val"
@@ -215,6 +215,7 @@ import {
   type PluginOrdering,
   type CustomSchemas,
   type PluginValidityChangeEvent,
+  type EntityCreateEvent,
 } from '../types'
 import PluginEntityForm from './PluginEntityForm.vue'
 import PluginFormActionsWrapper from './PluginFormActionsWrapper.vue'
@@ -252,7 +253,7 @@ const emit = defineEmits<{
   (e: 'showNewPartialModal', redisType: string): void
   (e: 'globalAction', name: GlobalAction, payload: any): void
   (e: 'click:learn-more', entity: string): void
-  (e: 'click:create-principal'): void
+  (e: 'click:create-entity', payload: EntityCreateEvent): void
 }>()
 
 // Component props - This structure must exist in ALL entity components, with the exclusion of unneeded action props (e.g. if you don't need `canDelete`, just exclude it)
--- a/packages/entities/entities-plugins/src/components/PluginForm.vue
+++ b/packages/entities/entities-plugins/src/components/PluginForm.vue
@@ -56,6 +56,8 @@
         :raw-schema="loadedSchema"
         :record="record"
         :schema="finalSchema"
+        @click:create-principal="() => $emit('click:create-principal')"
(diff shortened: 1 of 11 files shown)


packages/entities/entities-plugins/src/components/PluginForm.vue
```diff
@@ -249,6 +251,8 @@ const emit = defineEmits<{
   ): void
   (e: 'showNewPartialModal', redisType: string): void
   (e: 'globalAction', name: GlobalAction, payload: any): void
+  (e: 'click:learn-more', entity: string): void
+  (e: 'click:create-principal'): void
```
nekolab: I'm still wondering if it's too specific to use `click:create-principal` in `<PluginForm />`.
And this PR conflicts with #3476 if it gets merged.
Could you port the event `(e: 'click:create-entity', payload: EntityCreateEvent): void` to this PR?
You can use the event structure in https://github.com/Kong/public-ui-components/pull/3476#discussion_r3454537940

packages/entities/entities-plugins/src/components/fields/kong-identity/ConfigFormContent.vue
```diff
@@ -175,6 +252,19 @@ function hasActualRealm(realms: any[] | null | undefined): boolean {
...
 
+// Run the directory/principals lookup each time the user enters Kong Identity mode.
+// `prev === false` means a real user switch (create flow) so we adopt the directory name;
+// the immediate run (`prev` is undefined, i.e. edit-load) only checks principals and leaves
+// the stored directory intact. Declared after `selectedMode`/`hasPrincipals` because the
+// immediate run reads them synchronously during setup.
+watch(
+  () => isKonnect.value && hasPrincipals.value && selectedMode.value === 'kong-identity',
+  (active, prev) => {
+    if (active) fetchPrincipalsState({ setDirectory: prev === false })
```
nekolab: `fetchPrincipalsState` is an async function. We should avoid using async functions in a `watch` call.

A better approach might be to load the directory/principal data in `onMounted` and cache the result.

I also wonder whether `formData.config.principals.directory` will be cleared when `selectedMode` switches to a value other than `kong-identity`.

TT1228: Resolved. 
When switching away from `kong-identity`, principals object gets default value

nekolab (approved):
LGTM

PR #3476, opened 2026-06-22
feat(entities-plugins): add Kong Identity principals guide for OIDC plugin

TT1228:
# Summary

- OIDCPrincipals: query directories + principals on Konnect to detect principals existence; show guide banner when none exist
- Disable all Kong Identity fields while loading or when no principals exist
- Consolidate 3 create actions into click:create event (KongIdentityCreateEvent) with type: 'principal' | 'auth-server' | 'client'; authServerId on client
- Adopt directory name from API response instead of hardcoding 'default'
- Add leave-page KPrompt before emitting click:create for principal creation
- Emit click:learn-more with entity string; host app handles navigation
- OIDCForm, PluginEntityForm, PluginForm forward click:create + click:learn-more
- Remove createAuthServerUrl / createClientUrl from KonnectPluginFormConfig
- Add KongIdentityCreateEvent type export


packages/core/forms/src/components/forms/OIDCPrincipals.vue
```diff
@@ -690,13 +778,13 @@ export default {
...
         // eslint-disable-next-line vue/no-mutating-props
-        this.formModel['config-principals-directory'] = 'default'
-        // eslint-disable-next-line vue/no-mutating-props
         this.formModel['config-principals-error_on_miss'] = true
         // eslint-disable-next-line vue/no-mutating-props
         this.formModel['config-principals-match_consumer'] = true
         // eslint-disable-next-line vue/no-mutating-props
         this.formModel['config-principals-match_consumer_groups'] = true
+        // Re-check principals and adopt the directory name from the API (not hardcoded)
+        this.fetchPrincipalsState({ setDirectory: true })
```
nekolab, line 699: non-blocking: `this.fetchPrincipalsState` is an async function and calls the API endpoint every time. Using an async function here isn’t recommended because of timing issues and redundancy. We could keep a directory name cache in the component and fetch the directory name each time the component mounts.

nekolab (dismissed):
Left a comment. It's not a blocker, so we can merge the PR first and address it later

packages/entities/entities-plugins/src/types/plugin-form.ts
```diff
@@ -74,15 +74,21 @@ export interface KonnectPluginFormConfig extends BasePluginFormConfig, KonnectBa
...
 
 /** Kong Manager Plugin form config */
 export interface KongManagerPluginFormConfig extends BasePluginFormConfig, KongManagerBaseFormConfig { }
 
+/**
+ * Payload for the `click:create` event emitted from the OIDC/Kong Identity principals UI.
+ * The consuming app navigates to the matching create page; `authServerId` is provided when
+ * creating a client (scoped to the selected auth server).
+ */
+export interface KongIdentityCreateEvent {
```
nekolab: Maybe this event could use a more generic name, like `EntityCreateEvent`, and the definition could be made more generic as well.

```typescript
export type EntityCreateEvent = {
  type: 'principal'
} | {
  type: 'auth-server'
} | {
  type: 'client'
  authServerId?: string
}
```

packages/core/forms/src/components/forms/PrincipalLookupSettings.vue
```diff
@@ -0,0 +1,296 @@
...
+    handleEnableToggle(enabled) {
+      if (this.onEnabledChange) {
+        this.onEnabledChange(enabled)
+      }
+    },
+    getIdentifierClaimInputValue() {
+      const claim = this.formModel['config-principals-principal_claim']
+      if (Array.isArray(claim)) {
+        // Escape literal dots within each part before joining
+        return claim.map(part => part.replace(/\./g, '\\.')).join('.')
```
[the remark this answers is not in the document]
TT1228: Display-only encode/decode for a JWT claim-path field; the value is never used in an injection sink. The unescaped-backslash edge requires a literal \ inside a claim segment, which isn't a valid claim name (we expose \ as the dot-escape syntax). No security impact — dismissing as false positive.

hangrao (approved):

PR #3498, opened 2026-06-25
fix: Allow platform preset filters

filipgutica:
## Ticket
- https://konghq.atlassian.net/browse/MA-5159

## Summary
- Add `filterablePlatformPresetFilterDimensions`, copied from `kanalytics` `filterablePlatformDimensions`, for dashboard preset-filter validation.
- Update dashboard `preset_filters` to use one strict `filtersFn` schema with the shared explore filter fields plus platform preset filter fields in the same enum.
- Keep standalone platform query schemas loose while making dashboard preset filters use the combined strict allowlist.
- Add focused coverage for the combined preset-filter enum.

## Validation impact
Dashboard `preset_filters` now use a single strict allowlist. Platform preset filter fields are accepted through the same `filtersFn` schema as the existing explore preset filters, and arbitrary preset filter fields remain rejected by the enum.

## Verification
- `pnpm --filter @kong-ui-public/analytics-utilities exec vitest run src/dashboardSchema.v2.spec.ts`
- `pnpm --filter @kong-ui-public/analytics-utilities typecheck`
- `pnpm --filter @kong-ui-public/analytics-utilities lint`
- `pnpm --filter @kong-ui-public/analytics-utilities stylelint`
- `npx fallow audit --base origin/main -w @kong-ui-public/analytics-utilities --format json --quiet`
- `git diff --check`

### DIFF
--- a/packages/analytics/analytics-utilities/src/dashboardSchema.v2.spec.ts
+++ b/packages/analytics/analytics-utilities/src/dashboardSchema.v2.spec.ts
@@ -49,6 +49,26 @@ describe('dashboardSchema.v2', () => {
       ...filterableAgenticExploreDimensions,
     ]),
   ]
+  const platformPresetFilterableDimensions = [
+    'control_plane',
+    'gateway_service',
+    'realm',
+    'route',
+    'plugin',
+    'plugin_name',
+    'plugin_scope',
+    'data_plane_node_version',
+    'env',
+    'team',
+    'region',
+    'hostname',
+  ]
+  const presetFilterableDimensions = [
+    ...new Set([
+      ...sharedPresetFilterableDimensions,
+      ...platformPresetFilterableDimensions,
+    ]),
+  ]
 
   const platformQuery = {
     datasource: 'platform',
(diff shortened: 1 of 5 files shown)


packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
```diff
@@ -820,14 +820,19 @@ export const dashboardConfigSchema = {
...
+      anyOf: [
+        filtersFn([
+          ...new Set([
+            ...filterableExploreDimensions,
+            ...filterableBasicExploreDimensions,
+            ...filterableAiExploreDimensions,
+            ...filterableAgenticExploreDimensions,
+          ]),
+        ]),
+        platformFiltersFn(),
```
shifan-w: why would this need to be a new `platformFiltersFn()`.  Instead of another element in the Set().  `...filterablePlatformExploreDimensions`.

Is it because platformFilter is not a explore Dimension?

filipgutica: Because of what I mentioned in the slack thread. Platform  queries and filters have a looser schema where fields and dimensions are just strings, not constrained to an enum

packages/analytics/analytics-utilities/src/dashboardSchema.v2.spec.ts
```diff
@@ -49,6 +49,26 @@ describe('dashboardSchema.v2', () => {
       ...filterableAgenticExploreDimensions,
     ]),
   ]
+  const platformPresetFilterableDimensions = [
```
adorack: I think it'd be good to not have this duplicated here.  I'm not strictly sure that you need the test?  If you really want one, maybe just import the list like you do for the other datasources.

adorack (changes requested):

packages/analytics/analytics-utilities/src/dashboardSchema.v2.spec.ts
filipgutica: done

adorack (approved):

PR #3501, opened 2026-06-26
feat(redis): add inline Redis create panel

aanchalm01:
# Summary
WIP


aanchalm01:
Closing this PR since we are not going with inline Redis create panel instead Redis create option is going to be removed.

PR #3502, opened 2026-06-26
fix: table tiles encoded as chart tile types

filipgutica:
# Summary

Jira: MA-5164 - https://konghq.atlassian.net/browse/MA-5164

This pivots the unreleased dashboard table tile contract so table visuals remain dashboard chart tiles. Table rendering is now selected by `definition.chart.type: 'table'`, while the top-level tile `type` stays `'chart'`.

## Shape change

Previous unreleased shape:

```ts
{
  type: 'table',
  definition: {
    config: { title: 'Routes' },
    query: {
      datasource: 'platform',
      entity: 'route',
      columns: ['name', 'control_plane'],
    },
  },
}
```

New shape:

```ts
{
  type: 'chart',
  definition: {
    chart: { type: 'table', title: 'Routes' },
    query: {
      datasource: 'platform',
      entity: 'route',
      columns: ['name', 'control_plane'],
    },
  },
}
```

## Changes

- Updates `analytics-utilities` schema validation so table charts require the platform tabular query shape, while non-table charts keep the existing chart query shape.
- Routes `definition.chart.type === 'table'` through `DashboardTile` to `TableDataGridRenderer`.
- Preserves table chart shape through dashboard grid updates, duplication, and analytics-tier configuration utilities.
- Updates `DashboardTilePreview`, sandbox examples, and README docs to use the chart subtype contract.
- Keeps table charts out of CSV export and API Requests links, while allowing Explore links when context-link prerequisites are met.
- Keeps table previews with explicit `columns: []` in the not-configured state, while allowing omitted columns so `TableDataGridRenderer` can fall back to response metadata.

### DIFF
--- a/packages/analytics/dashboard-renderer/README.md
+++ b/packages/analytics/dashboard-renderer/README.md
@@ -16,7 +16,7 @@ Render Analytics charts on a page from a JSON definition.
   - [Tile Definition](#tile-definition)
   - [Chart Options](#chart-options)
   - [Chart Types](#chart-types)
-  - [Table Tile Configuration](#table-tile-configuration)
+  - [Table Chart Configuration](#table-chart-configuration)
   - [Common Chart Properties](#common-chart-properties)
   - [Query Configuration](#query-configuration)
   - [Time Range Configuration](#time-range-configuration)
@@ -29,7 +29,7 @@ Render Analytics charts on a page from a JSON definition.
 - A plugin providing an `AnalyticsBridge` must be installed in the root of the application.
   - This plugin must `provide` the necessary methods to adhere to the [AnalyticsBridge](https://github.com/Kong/public-ui-components/blob/main/packages/analytics/analytics-utilities/src/types/query-bridge.ts) interface defined in `@kong-ui-public/analytics-utilities`.
   - The plugin's query method is in charge of passing the query to the correct API for the host app's environment.
-  - For `table` tiles, the plugin must provide `tabularQueryFn`; the renderer passes a datasource-aware tabular query to that function and renders the returned records.
+  - For table chart tiles, the plugin must provide `tabularQueryFn`; the renderer passes a datasource-aware tabular query to that function and renders the returned records.
   - See the sandbox plugin [sandbox-query-provider.ts](https://github.com/Kong/public-ui-components/blob/main/packages/analytics/dashboard-renderer/sandbox/sandbox-query-provider.ts) for an example that simply returns static data rather than consuming an API.
 - The host application must supply peer dependencies for:
   - `@kong-ui-public/analytics-chart`
@@ -437,7 +437,7 @@ Configuration for individual dashboard tiles.
 
 ```typescript
 interface TileConfig {
-  type: 'chart' | 'table'      // The top-level tile kind
+  type: 'chart'                // The top-level tile kind
   definition: TileDefinition   // The tile's renderer-specific definition
   layout: TileLayout           // The tile's position and size in the grid
   id?: string                  // Optional unique identifier (required for editable dashboards)
@@ -448,15 +448,15 @@ interface TileConfig {
 Configuration for the renderer and query within a tile.
 
 ```typescript
-type TileDefinition = ChartTileDefinition | TableTileDefinition
+type TileDefinition = ChartTileDefinition | TableChartTileDefinition
 
 interface ChartTileDefinition {
-  chart: ChartOptions         // Configuration for the chart type and options
+  chart: ChartOptions         // Configuration for a non-table chart type and options
   query: ValidDashboardChartQuery  // Configuration for the chart data query
 }
 
-interface TableTileDefinition {
-  config: TableTileConfig     // Configuration for the table tile
+interface TableChartTileDefinition {
+  chart: TableChartOptions    // Configuration for the table chart
   query: PlatformTabularQuery & { datasource: 'platform' } // Configuration for the platform tabular query
 }
 ```
@@ -489,12 +489,13 @@ The following chart types are supported:
 
 Each chart type has its own configuration schema with specific options.
 
-### Table Tile Configuration
+### Table Chart Configuration
 
-Table tiles use `type: 'table'` on the tile itself. They are not chart tiles, so their definition uses `config` instead of `chart`.
+Table visuals are chart tiles with `type: 'chart'` on the tile itself and `definition.chart.type: 'table'`. Their `definition.query` uses the platform tabular query shape and renders through `TableDataGridRenderer`.
 
 ```typescript
-interface TableTileConfig {
+interface TableChartOptions {
+  type: 'table'
   title?: string
 }
 
@@ -507,17 +508,20 @@ interface PlatformTabularQuery {
 }
 ```
 
-`definition.query.columns` controls the visible table columns. If columns are omitted, the renderer can fall back to response `meta.columns` after the first tabular response. `definition.config.title` controls the tile title.
+`definition.query.columns` controls the visible table columns. If columns are omitted, the renderer can fall back to response `meta.columns` after the first tabular response. `definition.chart.title` controls the tile title.
 
 The dashboard config query includes `datasource: 'platform'`. The host application's `AnalyticsBridge.tabularQueryFn` receives a datasource-aware tabular query, currently `{ datasource: 'platform', query: { entity, columns, filters, page_size, cursor } }`.
 
 The renderer replaces raw record values with display labels when the tabular response includes a matching `meta.display[column][value].name`. Missing display entries and `null` values are rendered unchanged.
 
+Table charts do not expose CSV export or API Requests links. They may expose Explore links when the host application provides the Explore context-link bridge.
+
 ```typescript
 const tableTile: TileConfig = {
-  type: 'table',
+  type: 'chart',
   definition: {
-    config: {
+    chart: {
+      type: 'table',
       title: 'Platform routes',
     },
     query: {
@@ -575,7 +579,7 @@ Chart query fields include:
 - `granularity`: Time bucket size for temporal queries
 - `limit`: Number of results to return
 
-Table tile queries use the platform tabular query shape:
+Table chart queries use the platform tabular query shape:
 - `datasource`: Must be `platform`
 - `entity`: Entity collection to query, such as `route`
 - `columns`: Ordered table columns to request and render
--- a/packages/analytics/dashboard-renderer/README.md
+++ b/packages/analytics/dashboard-renderer/README.md
@@ -496,7 +496,7 @@ Table visuals are chart tiles with `type: 'chart'` on the tile itself and `defin
 ```typescript
 interface TableChartOptions {
   type: 'table'
-  title?: string
+  chart_title?: string
 }
 
 interface PlatformTabularQuery {
@@ -508,7 +508,7 @@ interface PlatformTabularQuery {
 }
 ```
 
-`definition.query.columns` controls the visible table columns. If columns are omitted, the renderer can fall back to response `meta.columns` after the first tabular response. `definition.chart.title` controls the tile title.
+`definition.query.columns` controls the visible table columns. If columns are omitted, the renderer can fall back to response `meta.columns` after the first tabular response. `definition.chart.chart_title` controls the tile title.
 
 The dashboard config query includes `datasource: 'platform'`. The host application's `AnalyticsBridge.tabularQueryFn` receives a datasource-aware tabular query, currently `{ datasource: 'platform', query: { entity, columns, filters, page_size, cursor } }`.
 
@@ -522,7 +522,7 @@ const tableTile: TileConfig = {
   definition: {
     chart: {
       type: 'table',
-      title: 'Platform routes',
+      chart_title: 'Platform routes',
     },
     query: {
       datasource: 'platform',
--- a/packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
+++ b/packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
@@ -152,7 +152,7 @@
     </div>
     <div
       class="tile-content"
-      :class="`type-${tileTypeClass}`"
+      :class="`type-${tileType}-${chart.type}`"
       :data-testid="`tile-content-${tileId}`"
     >
       <component
@@ -255,9 +255,8 @@ const chart = computed(() => props.definition.chart)
 const tileTitle = computed<string | undefined>(() => {
   return 'chart_title' in chart.value ? chart.value.chart_title : undefined
 })
-const tileDescription = computed<string | undefined>(() => !isTableChartDefinition(props.definition) && 'description' in chart.value ? chart.value.description : undefined)
-const tileTypeClass = computed<string>(() => isTableChartDefinition(props.definition) ? 'table' : chart.value.type)
-const isSlottableTile = computed<boolean>(() => !isTableChartDefinition(props.definition) && chart.value.type === 'slottable')
+const tileDescription = computed<string | undefined>(() => 'description' in chart.value ? chart.value.description : undefined)
+const isSlottableTile = computed<boolean>(() => chart.value.type === 'slottable')
 const canExportCsv = computed<boolean>(() => {
   if (isTableChartDefinition(props.definition)) {
     return false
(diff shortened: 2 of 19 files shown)


packages/analytics/dashboard-renderer/src/components/DashboardTilePreview.vue
```diff
@@ -89,7 +90,17 @@ const { internalContext, queryReady } = composables.useDashboardInternalContext(
...
+const isTableDefinition = (definition: TileDefinition): boolean => {
+  return 'config' in definition
+}
+
+const tileType = computed<'chart' | 'table'>(() => isTableDefinition(definition) ? 'table' : 'chart')
+
 const chartNotConfigured = computed(() => {
+  if (isTableDefinition(definition)) {
+    return false
+  }
```
jordojordo: I suppose requiring any validation of a table tile outside of this component is fine, just worth noting.

jordojordo (dismissed):

packages/analytics/dashboard-renderer/src/components/DashboardTilePreview.vue
```diff
@@ -89,7 +90,17 @@ const { internalContext, queryReady } = composables.useDashboardInternalContext(
...
 })
 
+const isTableDefinition = (definition: TileDefinition): boolean => {
+  return 'config' in definition
+}
+
+const tileType = computed<'chart' | 'table'>(() => isTableDefinition(definition) ? 'table' : 'chart')
+
 const chartNotConfigured = computed(() => {
+  if (isTableDefinition(definition)) {
```
Chalks: **nit, maybe blocker**

shouldn't this be:


```suggestion
  if (tileType.value === 'table' && !isTableDefinition(definition)) {
```

adorack: (Agree)

This also makes me think that maybe the "not configured" warnings should maybe be part of the chart tile.  Not sure without poking at it a bit more how hard that would be, though.

packages/analytics/dashboard-renderer/src/components/DashboardTilePreview.vue
```diff
@@ -89,7 +90,17 @@ const { internalContext, queryReady } = composables.useDashboardInternalContext(
   preview: computed(() => preview),
 })
 
+const isTableDefinition = (definition: TileDefinition): boolean => {
+  return 'config' in definition
+}
+
+const tileType = computed<'chart' | 'table'>(() => isTableDefinition(definition) ? 'table' : 'chart')
```
adorack: Thinking about it a bit more -- something is wrong with the architecture here if you're trying to guess whether it's a table or chart from the definition.  Taking a look at how it's used in Konnect...

adorack (changes requested):

packages/analytics/dashboard-renderer/src/components/DashboardTilePreview.vue
filipgutica: Yeah, refactoring it as we discussed in the slack thread.

Chalks (changes requested):
placeholder, I want to review this.

packages/analytics/dashboard-renderer/src/utils/duplicate-tile.ts
```diff
@@ -1,68 +1,59 @@
...
- * @returns A table tile config ready to be inserted into the dashboard.
+ * @returns A chart tile config ready to be inserted into the dashboard.
  */
-export const duplicateTableTile = (tile: GridTile<TileDefinition>): TileConfig => {
-  const tableMeta = tile.meta as TableTileDefinition
-  const config = {
-    ...tableMeta.config,
-    title: tableMeta.config.title ? `Copy of ${tableMeta.config.title}` : '',
-  }
+export const duplicateChartTile = (tile: GridTile<TileDefinition>): TileConfig => {
```
Chalks, line 14: **preference, maybe a blocker**

I found this whole function challenging to read. Strongly prefer structuring it so that it has only a single return, like so:

```
export const duplicateChartTile = (tile: GridTile<TileDefinition>): TileConfig => {
  const { chart } = tile.meta

  let updatedChart = undefined
  if (chart.type === 'table') {
    updatedChart = {
      ...chart,
      title: chart.title ? `Copy of ${chart.title}` : '',
    }
  } else if (chart.type === 'slottable') {
    updatedChart = { ...chart }
  } else {
    updatedChart = {
      ...chart,
      chart_title: chart.chart_title ? `Copy of ${chart.chart_title}` : '',
    }
  }

  return {
    id: crypto.randomUUID(),
    type: 'chart',
    definition: {
      ...tile.meta,
      chart: updatedChart,
    },
    layout: {
      position: {
        col: 0,
        row: 0,
      },
      size: tile.layout.size,
    },
  }
}
```

this pretty significantly reduces the complexity and makes it clear that the only thing really happening here is handling title updates.

Chalks (dismissed):
approved, but one very strong preference below.

packages/analytics/dashboard-renderer/src/components/DashboardTilePreview.vue
Chalks: resolved as no longer relevant

Chalks (dismissed):

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -251,13 +252,13 @@ const exportModalVisible = ref<boolean>(false)
...
-const isTableTile = computed((): boolean => props.tileType === 'table')
+const isTableTile = computed((): boolean => isTableChartDefinition(props.definition))
 
 const chartDefinition = computed<ChartTileDefinition>(() => props.definition as ChartTileDefinition)
-const tableDefinition = computed<TableTileDefinition>(() => props.definition as TableTileDefinition)
+const tableDefinition = computed<TableChartTileDefinition>(() => props.definition as TableChartTileDefinition)
 const chart = computed(() => chartDefinition.value.chart)
 const chartTitle = computed<string | undefined>(() => 'chart_title' in chart.value ? chart.value.chart_title : undefined)
-const tileTitle = computed<string | undefined>(() => isTableTile.value ? tableDefinition.value?.config.title : chartTitle.value)
+const tileTitle = computed<string | undefined>(() => isTableTile.value ? tableDefinition.value.chart.title : chartTitle.value)
```
adorack: This should now be standardized, so you don't need to figure it out separately for tables, right?

packages/analytics/dashboard-renderer/src/utils/duplicate-tile.ts
```diff
@@ -1,69 +1,46 @@
...
+    duplicatedDefinition = {
+      ...tile.meta,
+      chart: { ...tile.meta.chart },
+    }
+  } else {
+    duplicatedDefinition = {
+      ...tile.meta,
+      chart: {
+        ...tile.meta.chart,
+        chart_title: tile.meta.chart.chart_title ? `Copy of ${tile.meta.chart.chart_title}` : '',
```
adorack: Pretty sure tables use `title` now, meaning they're the same as regular charts?

packages/analytics/dashboard-renderer/README.md
```diff
@@ -507,17 +508,20 @@ interface PlatformTabularQuery {
...
 ```
 
-`definition.query.columns` controls the visible table columns. If columns are omitted, the renderer can fall back to response `meta.columns` after the first tabular response. `definition.config.title` controls the tile title.
+`definition.query.columns` controls the visible table columns. If columns are omitted, the renderer can fall back to response `meta.columns` after the first tabular response. `definition.chart.title` controls the tile title.
 
 The dashboard config query includes `datasource: 'platform'`. The host application's `AnalyticsBridge.tabularQueryFn` receives a datasource-aware tabular query, currently `{ datasource: 'platform', query: { entity, columns, filters, page_size, cursor } }`.
 
 The renderer replaces raw record values with display labels when the tabular response includes a matching `meta.display[column][value].name`. Missing display entries and `null` values are rendered unchanged.
 
+Table charts do not expose CSV export or API Requests links. They may expose Explore links when the host application provides the Explore context-link bridge.
```
adorack, line 517: They will expose CSV export, right?

filipgutica: They will yes.

packages/analytics/dashboard-renderer/src/utils/duplicate-tile.ts
filipgutica: tables use `title` but charts use `chart_title`

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
filipgutica: tables use title, charts use chart_title

packages/analytics/dashboard-renderer/src/utils/duplicate-tile.ts
filipgutica: done

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
filipgutica: done

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -310,14 +306,14 @@ watch(() => props.definition, async (newValue, oldValue) => {
...
-const kebabMenuHasItems = computed((): boolean => isTableTile.value
+const kebabMenuHasItems = computed((): boolean => isTableChartDefinition(props.definition)
   ? !!exploreLinkKebabMenu.value || props.context.editable
   : !!exploreLinkKebabMenu.value || canExportCsv.value || props.context.editable)
 
 // The shared header action container is hidden when tile actions are globally disabled.
 const canShowHeaderActions = computed((): boolean => !props.hideActions && canShowKebabMenu.value && kebabMenuHasItems.value)
 const hasHeaderActions = computed<boolean>(() => canShowHeaderActions.value && kebabMenuHasItems.value && !props.isFullscreen)
-const hasSignalsDescription = computed<boolean>(() => !isTableTile.value && chart.value.type === 'golden_signals' && Boolean(tileDescription.value))
+const hasSignalsDescription = computed<boolean>(() => !isTableChartDefinition(props.definition) && chart.value.type === 'golden_signals' && Boolean(tileDescription.value))
```
adorack: Why does this need to assert it's not a table chart?  Isn't saying `type === 'golden_signals'` enough?

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -310,14 +306,14 @@ watch(() => props.definition, async (newValue, oldValue) => {
 
 const csvFilename = computed<string>(() => i18n.t('csvExport.defaultFilename'))
 
-const kebabMenuHasItems = computed((): boolean => isTableTile.value
+const kebabMenuHasItems = computed((): boolean => isTableChartDefinition(props.definition)
```
adorack: Why does this need to do something different for tables?  They can't export CSVs, so `canExportCsv` is false, and when they can export CSVs, won't you want this to be the same path as other charts?

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -252,18 +250,16 @@ const exportModalVisible = ref<boolean>(false)
...
-const tileTitle = computed<string | undefined>(() => isTableTile.value ? tableDefinition.value.chart.title : chartTitle.value)
-const tileDescription = computed<string | undefined>(() => !isTableTile.value && 'description' in chart.value ? chart.value.description : undefined)
-const tileTypeClass = computed<string>(() => isTableTile.value ? 'table' : chart.value.type)
-const isSlottableTile = computed<boolean>(() => !isTableTile.value && chart.value.type === 'slottable')
+
+const chart = computed(() => props.definition.chart)
+const tileTitle = computed<string | undefined>(() => {
+  return 'chart_title' in chart.value ? chart.value.chart_title : undefined
+})
+const tileDescription = computed<string | undefined>(() => !isTableChartDefinition(props.definition) && 'description' in chart.value ? chart.value.description : undefined)
```
adorack: Since table charts can't have a description, this doesn't need to do anything special for tables, right?

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -252,18 +250,16 @@ const exportModalVisible = ref<boolean>(false)
...
+const tileTypeClass = computed<string>(() => isTableChartDefinition(props.definition) ? 'table' : chart.value.type)
+const isSlottableTile = computed<boolean>(() => !isTableChartDefinition(props.definition) && chart.value.type === 'slottable')
```
adorack: This doesn't need to check for table, does it?

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -341,12 +337,14 @@ const componentEventHandlers = computed(() => ({
 }))
 
 const componentData = computed(() => {
-  if (isTableTile.value) {
+  const definition = props.definition
+
+  if (isTableChartDefinition(definition)) {
```
adorack: Is it possible for this to be the same as the regular chart branch?  It looks like the difference is that regular charts set a height -- is it important that tables don't?  Given how much of the rest is the same, I think I'd advocate for splitting out `height` if it's important.

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -439,15 +436,15 @@ const chartDataGranularity = computed(() => {
 })
 
 const isTimeSeriesChart = computed(() => {
-  return !isTableTile.value && ['timeseries_line', 'timeseries_bar'].includes(chart.value.type)
+  return !isTableChartDefinition(props.definition) && ['timeseries_line', 'timeseries_bar'].includes(chart.value.type)
```
adorack: I don't think this needs to do anything special for tables.

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -439,15 +436,15 @@ const chartDataGranularity = computed(() => {
...
 })
 
 const isAgedOutQuery = computed(() => {
-  if (!isTimeSeriesChart.value || !props.queryReady || loadingChartData.value) {
+  if (isTableChartDefinition(props.definition) || !isTimeSeriesChart.value || !props.queryReady || loadingChartData.value) {
```
adorack, line 431: This doesn't need to do anything special for tables.

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -458,7 +455,7 @@ const isAgedOutQuery = computed(() => {
 
 const agedOutWarning = computed(() => {
   const currentGranularity = msToGranularity(chartData.value?.meta.granularity_ms ?? 0) ?? 'unknown'
-  const savedGranularity = chartDefinition.value.query.granularity ?? 'unknown'
+  const savedGranularity = isTableChartDefinition(props.definition) ? 'unknown' : props.definition.query.granularity ?? 'unknown'
```
adorack, line 447: This doesn't need to do anything special for tables (they don't have `query.granularity`).  Is there something here about appeasing Typescript?

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
```diff
@@ -472,14 +469,14 @@ const agedOutWarning = computed(() => {
  * @returns Array of scoped filter objects to a datasource
  */
 const datasourceScopedFilters = computed(() => {
-  if (isTableTile.value) {
+  if (isTableChartDefinition(props.definition)) {
```
adorack: Wait, we want tables to apply filters.  Maybe this should be setting `datasource` to `platform` if it's a table chart, or something?

packages/analytics/dashboard-renderer/src/utils/duplicate-tile.ts
adorack: Please get rid of the extra table-focused branch.  (It looks like you got rid of the `slottable` branch, which I imagine is OK because you can't duplicate slottable tiles, but maybe either leave that alone for now or throw an exception for the sake of clarity?

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
filipgutica: done

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
filipgutica: done

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
filipgutica: done

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
filipgutica: got rid of tileTypeClass and yes

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
filipgutica: done

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
filipgutica: yes this one is for appeasing typescript, added a comment

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
filipgutica: This one is also for appeasing typescript, added a comment.

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
filipgutica: This is the chart zoom-action path, not table fetching or the kebab Explore link. Table fetching already sanitizes filters inside the table fetcher, and table charts do not emit `select-chart-range`.

The actual table Explore-link filter fix is in `useContextLinks`, where we preserve the tabular query shape and include sanitized filters in the `c=table` Explore URL.

packages/analytics/dashboard-renderer/src/components/DashboardTile.vue
adorack: Now that this is setting `height` the same, can it be the same branch as the regular chart?

filipgutica: done

adorack (approved):

PR #3505, opened 2026-06-29
chore(entities-plugins): kong identity updates

TT1228:
## Summary
- **forms**: Show the Identifier claim input in both principal lookup modes (`principal_claim` is used by the gateway in both OIDC and custom-identity lookups). Infer the initial mode from `principal_by` alone, make the help text mode-aware, and drop the `sub`-only wording.
- **entities-plugins**: Always show the Centrally Managed option when the schema supports `identity_realms`, even with no realms yet. Hiding it is deferred to a platform-wide fast-follow.

## Tests
- Updated `PrincipalLookupSettings.spec.ts` (new `principal_claim`-only → kong-identity case).
- Updated `ConfigFormContent.cy.ts` to assert Centrally Managed shows with no existing realms.

### DIFF
--- a/packages/entities/entities-plugins/src/components/fields/kong-identity/KongIdentityField.vue
+++ b/packages/entities/entities-plugins/src/components/fields/kong-identity/KongIdentityField.vue
@@ -84,7 +84,6 @@ defineOptions({ name: 'KongIdentityField' })
 
 const props = defineProps<{
   identityRealmsInSchema?: boolean
-  hasExistingRealms?: boolean
   loadingRealms?: boolean
 }>()
 const { i18n } = composables.useI18n()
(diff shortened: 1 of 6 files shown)


kingyue737 (dismissed):

packages/entities/entities-plugins/src/components/fields/kong-identity/KongIdentityField.vue
kingyue737: ~~Do we still need to keep this prop?~~

intentionally retain according to the comment

TT1228: Good catch, resolving it

kingyue737 (approved):

PR #3507, opened 2026-06-29
fix(entities): background color fixes [KHCP-20980]

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-20980

Fix background color in elements in entities packages

### DIFF
--- a/packages/entities/entities-plugins/src/components/PluginForm.vue
+++ b/packages/entities/entities-plugins/src/components/PluginForm.vue
@@ -1663,6 +1663,10 @@ async function getClonedPlugin(clonedPluginName: string): Promise<void> {
 .kong-ui-entities-plugin-form-container {
   width: 100%;
 
+  :deep(.kong-ui-entity-base-form) {
+    background-color: var(--kui-color-background-transparent, $kui-color-background-transparent);
+  }
+
   .form-action-button {
     margin-left: var(--kui-space-60, $kui-space-60);
   }
--- a/packages/entities/entities-plugins/src/components/PluginForm.vue
+++ b/packages/entities/entities-plugins/src/components/PluginForm.vue
@@ -1700,10 +1700,6 @@ async function getClonedPlugin(clonedPluginName: string): Promise<void> {
   }
 
   &.new-form-layout {
-    :deep(.kong-ui-entity-base-form) {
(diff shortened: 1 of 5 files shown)


packages/entities/entities-plugins/src/components/PluginForm.vue
```diff
@@ -1700,6 +1700,10 @@ async function getClonedPlugin(clonedPluginName: string): Promise<void> {
   }
 
   &.new-form-layout {
+    :deep(.kong-ui-entity-base-form) {
+      background-color: var(--kui-color-background-transparent, $kui-color-background-transparent);
+    }
```
adamdehaven: shouldn't this be fixed in the base form instead of here?

packages/entities/entities-shared/src/components/entity-form-block/EntityFormBlock.vue
```diff
@@ -121,7 +121,7 @@ const slots = defineSlots<{
   }
 
   .content {
-    background: var(--kui-color-background-neutral-weakest, $kui-color-background-neutral-weakest);
+    background: var(--kui-color-background, $kui-color-background);
```
adamdehaven, line 124: this would impact old UX as well; is that safe?

portikM: It's going to change the elements background in the old nav but I don't think it breaks it. I say it's safe.

Before
Image

After
Image

packages/entities/entities-plugins/src/components/PluginForm.vue
portikM: Good call -- fixed. Checked both old and new UX.

packages/entities/entities-shared/src/components/entity-form-block/EntityFormBlock.vue
adamdehaven: What does it do in new nav?

portikM: It adds background color. Otherwise those step containers are transparent.

Screenshot 2026-06-30 at 8 11 04 AM

adamdehaven (approved):

jillztom (approved):

PR #3510, opened 2026-06-30
feat(entities-plugins): add unauthorization guard for kong identity

TT1228:
# Summary

Guard KI related queries, treating 401s as feature off.

### DIFF
--- a/packages/core/forms/src/components/forms/OIDCForm.vue
+++ b/packages/core/forms/src/components/forms/OIDCForm.vue
@@ -34,7 +34,6 @@
             :on-model-updated="onModelUpdated"
             @click:create-entity="(payload) => $emit('click:create-entity', payload)"
             @click:learn-more="(entity) => $emit('click:learn-more', entity)"
-            @kong-identity-unavailable="handleKongIdentityUnavailable"
             @mode-change="handlePrincipalsModeChange"
           >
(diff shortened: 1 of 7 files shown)


packages/core/forms/src/components/forms/OIDCForm.vue
```diff
@@ -295,9 +295,10 @@ export default {
       return (this.formModel.id && this.isEditing) || !this.isEditing
     },
     hasPrincipalsFields() {
-      return this.identityPrincipalsUiEnabled && this.formSchema.fields?.some(
-        (field) => PRINCIPALS_FIELD_MODELS.has(field.model),
-      )
+      return this.identityPrincipalsUiEnabled
+        && this.formsConfig?.isKongIdentityAuthServersAvailable !== false
```
nekolab, line 299: The `hasPrincipalsFields` field is a bit misleading. As the feature has evolved, it no longer only controls Principal Lookup, the Kong Identity auth server is also configured through this component.

The current logic acts like feature flags: this component is rendered only when both FFs are enabled. But Kong Identity Principal Lookup and the Auth Server are actually two separate features, so they should be controlled independently.

I already discussed this with @TT1228  offline. We’ll check with design on how the UI should look when these are toggled separately, and I’ll address this in a follow-up PR.

nekolab (approved):
LGTM

PR #3518, opened 2026-07-01
feat(dashboard-renderer): add markdown as first-class tile type

sichvoge:
# Summary

Adds markdown as a first-class tile type in the analytics dashboard schema and renderer.

- Schema (analytics-utilities): adds markdownTileDefinitionSchema, MarkdownTileDefinition, markdownTileConfigSchema, and MarkdownTileConfig types. The top-level tileConfigSchema is now a oneOf: [chartTileConfigSchema, markdownTileConfigSchema] discriminated union. All types are derived via FromSchemaWithOptions so the schema and types stay in sync.
- Renderer (dashboard-renderer): DashboardRenderer.vue routes tile.type === 'markdown' to a new MarkdownTile component. MarkdownTile.vue renders the header chrome (edit / kebab with duplicate+delete), delegating content to MarkdownTileRenderer.vue which uses @kong/markdown for safe rendering. Tier-aware configure-definition.ts now short-circuits markdown tiles (no datasource to migrate or filter).
- Tests: unit tests for the updated tileConfigSchema (AJV validation), component tests for MarkdownTileRenderer including XSS injection guards (script tags, inline event handlers).

image

### DIFF
--- a/packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
+++ b/packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
@@ -799,7 +799,42 @@ export const chartTileConfigSchema = {
   additionalProperties: false,
 } as const satisfies JSONSchema
 
-export const tileConfigSchema = chartTileConfigSchema
+export type ChartTileConfig = FromSchemaWithOptions<typeof chartTileConfigSchema>
+
+export const markdownTileDefinitionSchema = {
+  type: 'object',
+  properties: {
+    content: { type: 'string' },
+  },
+  required: ['content'],
+  additionalProperties: false,
+} as const satisfies JSONSchema
+
+export type MarkdownTileDefinition = FromSchemaWithOptions<typeof markdownTileDefinitionSchema>
+
+export const markdownTileConfigSchema = {
+  type: 'object',
+  properties: {
+    type: {
+      type: 'string',
+      enum: ['markdown'],
+    },
+    definition: markdownTileDefinitionSchema,
+    layout: tileLayoutSchema,
+    id: {
+      type: 'string',
+      description: 'Unique identifier for the tile.  If not provided, one will be generated.',
+    },
+  },
+  required: ['type', 'definition', 'layout'],
+  additionalProperties: false,
+} as const satisfies JSONSchema
+
+export type MarkdownTileConfig = FromSchemaWithOptions<typeof markdownTileConfigSchema>
+
+export const tileConfigSchema = {
+  oneOf: [chartTileConfigSchema, markdownTileConfigSchema],
+} as const satisfies JSONSchema
 
 export type TileConfig = FromSchemaWithOptions<typeof tileConfigSchema>
 
--- a/packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
+++ b/packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
@@ -33,7 +33,13 @@
         <template #tile="{ tile }">
           <!-- eslint-disable @kong/eslint-plugin-design-tokens/token-constant-requires-css-var -->
           <div
-            v-if="isSlottableTile(tile)"
+            v-if="tile.type === 'markdown'"
+            class="tile-container"
+          >
+            <MarkdownTileRenderer :content="(tile.meta as any).content" />
+          </div>
+          <div
+            v-else-if="isSlottableTile(tile)"
             class="tile-container slottable-tile"
           >
             <slot :name="getSlottableSlotName(tile)" />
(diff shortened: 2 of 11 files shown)


packages/analytics/dashboard-renderer/src/components/DashboardRenderer.vue
```diff
@@ -136,7 +149,23 @@ const tileSortFn = (a: TileConfig, b: TileConfig) => {
...
-  return model.value.tiles.map((tile: TileConfig) => {
+  return model.value.tiles.map((tile: TileConfig): GridTile<TileDefinition> => {
+    if (tile.type === 'markdown') {
+      if (internalContext.value.editable && !tile.id) {
+        console.warn(
+          'No id provided for tile. One will be generated automatically,',
+          'however tracking changes to this tile may not work as expected.',
+          tile,
+        )
+      }
```
filipgutica, line 160: hmm we're doing this check and logging this warning in two spots now. 

I think this whole `if tile.type === 'markdown'` block is redundant actually and can just continue using the same code path

packages/analytics/dashboard-renderer/src/components/MarkdownTile.vue
```diff
@@ -0,0 +1,130 @@
...
+            <KDropdownItem
+              :data-testid="`remove-tile-${tileId}`"
+              @click="emit('remove-tile')"
+            >
+              <span class="delete-option">{{ i18n.t('renderer.delete') }}</span>
+            </KDropdownItem>
+          </template>
+        </KDropdown>
+      </div>
+    </div>
```
filipgutica, line 46: This whole block is almost identical to the tile-header block in `DashboardTile` we should extract it to a re-usable component, re-used in both spots.

packages/analytics/dashboard-renderer/src/components/MarkdownTile.vue
```diff
@@ -0,0 +1,130 @@
...
+    <div class="tile-content">
+      <MarkdownTileRenderer :content="content" />
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
```
filipgutica, line 53: This whole component is very similar to DashboardTile which is why I was suggesting to keep using DashboardTile and have it render MarkdownTileRenderer rather than introducing a new, near-duplicate MarkdownTile component. 

I think if we want to go down this path, we should extract the common base tile behaviour to a TileShell component that both MarkDownTile and DashboardTile use/inherit from.

packages/analytics/dashboard-renderer/src/components/MarkdownTileRenderer.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+  <div class="markdown-tile-renderer">
+    <MarkdownUi
+      :editable="false"
+      mode="read"
+      :model-value="content"
+      theme="light"
```
filipgutica, line 7: are we just hardcoding theme="light" ? 

Just calling this out for the future when we want to support themes we'll have to update this so it's dynamic/inherited from the parent/host app some how.

filipgutica (changes requested):
Left some comments about how the components are organized. I think if we want to ship this we should go with the first proposal you had with a shared <TileShell> component to avoid duplicate logic/code between MarkdownTile and DashboardTile. 

Could you also update the sandbox so we can play around with and see what markdown tiles in the dashboard look/feel like?

packages/analytics/dashboard-renderer/src/components/MarkdownTileRenderer.vue
sichvoge: Yes. I looked at `LabDetailContent.vue`, `ApiDocumentation.vue`, and `APIPackagesDocumentation.vue` inside konnect-ui-apps who also use this package and they all hardcoded it. How do we keep track on this best?

packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
```diff
@@ -799,7 +799,42 @@ export const chartTileConfigSchema = {
   additionalProperties: false,
 } as const satisfies JSONSchema
 
-export const tileConfigSchema = chartTileConfigSchema
+export type ChartTileConfig = FromSchemaWithOptions<typeof chartTileConfigSchema>
+
+export const markdownTileDefinitionSchema = {
+  type: 'object',
+  properties: {
+    content: { type: 'string' },
```
Chalks, line 807: **recommendation, potentially blocker**

I think it's reasonable to add a character limit here. Slack limits to 12,000 characters, reddit comments limit to 10,000, etc.

Particularly since we're storing this in the DB alongside the config itself.

PR #3523, opened 2026-07-02
fix: refresh table tiles on query changes

filipgutica:
## Summary
- Refresh table data when effective table query/context inputs change
- Keep normal pagination behavior for pagination-only changes by keying refreshes from query-affecting inputs
- Add regression coverage for entity, columns, filters/context, cursor reset, and non-query prop churn

### DIFF
--- a/packages/analytics/dashboard-renderer/src/components/TableDataGridRenderer.vue
+++ b/packages/analytics/dashboard-renderer/src/components/TableDataGridRenderer.vue
@@ -14,7 +14,7 @@
       :fetcher="fetcher"
       :headers="headers"
       :page-size="query.page_size"
-      :refresh-key="refreshCounter"
+      :refresh-key="tableRefreshKey"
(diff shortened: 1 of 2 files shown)


packages/analytics/dashboard-renderer/src/components/TableDataGridRenderer.vue
```diff
@@ -65,6 +65,21 @@ const datasourceAwareQuery = computed<DatasourceAwareTabularQuery>(() => ({
     page_size: props.query.page_size,
   },
 }))
+const tableRefreshKey = computed(() => JSON.stringify({
```
jordojordo, line 68: Not a blocker, this is probably fine as is, but I think in the future we should add in the abort controller into this key. Whenever the key changes the previous fetches should be canceled.

jordojordo (approved):

PR #3526, opened 2026-07-04
fix: prune tile filters by query fields

filipgutica:
## Summary

Closes [MA-5180](https://konghq.atlassian.net/browse/MA-5180)

- Add `queryFields` as the preferred `stripUnknownFilters` option while keeping `metrics` as a deprecated alias.
- Pass platform table tile `entity` as the query field when pruning table request filters.
- Update dashboard-renderer call sites to pass selected metric fields through `queryFields` instead of the deprecated `metrics` option.

## Why

`metrics` was too narrow as an option name. Filter availability is not affected only by selected metric fields; any datasource field with `supportedDimensions` can constrain which filters are valid. Table tiles expose that through `entity`, so a platform table query for `control_plane` should prune dashboard/context filters against the `control_plane` field's supported dimensions and drop unsupported filters such as `consumer`.

Keeping `metrics` as a deprecated alias preserves existing callers while making the new contract clearer for both metric-backed charts and entity-backed table queries.

[MA-5180]: https://konghq.atlassian.net/browse/MA-5180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics/dashboard-renderer/src/utils/table-data-grid-renderer.ts
+++ b/packages/analytics/dashboard-renderer/src/utils/table-data-grid-renderer.ts
@@ -122,6 +122,7 @@ export const tableDataGridFetcherByDatasource = {
         ...((platformQuery.filters ?? []) as AllFilters[]),
         ...context.filters,
       ],
+      queryFields: platformQuery.entity === undefined ? undefined : [platformQuery.entity],
     }) as PlatformTabularQuery['filters']
 
     const response = await tabularQueryFn({
--- a/packages/analytics/analytics-config-store/src/stores/datasource-config.ts
+++ b/packages/analytics/analytics-config-store/src/stores/datasource-config.ts
@@ -112,29 +112,34 @@ export const useDatasourceConfigStore = defineStore('datasource-config', () => {
     return ({
       datasource,
       filters,
+      queryFields = undefined,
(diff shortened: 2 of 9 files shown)


packages/analytics/dashboard-renderer/src/utils/table-data-grid-renderer.ts
```diff
@@ -122,6 +122,7 @@ export const tableDataGridFetcherByDatasource = {
         ...((platformQuery.filters ?? []) as AllFilters[]),
         ...context.filters,
       ],
+      queryFields: platformQuery.entity === undefined ? undefined : [platformQuery.entity],
```
filipgutica, line 125: This line is the actual fix. Rest of the changes are just renaming the argument "metrics" to "queryFields"

jordojordo (commented):
The changes look good, my only concern is the naming scheme of `queryFields`. That seems too broad a name for what's actually being passed, and considering that the filters and dimensions are part of the query, one could assume they are also "query fields". Maybe something a bit more specific like `selectedFields`. I'm open to anything though.

jordojordo (approved):

PR #3536, opened 2026-07-06
feat(entities-plugins): show data plane not available alert

TT1228:
# Summary

Add alert when data plane not supporting KI, and update copy accordingly.


nekolab (approved):
limited scenarios manually tested and codewise LGTM

packages/entities/entities-plugins/src/types/plugin-form.ts
```diff
@@ -80,16 +80,45 @@ export interface KonnectPluginFormConfig extends BasePluginFormConfig, KonnectBa
...
+   * `true`; there is no other circumstance where it should be. `undefined` means not yet
+   * resolved (drives a loading state).
+   */
+  principalsCreationGuideVisible?: boolean
+  /**
+   * Kong Gateway versions of the data plane nodes connected to the current control plane
+   * (deduped). Used to warn when Kong Identity principals — which require Gateway 3.15+ —
+   * are configured but a connected DP node can't process them.
+   */
+  dataPlaneVersions?: string[]
```
2eha0, line 121: nit: I think we need a way to manage these OIDC-specific configs, instead of letting the plugin form config expand infinitely.

TT1228: I will start some refactoring. I expect design updates are on the way.

PR #3539, opened 2026-07-06
 fix(analytics): ensure old data is shown when query is aborted [MA-5146]

jordojordo:
# Summary

Fix [MA-5146](https://konghq.atlassian.net/browse/MA-5146)

A few issues have been fixed here:
- In the `QueryDataProvider`, queries were not being properly aborted when a new query has been issued which causes undue stress to the backend and can also create some odd behavior on the frontend when an older query eventually finishes and overwrites a new one that fails.
- Requests that timeout were showing a generic error message rather than the proper "Timed out" message already in place.

[MA-5146]: https://konghq.atlassian.net/browse/MA-5146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics/dashboard-renderer/src/components/QueryDataProvider.vue
+++ b/packages/analytics/dashboard-renderer/src/components/QueryDataProvider.vue
@@ -125,7 +125,9 @@ const hasError = computed(() => state.value === STATE.ERROR || !!queryError.valu
 const isLoading = computed(() => !props.queryReady || state.value === STATE.PENDING)
 
 const oldData = ref<ExploreResultV4 | undefined>()
-const displayData = computed(() => v4Data.value ?? oldData.value)
+const displayData = computed(() => {
+  return v4Data.value ?? (props.queryReady && state.value === STATE.PENDING ? oldData.value : undefined)
+})
 
 watch([() => v4Data.value, () => state.value], ([data, state]) => {
   if (data && (state === 'SUCCESS_HAS_DATA' || state === 'SUCCESS') && data !== oldData.value) {
--- a/packages/analytics/dashboard-renderer/src/components/QueryDataProvider.vue
+++ b/packages/analytics/dashboard-renderer/src/components/QueryDataProvider.vue
@@ -1,11 +1,11 @@
 <template>
   <KSkeleton
-    v-if="isLoading || (!v4Data && !hasError)"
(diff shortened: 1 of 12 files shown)


packages/analytics/dashboard-renderer/src/components/QueryDataProvider.vue
```diff
@@ -1,11 +1,11 @@
 <template>
   <KSkeleton
-    v-if="isLoading || (!v4Data && !hasError)"
+    v-if="(isLoading || (!v4Data && !hasError)) && !displayData"
```
filipgutica, line 3: I think there’s a state this misses: after a tile has loaded data once, `oldData` stays populated. 
When `queryReady` becomes false, `queryKey()` returns `null`, so SWRV stops revalidating but does not necessarily clear previous `data`. With `displayData` falling back to existing data, the tile can keep showing stale results while no current query is allowed to run. 

Should the old-data/current-data display path be disabled when `queryReady` is false?
I think I’d expect the skeleton because the current query inputs are not ready.

| State | Expected UI |
| --- | --- |
| `queryReady === false` | Show skeleton/loading |
| `queryReady === true` + replacement query pending + `oldData` exists | Show `oldData` |
| `queryReady === true` + current query succeeded | Show current data |
| `queryReady === true` + current query settled with error | Show error state |


One possible shape would be to make `displayData` fall back to `oldData` only when `props.queryReady && state.value === STATE.PENDING`.

filipgutica (changes requested):

packages/analytics/dashboard-renderer/src/components/QueryDataProvider.vue
jordojordo: I don't think this state you mention is possible in the way we're setting `queryReady` currently, which is done solely by `configStore.loading` which is set when the store loads, we wouldn't have any data prior to the store loading so the case of old data being displayed and swrv not being able to make a query is unreachable.

Although, I think adding in that guard is fine incase the `useDashboardInternalContext` composable is updated which _could_ cause things to change.

filipgutica (approved):

PR #3547, opened 2026-07-08
fix: allow platform and platform_usage for datasource [MA-5024]

shifan-w:
# Summary

MA-5024-dashboard-add-data-source
Add an additional allowed data source to platform analytics dashboard query.  `platform_usage` will act the same as `platform`

### DIFF
--- a/packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
+++ b/packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
@@ -610,13 +610,18 @@ export const agenticUsageSchema = {
 
 export const platformQuerySchema = {
   type: 'object',
-  description: 'A query to launch at the platform dashboard API',
+  description: "A query to launch at the platform dashboard API. Use datasource 'platform_usage'; 'platform' is accepted for backward compatibility but deprecated.",
   properties: {
     datasource: {
-      type: 'string',
-      enum: [
-        'platform',
-        'platform_usage',
+      oneOf: [
+        {
+          const: 'platform',
+          deprecated: true,
+          description: "Deprecated: use 'platform_usage'.",
+        } as unknown as JSONSchema,
+        {
+          const: 'platform_usage',
+        },
       ],
     },
     metrics: metricsFn(),
@@ -630,13 +635,18 @@ export const platformQuerySchema = {
 
 export const platformTabularQuerySchema = {
   type: 'object',
-  description: 'A query to launch at the platform tabular explore API',
+  description: "A query to launch at the platform tabular explore API. Use datasource 'platform_usage'; 'platform' is accepted for backward compatibility but deprecated.",
   properties: {
     datasource: {
-      type: 'string',
-      enum: [
-        'platform',
-        'platform_usage',
+      oneOf: [
+        {
+          const: 'platform',
+          deprecated: true,
+          description: "Deprecated: use 'platform_usage'.",
+        } as unknown as JSONSchema,
+        {
+          const: 'platform_usage',
+        },
       ],
     },
     entity: {
--- a/packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
+++ b/packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
@@ -616,6 +616,7 @@ export const platformQuerySchema = {
       type: 'string',
       enum: [
         'platform',
+        'platform_usage',
       ],
     },
     metrics: metricsFn(),
@@ -635,6 +636,7 @@ export const platformTabularQuerySchema = {
       type: 'string',
       enum: [
         'platform',
+        'platform_usage',
       ],
     },
     entity: {
--- a/packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
+++ b/packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
@@ -608,22 +608,24 @@ export const agenticUsageSchema = {
   additionalProperties: false,
 } as const satisfies JSONSchema
 
+const platformDatasourceSchema = {
+  oneOf: [
+    {
+      const: 'platform',
+      deprecated: true,
+      description: "Deprecated: use 'platform_usage'.",
+    },
+    {
+      const: 'platform_usage',
+    },
+  ],
+} as const
+
 export const platformQuerySchema = {
   type: 'object',
   description: "A query to launch at the platform dashboard API. Use datasource 'platform_usage'; 'platform' is accepted for backward compatibility but deprecated.",
   properties: {
-    datasource: {
-      oneOf: [
-        {
-          const: 'platform',
-          deprecated: true,
-          description: "Deprecated: use 'platform_usage'.",
-        } as unknown as JSONSchema,
-        {
-          const: 'platform_usage',
-        },
-      ],
-    },
+    datasource: platformDatasourceSchema,
     metrics: metricsFn(),
     dimensions: dimensionsFn(),
     filters: platformFiltersFn(),
@@ -637,18 +639,7 @@ export const platformTabularQuerySchema = {
   type: 'object',
   description: "A query to launch at the platform tabular explore API. Use datasource 'platform_usage'; 'platform' is accepted for backward compatibility but deprecated.",
   properties: {
-    datasource: {
-      oneOf: [
-        {
-          const: 'platform',
-          deprecated: true,
-          description: "Deprecated: use 'platform_usage'.",
-        } as unknown as JSONSchema,
-        {
-          const: 'platform_usage',
-        },
-      ],
-    },
+    datasource: platformDatasourceSchema,
     entity: {
       type: 'string',
     },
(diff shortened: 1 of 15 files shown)


packages/analytics/dashboard-renderer/src/utils/table-data-grid-renderer.ts
```diff
@@ -42,108 +42,112 @@ const platformTabularResponseToTableDataGridRows = (
...
-    canTranslate,
-    columns,
-    translate,
-  }: {
-    canTranslate: (key: string) => boolean
-    columns: string[]
-    translate: (key: string) => string
-  }): Array<TableDataGridHeader<TableDataGridRow>> => columns.map(column => {
-    const labelKey = `chartLabels.${column}`
+const platformTableDataGridHeaders = ({
```
shifan-w, line 54: This changed from `platform` to `platformTableDataGridHeaders`

packages/analytics/dashboard-renderer/src/utils/table-data-grid-renderer.ts
```diff
@@ -42,108 +42,112 @@ const platformTabularResponseToTableDataGridRows = (
...
+  return {
+    key: column,
+    label: canTranslate(labelKey) ? translate(labelKey) : column,
+  }
+})
+
+/** Maps a table tile datasource to the header builder used by TableDataGridRenderer. */
+export const tableDataGridHeadersByDatasource = {
+  platform: platformTableDataGridHeaders,
+  platform_usage: platformTableDataGridHeaders,
```
shifan-w, line 75: Now `platformTableDataGridHeaders` supports both `platform` and `platform_usage`.

packages/analytics/analytics-utilities/src/types/query-bridge.ts
```diff
@@ -40,7 +40,7 @@ export interface PlatformDatasourceQuery {
 export type DatasourceAwareQuery = BasicDatasourceQuery | AdvancedDatasourceQuery | AiDatasourceQuery | AgenticDatasourceQuery | PlatformDatasourceQuery
 
 export interface PlatformDatasourceTabularQuery {
-  datasource: 'platform'
+  datasource: 'platform' | 'platform_usage'
```
shifan-w: I wonder what if we remove `platform` here.  Just leave `platform_usage`.

Will this be a breaking change that breaks what's already released to our customer?

packages/analytics/analytics-utilities/src/types/query-bridge.ts
filipgutica, line 37: Missed here 'platform' | 'platform_usage'

filipgutica (changes requested):

packages/analytics/analytics-utilities/src/types/query-bridge.ts
shifan-w: We can't remove because we released with `platform` for m1.

packages/analytics/analytics-utilities/src/types/query-bridge.ts
```diff
@@ -33,14 +33,15 @@ export interface AgenticDatasourceQuery {
 }
 
 export interface PlatformDatasourceQuery {
-  datasource: 'platform'
+  /** @deprecated Use `platform_usage`. `platform` will be removed in a future version. */
+  datasource: 'platform' | 'platform_usage'
```
filipgutica: I think this will mark `datasource` as deprecated, so I don't think we need the comment here

packages/analytics/analytics-utilities/src/types/explore/all.ts
```diff
@@ -55,6 +58,7 @@ export const datasourceToFilterableDimensions: Record<QueryDatasource, Set<strin
   llm_usage: new Set(filterableAiExploreDimensions),
   agentic_usage: new Set(filterableAgenticExploreDimensions),
   platform: new Set(),
```
filipgutica, line 61: /** @deprecated  **/ to platform?

packages/analytics/dashboard-renderer/src/utils/table-data-grid-renderer.ts
```diff
@@ -42,108 +42,112 @@ const platformTabularResponseToTableDataGridRows = (
...
-  }),
+  return {
+    key: column,
+    label: canTranslate(labelKey) ? translate(labelKey) : column,
+  }
+})
+
+/** Maps a table tile datasource to the header builder used by TableDataGridRenderer. */
+export const tableDataGridHeadersByDatasource = {
+  platform: platformTableDataGridHeaders,
```
filipgutica, line 74: mark platform deprecated

packages/analytics/dashboard-renderer/src/utils/table-data-grid-renderer.ts
```diff
@@ -42,108 +42,112 @@ const platformTabularResponseToTableDataGridRows = (
...
+    cursor: response.meta.cursor,
+    hasMore: Boolean(response.meta.cursor),
+  }
+}
+
+/**
+ * Maps a table tile datasource to the fetcher used by TableDataGridRenderer.
+ */
+export const tableDataGridFetcherByDatasource = {
+  platform: platformTableDataGridFetcher,
```
filipgutica, line 153: mark platform deprecated

packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
```diff
@@ -635,6 +636,7 @@ export const platformTabularQuerySchema = {
       type: 'string',
       enum: [
         'platform',
+        'platform_usage',
       ],
```
filipgutica: Lets mark "platform" as deprecated. With the current shared enum, there isn’t a place to attach deprecation metadata to just platform; adding deprecated: true on datasource would make platform_usage look deprecated too.

One option is to model the datasource values as separate branches:

```
datasource: {
  oneOf: [
    {
      const: 'platform',
      deprecated: true,
      description: 'Deprecated: use `platform_usage`.',
    },
    {
      const: 'platform_usage',
    },
  ],
},
```

packages/analytics/dashboard-renderer/src/composables/useContextLinks.ts
```diff
@@ -68,7 +68,7 @@ export default function useContextLinks(
     return true
   })
 
-  const canGenerateRequestsLink = computed(() => requestsBaseUrl.value && definition.value.query && !isTableChart.value && definition.value.query.datasource !== 'llm_usage' && definition.value.query.datasource !== 'platform' && isAdvancedAnalytics.value && !datasourceConfigLoading.value)
+  const canGenerateRequestsLink = computed(() => requestsBaseUrl.value && definition.value.query && !isTableChart.value && definition.value.query.datasource !== 'llm_usage' && definition.value.query.datasource !== 'platform' && definition.value.query.datasource !== 'platform_usage' && isAdvancedAnalytics.value && !datasourceConfigLoading.value)
```
filipgutica: This is getting a bit long, may be add

`const nonRequestsLinkDatasources = ['llm_usage', 'platform', 'platform_usage']`

Then call `&& !nonRequestsLinkDatasources.includes(definition.value.query.datasource)`

filipgutica (changes requested):

packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
shifan-w: About marking `platform` as deprecated, since we don't ever publicly document this field and marking a *value* of an enum is challenging.  
Maybe we don't need to annotate the deprecation, because it never existed in the documentation in the first place.  

What do you think?  I will give the deprecation annotation one more try and if it doesn't do what we want, maybe we don't mark it as deprecated.

packages/analytics/analytics-utilities/src/types/explore/platform.ts
```diff
@@ -37,7 +43,8 @@ export interface PlatformTabularResponseMeta {
   page_size: number
   display: DisplayBlob
   cursor?: string
-  datasource?: 'platform'
+  /** @deprecated Will be `platform_usage` in future responses. */
```
filipgutica: Remove the comment here cause it will make "datasource" deprecated

packages/analytics/analytics-utilities/src/types/query-bridge.ts
```diff
@@ -33,14 +33,14 @@ export interface AgenticDatasourceQuery {
 }
 
 export interface PlatformDatasourceQuery {
-  datasource: 'platform'
+  datasource: 'platform' | 'platform_usage'
   query: PlatformExploreQuery
 }
```
filipgutica, line 39: I think we should mark the entire interface as deprecated and make a new one called 

What do you think? 

Same thing in the other spots where we are setting `datasource: 'platform' | 'platform_usage'` 

```
PlatformUsageDatasourceQuery { 
  ...
  datasource: 'platform_usage'
}
```

filipgutica (changes requested):
One more comment

filipgutica (approved):
One more non-blocking comment. I'll leav

PR #3556, opened 2026-07-13
Ma 5024 dashboard add data source

shifan-w:
# Summary

MA-5210 (related ticket to MA-5024).
https://konghq.atlassian.net/browse/MA-5210
- remove the hard code option `platform` in favor of query.datasource
- reorder the enum check and test check order.
- Update unit tests

## Additional test plan
Import this version in `Konnect-UI-Apps` and test on local UI.

### DIFF
--- a/packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
+++ b/packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
@@ -611,12 +611,12 @@ export const agenticUsageSchema = {
 const platformDatasourceSchema = {
   oneOf: [
     {
+      const: 'platform_usage',
+    },
+    {
       const: 'platform',
       deprecated: true,
       description: "Deprecated: use 'platform_usage'.",
-    },
-    {
-      const: 'platform_usage',
     },
   ],
 } as const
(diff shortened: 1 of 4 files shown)


packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
```diff
@@ -610,14 +610,14 @@ export const agenticUsageSchema = {
 
 const platformDatasourceSchema = {
   oneOf: [
+    {
+      const: 'platform_usage',
+    },
```
Chalks, line 615: **question, maybe blocker**

why are you adding this and also removing it? Did you intend to add `const: 'platform',` and remove `platform_usage`?

Chalks: or does order matter here?

Chalks: after reading your PR description (which I should have done first) I see you're doing it for order... why does order matter?

shifan-w: There was a test using `anyOf` to check order of enum in `kAnalytics`.  There was a non-blocking comment in my previous PR wishing to have it resolved, so I resolved it here.
background:
https://github.com/kong-konnect/kanalytics/pull/1503#discussion_r3553720227

Chalks (approved):

PR #3578, opened 2026-07-21
fix(entities-plugins): stop fillRecord from re-toggling an already-set switch

2eha0:
## Summary
- `fillRecord` (playwright filler) unconditionally `.click()`'d an optional record field's enable/disable switch, with no check of its current state.
- Re-filling a record that's already enabled (e.g. updating a plugin whose existing config already populated that field) incorrectly **toggles the switch off**, collapsing the object's `SlideTransition` content (`v-if="expanded"`) and detaching any child element the filler is mid-interaction with.
- Fixed by using `.check()`/`.uncheck()` instead of `.click()` - both are no-ops if the element is already in the target state, matching what the cypress `fillRecord` handler already does correctly.
- Added a regression test to both fillers: fill an optional record twice (simulating create then update) and assert the switch stays checked and children remain fillable.

## Root cause
Traced from a Kong Manager test failure: `TimeoutError ... element was detached from the DOM, retrying` while updating a nested `authentication` record (`mode`/`basic`/`oauth2`, discriminated by `mode`) that already had data from a prior create step. The record's switch was already checked; the second `fill()` call's `fillRecord` step for `authentication` unconditionally clicked the switch, unchecking it and collapsing everything inside - including the `mode` field the filler was about to click next.

## Test plan
- [x] New playwright test **fails** against the pre-fix handler (`git stash` the fix) with `Received: unchecked` on the second fill - confirms it reproduces the exact bug
- [x] New playwright test **passes** with the fix - `npx playwright test fill-form.pw.ts`: 35/35 passed
- [x] New cypress test **passes** as-is (cypress's handler already uses `.check()`/`.uncheck()`) - `cypress run --component --spec fill-form.cy.ts`: 38/38 passed
- [x] `eslint` clean on all changed files

### DIFF
--- a/packages/entities/entities-plugins/src/components/free-form/filler/cypress/fill-form.cy.ts
+++ b/packages/entities/entities-plugins/src/components/free-form/filler/cypress/fill-form.cy.ts
@@ -206,6 +206,44 @@ describe('Filler - Cypress', () => {
       cy.get('[data-testid="ff-config.port"]').should('have.value', '6379')
     })
 
+    it('record: re-filling an already-enabled optional object does not collapse it', () => {
+      // Companion to the playwright regression test: cypress's fillRecord already
+      // uses .check()/.uncheck() (idempotent - no-op if already in that state),
+      // so it doesn't have the playwright handler's bug (a plain .click() that
+      // re-toggles, and thus collapses, an already-enabled object). Kept here as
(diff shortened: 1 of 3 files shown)


2eha0:
Merged into #3577 instead — both filler fixes (foreign fields + record switch re-toggle) are combined there.

PR #3598, opened 2026-07-22
fix(table-data-grid): show tooltips for truncated cells

filipgutica:
## Summary

Closes [MA-5198](https://konghq.atlassian.net/browse/MA-5198).

Truncates overflowing TableDataGrid cell content and shows the full value in a KTooltip. The tooltip is teleported to the document body so AG Grid cells do not clip it during normal rendering or row virtualization.

The sandbox now includes a long-content column and lets the table fill the available width.

Preview PR: https://github.com/kong-konnect/konnect-ui-apps/pull/12805

image

[MA-5198]: https://konghq.atlassian.net/browse/MA-5198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/table-data-grid/sandbox/App.vue
+++ b/packages/core/table-data-grid/sandbox/App.vue
@@ -237,6 +237,7 @@ import { computed, ref } from 'vue'
 import { TableDataGrid } from '../src'
 
 type SandboxRow = {
+  description: string
   id: string
   name: string
   status: string
@@ -295,6 +296,7 @@ const collapsedSections = ref<Record<SandboxSectionId, boolean>>({
 
 const headers: Array<TableDataGridHeader<SandboxRow>> = [
   { key: 'name', label: 'Name', minWidth: 220 },
+  { key: 'description', label: 'Description', width: 240 },
   { key: 'status', label: 'Status', minWidth: 140 },
   { key: 'latency', label: 'Latency', minWidth: 140 },
   { key: 'region', label: 'Region', minWidth: 140 },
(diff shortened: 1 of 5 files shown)


packages/core/table-data-grid/src/components/TableDataGridCellRenderer.vue
```diff
@@ -0,0 +1,125 @@
+<template>
+  <span class="table-data-grid-cell-renderer">
+    <KTooltip
+      class="table-data-grid-cell-tooltip"
+      :disabled="!isOverflowing"
+      max-width="300"
```
Chalks, line 7: **suggestion**

This is maybe(?) an unusual circumstance, but when you've got a bunch of long tooltips and scroll, it looks a little jank:

https://github.com/user-attachments/assets/165cd711-5e58-433f-9373-f6a58e164745

```suggestion
      :kpop-attributes="{ popoverDelay: 400 }"
      max-width="300"
```

If we add that delay, it does this instead:

https://github.com/user-attachments/assets/b7e62c7a-432d-44bf-86b2-206295cd5183

packages/core/table-data-grid/sandbox/App.vue
```diff
@@ -295,6 +296,7 @@ const collapsedSections = ref<Record<SandboxSectionId, boolean>>({
 
 const headers: Array<TableDataGridHeader<SandboxRow>> = [
   { key: 'name', label: 'Name', minWidth: 220 },
+  { key: 'description', label: 'Description', width: 240 },
```
Chalks, line 299: **question, maybe a blocker?**

Why is this one `width: 240` while all the other headers are `minWidth`? Probably doesn't matter much, but it's a little unclear why it's this way.

filipgutica: It was just to fix the width to ensure overflow. It's pretty arbitrary... just sandbox code so didn't think too much about it

Chalks (dismissed):

Chalks (approved):

PR #3601, opened 2026-07-23
fix(page-layout): truncate title at first line

portikM:
# Summary

Addresses: https://konghq.atlassian.net/browse/KHCP-21147

Truncate PageLayout title at first line

Before

Screenshot 2026-07-23 at 3 08 18 PM

After

Screenshot 2026-07-23 at 3 08 47 PM

### DIFF
--- a/packages/core/page-layout/src/components/PageLayout.vue
+++ b/packages/core/page-layout/src/components/PageLayout.vue
@@ -301,11 +301,16 @@ watch([() => pageShortcutData, () => route?.fullPath], () => {
 
           .page-layout-title-wrapper {
             > * {
+              -webkit-box-orient: vertical;
               color: var(--kui-color-text, $kui-color-text);
+              display: -webkit-box;
               font-size: var(--kui-font-size-50, $kui-font-size-50);
               font-weight: var(--kui-font-weight-semibold, $kui-font-weight-semibold);
+              -webkit-line-clamp: 1;
+              line-clamp: 1;
               line-height: var(--kui-line-height-40, $kui-line-height-40);
               margin: var(--kui-space-0, $kui-space-0);
+              overflow: hidden;
             }
           }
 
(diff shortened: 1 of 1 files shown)


adamdehaven (changes requested):
CSS question

packages/core/page-layout/src/components/PageLayout.vue
```diff
@@ -301,11 +301,16 @@ watch([() => pageShortcutData, () => route?.fullPath], () => {
 
           .page-layout-title-wrapper {
             > * {
+              -webkit-box-orient: vertical;
```
adamdehaven: `box-orient` is deprecated 🤔

portikM: Without it line clamping doesn't work

adamdehaven: The property is deprecated in most browsers -- how does it work in non-webkit browsers?

portikM: Yes -- checked in Chrome, Safari & Firefox

adamdehaven: If we're always truncating at one line, why can't we just do this:

```scss
white-space: nowrap;
overflow: hidden;
text-overflow: ellipsis;
```

You only need what you have above if supporting multiple lines -- if only 1, this suggestion is more compatible

portikM: Ellipsis overflow results in horizontal scrollbar. I could look into fixing that but it will require some style changes on other wrapping elements

Image

adamdehaven (dismissed):

packages/core/page-layout/src/components/PageLayout.vue
portikM: Fixed, now using ellipsis for truncation instead of line clamp

adamdehaven (approved):

jillztom (approved):

ValeryG (approved):

PR #3612, opened 2026-07-27
feat(redis): cloud auth supports oauth

Leopoldthecoder:
# Summary

A new option "OAuth" is added:
<p align="center">
image</p>

Reveal relevant fields when "OAuth" is selected:
<p align="center">
image</p>

KM-2969

### DIFF
--- a/packages/entities/entities-redis-configurations/src/components/RedisConfigurationConfigCard.vue
+++ b/packages/entities/entities-redis-configurations/src/components/RedisConfigurationConfigCard.vue
@@ -570,6 +570,17 @@ const recordResolver = (payload: Record<string, any>): AddOnRecord => {
       true,
     ) : null
 
+  // The flat display drops nested objects; keep the whole `oauth` record so it renders as a JSON code block
+  if (
+    cloudAuthForDisplay &&
+    rawCloudAuth !== null &&
+    typeof rawCloudAuth === 'object' &&
+    !Array.isArray(rawCloudAuth) &&
+    rawCloudAuth.oauth
+  ) {
+    cloudAuthForDisplay.oauth = rawCloudAuth.oauth as AddOnValue
+  }
+
   return {
     id: row.id,
     name: row.name,
--- a/packages/entities/entities-redis-configurations/src/components/RedisConfigurationForm.vue
+++ b/packages/entities/entities-redis-configurations/src/components/RedisConfigurationForm.vue
@@ -117,6 +117,12 @@
             :config="props.config"
             :readonly="form.readonly"
           />
+          <OauthFields
+            v-if="form.fields.config.cloud_authentication!.auth_provider === AuthProvider.OAUTH"
+            v-model="form.fields.config.cloud_authentication!.oauth!"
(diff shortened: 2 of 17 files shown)


packages/entities/entities-redis-configurations/src/components/RedisConfigurationForm.vue
```diff
@@ -684,6 +691,15 @@ const cloudAuthOptions = computed<SelectItem[]>(() => {
       value: AuthProvider.AZURE,
     },
   ]
+
+  if (props.config.oauthCloudAuthAvailable) {
```
TT1228, line 695: I was thinking(non-blocking): should we query redis-partial schema and decide whether to show `oauth` option by reading schema? This is easier to maintain.

Leopoldthecoder: I like this idea. If we read from the schema, all the following props can be removed:
https://github.com/Kong/public-ui-components/blob/d171927d1e464588dea27903e0cb3fe12ba540a8/packages/entities/entities-redis-configurations/src/types/redis-configuration-form.ts#L19-L42

We can do this later and release a new major version for `@kong-ui-public/entities-redis-configurations`. Ticket created: https://konghq.atlassian.net/browse/KM-2982

packages/entities/entities-redis-configurations/src/components/RedisConfigurationConfigCard.vue
```diff
@@ -570,6 +570,17 @@ const recordResolver = (payload: Record<string, any>): AddOnRecord => {
       true,
     ) : null
 
+  // The flat display drops nested objects; keep the whole `oauth` record so it renders as a JSON code block
```
aanchalm01, line 573: This **recordResolver** change looks fine for self-managed/legacy partial detail. The managed cache tab goes through _addOnRecordResolver_ >>> _managed-cache-add-on-display_, so it won’t use this path. No concerns for managed redis flows from this change.

packages/entities/entities-redis-configurations/src/helpers.ts
```diff
@@ -160,6 +160,11 @@ export const pickCloudAuthFields =
           azure_client_secret: cloudAuth.azure_client_secret,
           azure_tenant_id: cloudAuth.azure_tenant_id,
         }
+      case AuthProvider.OAUTH:
```
aanchalm01, line 163: **pickCloudAuthFields** is also used by _managed-cache-add-on-display_, but for now _{ auth_provider }_ is a no-op for _Konnect managed add-ons_, in future if _OAuth_ ever shows up on _add-on payloads_, we’d likely need parallel display handling there as well.

Leopoldthecoder: > in future if OAuth ever shows up on add-on payloads

Thanks. If that ever happens, please let us know if there's anything you need from us.

2eha0 (approved):

kingyue737 (approved):

PR #3620, opened 2026-07-28
feat(expressions): migrate editor to @kong-ui-public/monaco-editor

kingyue737:
[KM-2202]

## Problem

`@kong-ui-public/expressions` shipped its own Monaco integration, including a **custom Monaco theme** (`kong-expr-theme`) for syntax highlighting and a local `MonacoEditor.vue` wrapper. This PR moves it onto the shared `@kong-ui-public/monaco-editor`.

The shared editor takes Monaco over with Shiki and a single global theme. That breaks the old approach in a specific way: Shiki overrides `monaco.editor.setTheme` and **throws on any theme name it doesn't recognize**, so activating a custom theme is no longer possible. The challenge is keeping both syntax highlighting and completion working under this Shiki-driven setup.

### Syntax highlighting & completion

Split the two concerns the old editor fused into one tokenizer:

- **Highlighting** — a Monarch tokens provider whose token types are *TextMate-scope names* (`keyword.operator`, `variable.other.property`, `entity.name.function`, `string.quoted.double`, …). The shared Shiki theme already has rules for those scopes, so tokens are coloured automatically and consistently with the rest of the app. **No custom Monaco theme is defined or activated**, which sidesteps the `setTheme`-throws problem entirely.
- **Completion analysis** — driven by the standard `monaco.editor.tokenize()`. The `TokenType` constants hold the scope names as their values, so a single vocabulary serves both concerns; `shortenTokenType()` strips the `.<languageId>` suffix to recover the token type the analysis utilities expect.

### Component integration

Both editors now render the shared `<MonacoEditor>` component with `appearance="standalone"`, so they inherit the design-system container styling (border, hover, focus) instead of re-implementing it:

## Changes

- `src/monaco.ts`: Monarch provider emitting scope-named tokens; `TokenType` values are TextMate scopes; `shortenTokenType()` recovers them for analysis; custom theme removed.
- `src/components/ExpressionsEditor.vue`: render the shared `<MonacoEditor>` component (standalone) + local `invalid` styling; completion runs off `monaco.editor.tokenize()`.
- `src/components/RequestImportModal.vue`: use the shared `<MonacoEditor>`; drop the local wrapper.
- `src/components/MonacoEditor.vue`: **removed** (superseded by the shared package).
- `vite.config.ts`: switch to `@kong-ui-public/monaco-editor/vite-plugin`. The worker setup only takes effect in the sandbox build; in the library build `monaco-editor` (and `@kong-ui-public/monaco-editor`) stay external, so no Monaco or worker chunks are bundled.

[KM-2202]: https://konghq.atlassian.net/browse/KM-2202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/expressions/src/components/ExpressionsEditor.vue
```diff
@@ -342,21 +340,21 @@ watch(() => parseResult.value, (result?: ParseResult) => {
...
-  border-radius: 3px;
-  min-height: 200px;
-  overflow: hidden;
-  transition: border-color linear 150ms;
-  width: 100%;
+
+  :deep(.monaco-editor-target) {
+    min-height: 200px;
+    // flex: 1;
+    // min-height: 0;
```
sumimakito, line 353: nit: Do we need to keep them?

sumimakito (commented):
Manually tested and LGTM

sumimakito (approved):

sumimakito:
Just a small nit: the highlight color looks a bit dark on some tokens in the dark theme:

Screenshot 2026-07-30 at 16 55 38

kingyue737:
> Just a small nit: the highlight color looks a bit dark on some tokens in the dark theme:

Good catch. Currently, all Monaco editors in `public-ui-components` use a light theme even in dark mode, so they're rendered with a light theme on a dark background. We need to fix that consistently. https://kongstrong.slack.com/archives/C0A0RQLADPG/p1785381829258449

packages/core/expressions/src/components/ExpressionsEditor.vue
kingyue737: Good catch! I'll remove it in the next PR

Leopoldthecoder (approved):

PR #3622, opened 2026-07-28
feat(entities-shared): deep sensitive fields [khcp-20573]

kaiarrowood:
# Summary

Fix redaction of deeply nested sensitive fields in `record` and `array` schema types.
Part of [KHCP-20573](https://konghq.atlassian.net/browse/KHCP-20573).
Also adds tooltip for [KHCP-20982](https://konghq.atlassian.net/browse/KHCP-20982).

[KHCP-20573]: https://konghq.atlassian.net/browse/KHCP-20573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[KHCP-20982]: https://konghq.atlassian.net/browse/KHCP-20982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/entities/entities-shared/src/components/entity-base-config-card/EntityBaseConfigCard.vue
+++ b/packages/entities/entities-shared/src/components/entity-base-config-card/EntityBaseConfigCard.vue
@@ -479,16 +479,18 @@ const { redactByConfigSchema, redactByApiSchema, isObjectRecord, getApiSchemaFie
 
 const redactedCodeBlockRecord = computed((): Record<string, any> => {
   const source = codeBlockRecordFromApi.value || record.value
-  let rec = redactByConfigSchema(source, props.configSchema) as Record<string, any>
+  let rec = source
 
   if (isObjectRecord(source) && Array.isArray(schema?.value?.fields)) {
     const schemaRedactedRecord: Record<string, any> = {}
     for (const key in source) {
       const fieldSchema = getApiSchemaField(schema?.value?.fields, key)
-      schemaRedactedRecord[key] = fieldSchema ? redactByApiSchema(source[key], fieldSchema) : rec[key]
+      schemaRedactedRecord[key] = fieldSchema ? redactByApiSchema(source[key], fieldSchema) : source[key]
     }
     rec = schemaRedactedRecord
   }
+
+  rec = redactByConfigSchema(rec, props.configSchema) as Record<string, any>
 
   return props.codeBlockRecordResolver ? props.codeBlockRecordResolver(rec) : rec
 })
--- a/packages/entities/entities-shared/src/components/entity-base-config-card/EntityBaseConfigCard.vue
+++ b/packages/entities/entities-shared/src/components/entity-base-config-card/EntityBaseConfigCard.vue
@@ -27,6 +27,13 @@
           class="sensitive-fields-checkbox"
           data-testid="sensitive-fields-checkbox"
           :label="t('baseConfigCard.actions.sensitive_fields')"
+          :label-attributes="{
+            info: t('baseConfigCard.actions.sensitive_fields_tooltip'),
+            tooltipAttributes: {
+              maxWidth: '200',
+              placement: 'top',
+            },
+          }"
         />
 
         <div class="row">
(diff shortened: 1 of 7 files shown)


packages/entities/entities-shared/src/composables/useHelpers.ts
```diff
@@ -93,10 +95,124 @@ export default function useHelpers() {
...
+
+      if (!elementSchema) {
+        return value.slice()
+      }
+
+      // recursively redact array values
+      return value.map((item: unknown) => redactByApiSchema(item, {
+        ...elementSchema,
+        // pass down encryption flag since array children inherit from parent
+        encrypted: fieldSchema.encrypted,
```
kingyue737: I think this should preserve `elements.encrypted` and only force encryption when the parent array is encrypted.

Right now `encrypted: fieldSchema.encrypted` overwrites an element schema like `{ type: 'string', encrypted: true }` with `undefined` when the parent array itself does not have `encrypted: true`. We have Gateway schemas with that shape, e.g. `ai-mcp-proxy` has `server.session.client.secrets` as an array whose `elements.encrypted` is `true`.

Ref: https://github.com/Kong/gateway-schema-watcher/blob/86e51153f23d5125c94b14cd61fd7e45692d31f8/api-gateway/plugins/ai-mcp-proxy.json#L173

Image

kingyue737 (commented):
Thanks! AFAICT there's no `map` or `set` field currently marked `encrypted: true` in the plugin schemas. Do we also need to handle them, @2eha0 ?

packages/entities/entities-shared/src/components/entity-base-config-card/EntityBaseConfigCard.vue
```diff
@@ -466,13 +474,20 @@ const codeBlockRecordFromApi = computed((): Record<string, any> | undefined => {
...
+
+const redactedCodeBlockRecord = computed((): Record<string, any> => {
+  const source = codeBlockRecordFromApi.value || record.value
+  let rec = redactByConfigSchema(source, props.configSchema) as Record<string, any>
+
+  if (isObjectRecord(source) && Array.isArray(schema?.value?.fields)) {
+    const schemaRedactedRecord: Record<string, any> = {}
+    for (const key in source) {
+      const fieldSchema = getApiSchemaField(schema?.value?.fields, key)
+      schemaRedactedRecord[key] = fieldSchema ? redactByApiSchema(source[key], fieldSchema) : rec[key]
```
aanchalm01: Is API schema the source of truth here? If a field is marked Redacted/RedactedArray in _configSchema_ but the API field isn’t _encrypted: true_, it’ll show unmasked in the code tabs.

aanchalm01:
Functionally tested, looks good. Sensitive fields show up as redacted by default in the config card for all 4 formats and becomes visible when checking the Show sensitive fields checkbox.

**`Session`**

**Before**

screen capture 2026-07-29 at 2 14 41 PM

**After**

screen capture 2026-07-29 at 2 14 23 PM

`**ACME**`

**Before**

screen capture 2026-07-29 at 2 09 31 PM

**After**

screen capture 2026-07-29 at 2 05 51 PM

**`OpenID Connect`**

**Before**

screen capture 2026-07-29 at 1 51 04 PM

**After**

screen capture 2026-07-29 at 1 50 30 PM

packages/entities/entities-shared/src/components/entity-base-config-card/EntityBaseConfigCard.vue
kaiarrowood: Good catch, fixed

packages/entities/entities-shared/src/composables/useHelpers.ts
kaiarrowood: Applied the fix and added support for `set` and `map` type.

packages/entities/entities-shared/src/composables/useHelpers.ts
```diff
@@ -93,10 +95,137 @@ export default function useHelpers() {
...
+      return fieldSchema.encrypted ? REDACTED_MASK : value
+    }
+
+    if (fieldSchema.type === 'record' || fieldSchema.type === 'map') {
+      if (!isObjectRecord(value)) {
+        return value
+      }
+
+      if (fieldSchema.type === 'map' && !fieldSchema.values) {
+        // Do I need this? could I just fall through to record handling?
```
kingyue737: non-blocking nit: shall we remove this comment?

kingyue737 (dismissed):
LGTM🚀

aanchalm01 (approved):
LGTM!

PR #3631, opened 2026-07-30
feat: add managed cache usage MA-5168

alxgrant:
# Summary


packages/analytics/analytics-utilities/src/dashboardSchema.v2.ts
```diff
@@ -828,6 +851,7 @@ export const dashboardConfigSchema = {
         ...filterableAiExploreDimensions,
         ...filterableAgenticExploreDimensions,
         ...filterablePlatformPresetFilterDimensions,
+        ...filterableManagedCacheExploreDimensions,
```
alxgrant, line 854: Does this expose the new managed cache fields in UI? or is this just for validation? @filipgutica

filipgutica: Just for validation

PR #3638, opened 2026-08-03
fix(entities-consumer-groups): remove "Please"

zz505000:
remove "Please"

UI copy fixes applied via Kontent:
- `consumer_groups.errors.add_consumer`


zz505000:
testing

PR #3646, opened 2026-08-04
feat(entities-gateway-services): convert client & CA certificates to select

TT1228:
KM-3008

Convert client & CA certificates to select and guard them with feature flag.

https://github.com/user-attachments/assets/c8fa60a3-c886-4bb8-96bd-4af05dc892df


packages/entities/entities-gateway-services/package.json
```diff
@@ -36,6 +39,7 @@
     "@kong/icons": "^1.62.1",
     "@kong/kongponents": "9.64.3",
     "axios": "^1.17.0",
+    "reflect-metadata": "^0.2.2",
```
kingyue737: I think `reflect-metadata` should be added to `dependencies` since it’s used by the runtime code. If possible, I think it would be better to externalize this package.

packages/entities/entities-gateway-services/src/components/GatewayServiceForm.vue
```diff
@@ -558,6 +671,123 @@ const validators = useValidators()
...
+  debouncedQueryChange: debouncedCertificateCall,
+  loading: certificateLoading,
+  loadItems: loadCertificates,
+  results: certificateResults,
+} = useDebouncedFilter(props.config, endpoints.form[props.config.app].getCertificates)
+
+// Return the tags for a fetched certificate by its ID (empty if not loaded)
+const getCertificateTags = (id: string | number): string[] =>
+  certificateResults.value?.find((cert) => cert.id === id)?.tags ?? []
+
```
kingyue737, line 689: `getCertificateTags/getCaCertificateTags` do a linear `Array.find` scan per call from a render-time slot. Could we build a Map keyed by ID to avoid the O(m × n) complexity here?

packages/entities/entities-gateway-services/src/components/GatewayServiceForm.vue
```diff
@@ -558,6 +671,123 @@ const validators = useValidators()
...
+    caCertificateIssuers.value[id] = new X509Certificate(cert.cert).issuer
+  } catch {
+    caCertificateIssuers.value[id] = ''
+  }
+}
+
+// Resolve issuers whenever the fetched CA certificates change
+watch(caCertificateResults, (results) => {
+  (results ?? []).forEach((cert) => resolveCaCertificateIssuer(cert))
+}, { immediate: true })
```
kingyue737, line 774: `resolveCaCertificateIssuer` is currently invoked once per certificate in `caCertificateResults`, and each call writes to `caCertificateIssuers.value[id]` individually — with a large certificate list this triggers a large number of separate reactive updates. Consider having it accept the whole array instead, resolve all certificates concurrently internally, and merge the results into `caCertificateIssuers.value` in a single write.

kingyue737 (approved):
LGTM

Leopoldthecoder (approved):

PR #3647, opened 2026-08-04
feat(entities-plugins): rank and highlight plugin search matches

TT1228:
# Summary

KM-3029

Rank search results by match strength (prefix > word-initial acronym > consecutive substring > scattered subsequence) and highlight the matched characters in both the catalog grid card and list view.


packages/entities/entities-plugins/src/utils/helper.ts
Leopoldthecoder, line 1: Can we add some unit tests for these helper functions?

Leopoldthecoder (approved):

PR #3666, opened 2026-08-06
feat(entities-plugins): add allow_when/deny_when modes to ACL plugin

TT1228:
# Summary

Extend ACLModeCard to support the ACL plugin's newer allow_when/deny_when CEL-expression fields alongside allow/deny, gated on whether the active schema actually declares them (older Gateway versions won't). Switching modes uses an explicit v-if/v-else-if chain so each mode gets its own Field instance instead of reusing one across mode switches, which was causing the field label to freeze on whatever mode was active at first mount.

**Schema with no allow_when/deny_when field**
截屏2026-08-06 14 27 25

**Schema with allow_when/deny_when field**
截屏2026-08-06 14 27 58

### DIFF
--- a/packages/entities/entities-plugins/fixtures/schemas/acl.ts
+++ b/packages/entities/entities-plugins/fixtures/schemas/acl.ts
@@ -1,0 +1,127 @@
+import type { FormSchema } from '../../src/types/plugins/form-schema'
+
+// Mirrors the real Kong Gateway ACL plugin schema (`kong.plugins.acl.schema`).
+const aclSchema: FormSchema = {
+  type: 'record',
+  entity_checks: [
+    {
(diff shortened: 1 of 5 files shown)


Leopoldthecoder:
Shall we use a textarea for `allow_when` and `deny_when` inputs, since users are expected to fill in an expression?

2eha0 (approved):

TT1228:
> Shall we use a textarea for `allow_when` and `deny_when` inputs, since users are expected to fill in an expression?

I'm not sure, the conditions field is using input. I will forward this to Julie to find out if she request any updates on `allow_when`/`deny_when` field.

Leopoldthecoder (approved):

PR #3690, opened 2026-08-12
fix(dashboard-renderer): strip unsupported platform filters [MA-5344]

filipgutica:
# Summary

[MA-5344](https://konghq.atlassian.net/browse/MA-5344)

Although `ui-apps` hides these operators for non-table platform tile filters, preset and dashboard-level filters can still reach the tile aggregation query. The aggregation endpoint rejects them.

- For non-table visualizations, remove `empty` and `not_empty` from `platform_usage` and legacy `platform` queries after query and dashboard context filters merge.
- Keep other platform operators and all non-platform filters.
- Add regression tests for both datasource aliases, query filters, and dashboard context filters.

[MA-5344]: https://konghq.atlassian.net/browse/MA-5344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics/dashboard-renderer/src/composables/useIssueQuery.ts
+++ b/packages/analytics/dashboard-renderer/src/composables/useIssueQuery.ts
@@ -1,6 +1,6 @@
 import {
   type AllFilters, type AnalyticsBridge, type DatasourceAwareQuery, type ExploreFilterAll, type ExploreQuery,
-  type TimeRangeV4,
+  isPlatformDatasource, requestFilterTypeEmptyV2, type TimeRangeV4,
   type ValidDashboardChartQuery,
 } from '@kong-ui-public/analytics-utilities'
 import { useDatasourceConfigStore } from '@kong-ui-public/analytics-config-store'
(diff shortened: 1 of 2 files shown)


packages/analytics/dashboard-renderer/src/composables/useIssueQuery.ts
```diff
@@ -40,15 +40,18 @@ export default function useIssueQuery() {
...
+    // support `empty` and `not_empty` filters.
     const mergedFilters = stripUnknownFilters.value({
       datasource,
       filters: [
         ...(query.filters ?? []) as AllFilters[],
         ...context.filters,
       ],
       queryFields: query.metrics,
-    })
+    }).filter(({ operator }) => !isPlatformQuery || !requestFilterTypeEmptyV2.some(emptyOperator => emptyOperator === operator))
```
filipgutica: We can blindly strip all empty/not_empty filters for platform usage here because this is the path for non-table chart queries. 

For table tiles, querying happens through a different renderer/path

Chalks: so table tiles _can_ use empty, but everything else can't?

Chalks (approved):

PR #3706, opened 2026-08-18
feat(entities-plugins): add typed plugin context provide/inject mechanism

2eha0:
# Summary

- Adds a generic, typed provide/inject mechanism (`providePluginContext`/`usePluginContext`) under `free-form/shared/plugin-context/` for handing a plugin's free-form component host-supplied data (feature flags, API-fetched options, callbacks). Each plugin registers its own context type in a central `PluginContextRegistry`, so both functions are fully typed from the plugin name — no casts needed at call sites.
- This converges what was previously scattered across the catch-all `FORMS_CONFIG` blob and one-off `inject()` calls per plugin.
- `usePluginContext` returns `undefined` (never throws) when nothing was registered, and only logs a `console.warn` when some ancestor did provide context for at least one plugin but forgot this one — silent when the mechanism isn't in use at all, so existing hosts/tests aren't spammed with warnings for context that's meant to be optional.
- Adopts the new mechanism for `key-auth`: a `identityRealmsEnabled` context flag (default `true`) lets a host disable the Kong Identity "centrally managed consumers" flow entirely — hides the `identity_realms` field, skips the realms fetch, and hides the "Centrally managed consumers" mode option. `basic-auth` is unaffected since its schema never has `identity_realms`.
- Wires the sandbox `PluginFormPage` to call `providePluginContext('key-auth', { identityRealmsEnabled: true })` as a live usage example.
- New exports from the package entry (`src/index.ts`): `providePluginContext`, `usePluginContext`, `PluginContextRegistry`, `PluginContextName`.

## Test plan

- [x] `plugin-context/index.spec.ts` (vitest): covers ancestor→descendant read, multiple `providePluginContext` calls in one component merging instead of clobbering, warn-vs-silent behavior, and the setup()-only guard.
- [x] `ConfigFormContent.cy.ts` (Cypress component): new `identityRealmsEnabled context (key-auth)` describe block — hides the mode option, hides the field even with pre-existing centrally-managed data, skips the realms fetch, and confirms default-enabled behavior is unchanged. All 58 specs in the file pass (no regressions).
- [x] `npx eslint` and `npx vue-tsc -p tsconfig.build.json --noEmit` clean on all touched/added files (one pre-existing, unrelated `metadata.ts` error remains).

### DIFF
--- a/packages/entities/entities-plugins/sandbox/pages/PluginFormPage.vue
+++ b/packages/entities/entities-plugins/sandbox/pages/PluginFormPage.vue
@@ -70,7 +70,7 @@
 import { computed, defineComponent, provide, ref, type PropType } from 'vue'
 import { useRouter } from 'vue-router'
 
-import { PluginForm, TOASTER_PROVIDER, useProvideExperimentalFreeForms } from '../../src'
+import { PluginForm, providePluginContext, TOASTER_PROVIDER, useProvideExperimentalFreeForms } from '../../src'
 import { FEATURE_FLAGS } from '../../src/constants'
 
 import { ToastManager } from '@kong/kongponents'
@@ -126,6 +126,10 @@ const FeatureFlagProvider = defineComponent({
 })
 
 provideDeckCommandEditor()
+
+providePluginContext('key-auth', {
+  identityRealmsEnabled: true,
+})
 
 useProvideExperimentalFreeForms([
   'ace',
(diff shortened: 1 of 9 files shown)


TT1228:
> * (feature flags, API-fetched options, callbacks

Later do we need to refactor existing code to adopt this pattern? There are already many props derived from feature flags.

2eha0:
> > * (feature flags, API-fetched options, callbacks
> 
> Later do we need to refactor existing code to adopt this pattern? There are already many props derived from feature flags.

I think so.

TT1228 (commented):
In `ConfigFormContent.vue`(`keyAuth` and `basicAuth` form), `principalsDirectoryName`, `identityPrincipalsUiEnabled`, `FETCHED_REALMS_KEY`, `principalsCreationGuideVisible` and `dataPlaneVersions` depends on formConfig or `provide`. Would you rather do it in this PR too, or draft a new PR for them.

TT1228 (approved):

PR #3707, opened 2026-08-18
fix(entities-gateway-services): missing delete description in km

Leopoldthecoder:
# Summary

Adding the deletion description back, only for KM.

<p align=center>image</p>


2eha0 (approved):
Thank you! LGTM!

PR #3710, opened 2026-08-19
feat(entities-plugins): support AI Manager context for OIDC cache_tokens_salt

TT1228:
# Summary

KM-3117
Register an 'openid-connect' entry in the plugin-context registry so a host app can identify itself as 'ai-manager'. When set, cache_tokens_salt becomes required and is prefilled with a random 32-character value on create, since AI Manager needs a real salt up front to key its own token cache instead of relying on Kong's server-side auto-generation.

[TEST_COVERAGE_EXEMPT_REASON] The test will be in consuming app.


2eha0 (approved):
LGTM

jw-foss (approved):

hangrao (approved):

PR #3713, opened 2026-08-20
feat(table-data-grid): custom rendering and row click

afrzhussain:
### Summary

Extends TableDataGrid with the AG Grid features needed for the new Observability Logs view, while keeping all changes optional and backward-compatible. Existing consumers require no changes.

JIRA: https://konghq.atlassian.net/browse/OLLY-1237
Host PR: https://github.com/kong-konnect/shared-ui-components/pull/4289

### What's included

- Custom cell rendering — a template slot named after a column's key (e.g. #status) replaces that column's cell content entirely; the package's existing tooltip-on-overflow renderer remains the fallback for any column without a matching slot.
- Row & cell click — new row:click and cell:click emits, wired to AG Grid's onRowClicked/onCellClicked. A header's disableRowClick: true suppresses row:click for clicks landing in that column's cells (e.g. an actions column with its own buttons) without affecting cell:click or AG Grid's own selection mechanics.

### Screenshot( consumed host PR)
Screenshot 2026-08-21 at 12 07 11 AM


filipgutica (changes requested):
I'd like to review

filipgutica (changes requested):
Request changes.

This feature needs to preserve the ownership model established in https://github.com/Kong/public-ui-components/pull/3209:

- Headers define the column schema, capabilities, and initial/default state.
- `tableConfig` owns the current state and persisted overrides.
- Without controlled `tableConfig`, the component initializes internal state from the headers.
- The host can persist `tableConfig` for example within the dashboard definition (when used in dashboards) and replay it later.
- Fetchers receive package-owned request types.
- Dynamic column slots render custom cell content.
- AG Grid types and mechanics stay behind the wrapper.

The current implementation bypasses these boundaries for pinning, sorting, rendering, and row interactions.

Please update the README and sandbox with the same contract. The sandbox should also demonstrate controlled config and dynamic slots.

See https://github.com/Kong/public-ui-components/pull/3209 for context.

packages/core/table-data-grid/src/components/TableDataGrid.vue
```diff
@@ -113,6 +120,13 @@ const columnDefs = computed<Array<ColDef<Row>>>(() => headers.map((header) => {
     minWidth: header.minWidth,
     valueGetter: params => params.data?.[header.key],
     width: header.width,
+    pinned: header.pinned,
```
filipgutica: Blocking: please treat header pinning as initial/default state, not current grid state.

`headers[].pinned` defines how the column starts when no current state overrides it. `tableConfig.columns[key].pinned` must own the current and persisted value.

When controlled config is absent, the component can initialize internal config from the header default. The host can later save emitted config with the dashboard definition.

Please seed AG Grid from the header through `initialPinned`. Apply the resolved `tableConfig` as the active column state.

User pin and unpin actions must emit `update:tableConfig`. The config must support `pinned: false` to override a pinned header default.

Please cover the header default, a controlled override, explicit unpin, emitted changes, and config replay.

packages/core/table-data-grid/src/types/index.ts
```diff
@@ -9,18 +10,31 @@ export type TableDataGridStatePayload = {
...
   minWidth?: number
   maxWidth?: number
+  /** Pins the column to the left or right edge of the grid. */
+  pinned?: 'left' | 'right'
+  /** Vue component used to render this column's cells. */
+  cellRenderer?: Component
+  /** Extra props passed to `cellRenderer`. */
+  cellRendererParams?: Record<string, unknown>
+  /** Formats the raw cell value for display. */
+  valueFormatter?: ValueFormatterFunc<Row>
```
filipgutica: Blocking: please keep custom cell content slot-first, following the established KTableData pattern.

A dynamic slot named by `header.key` should render the column content. The slot should receive `{ row, rowValue, column, rowIndex, selected, refreshCell }`.

Keep raw AG Grid renderers, renderer parameters, and formatters under the documented `agGridColumnOptions` escape hatch. They should not become first-class header fields.

This keeps AG Grid internals behind the wrapper. It also preserves typed selection, tooltip, and renderer refresh behavior.

packages/core/table-data-grid/src/types/index.ts
```diff
@@ -9,18 +10,31 @@ export type TableDataGridStatePayload = {
...
+  /** Enables sorting UI for this column. The `fetcher` must apply `params.sort` server-side. */
+  sortable?: boolean
 }
 
 export interface TableDataGridInfiniteFetcherParams {
   mode: 'infinite'
   pageSize: number
   cursor?: unknown
+  sort?: SortModelItem[]
```
filipgutica: Blocking: please do not expose `SortModelItem` through the public fetcher contract.

`SortModelItem` is an AG Grid type and currently includes `type: 'default'`. This makes the backend contract depend on AG Grid's version and internal model.

The reference contract (https://github.com/Kong/public-ui-components/pull/3209) uses package-owned sort fields. Headers define whether a column supports sorting and can provide initial sort defaults.

`tableConfig` owns the current sort state. Sort changes must update that config and emit the package-owned `sort` and `update:tableConfig` events.

If multi-column sort is required, define and export a package-owned payload. Extend `tableConfig` and the events in the same change.

packages/core/table-data-grid/src/composables/useFetchInfinite.ts
```diff
@@ -299,10 +299,12 @@ export const useFetchInfinite = <Row extends object = TableDataGridRow>({
           // for the previous block, because that previous AG Grid range is what
           // produced the backend cursor needed to continue the chain.
           const cursor = blockIndex > 0 ? cursorMap.get(blockIndex - 1) : undefined
+
           const result = await fetcher({
             mode: 'infinite',
             pageSize,
             cursor,
+            sort: getRowsParams.sortModel,
```
filipgutica: Blocking: a sort change must start a new cursor chain and datasource generation.

A cursor belongs to the query state that produced it. Changing the sort makes every cursor from the previous sort invalid.

`cursorMap` and `blockCompletionMap` reset only when this composable rebuilds the datasource. Passing a new sort through the same datasource keeps the previous cursor state.

AG Grid can request block 1 before the new sorted block 0 resolves. That request can combine the new sort with an old cursor.

Please rebuild the datasource when the resolved sort changes. The new generation must reject responses from requests that used the previous sort.

Please add a test with a delayed old request and a second block request after sorting.

packages/core/table-data-grid/src/components/TableDataGrid.vue
```diff
@@ -149,6 +163,12 @@ useEmitState({
 const onGridReady = (event: GridReadyEvent<Row>) => {
   emit('grid:ready', event.api)
 }
+
+const onRowClicked = (event: RowClickedEvent<Row>) => {
+  if (event.data) {
+    emit('row:click', event.data)
```
filipgutica: Blocking: slot-based cells can contain buttons, links, menus, and other controls.

The current handler emits `row:click` for every column and discards the source event. A click inside an action slot will also trigger the row action.

Please add `headers[].disableRowClick` and suppress the event for those columns. Keep normal AG Grid selection behavior unchanged.

Please also emit `(row, event)` and provide the separate `cell:click` contract from #3209.

Add coverage for a clickable action slot that does not emit `row:click`.

packages/core/table-data-grid/src/types/index.ts
afrzhussain: updated implementation

packages/core/table-data-grid/src/components/TableDataGrid.vue
afrzhussain: updated implementation

packages/core/table-data-grid/src/types/index.ts
```diff
@@ -9,12 +9,35 @@ export type TableDataGridStatePayload = {
...
   key: Extract<keyof Row, string>
   label: string
   width?: number
   minWidth?: number
   maxWidth?: number
+  /**
+   * Disables the `row:click` emit for clicks landing inside this column's
+   * cells, e.g. an actions column with its own buttons. AG Grid's own
+   * selection/click mechanics are unaffected.
+   */
```
filipgutica, line 40: Blocking: these additions create new public package contracts, but the README still documents only the original API.

Please document:

- `disableRowClick`
- Column-key slots and the complete slot payload
- `row:click` and its `(row, event)` signature
- The `cell:click` payload
- The new exported payload types

The README, sandbox, exported types, and contract tests should use the same names and payloads.

packages/core/table-data-grid/src/components/TableDataGrid.cy.ts
```diff
@@ -824,4 +829,44 @@ describe('<TableDataGrid />', () => {
...
+        status: ({ rowValue }: TableDataGridCellSlotProps<TestRow>) => h(
+          'span',
+          { 'data-testid': 'custom-status-badge' },
+          `Badge: ${rowValue}`,
+        ),
+      } as TestTableDataGridSlots,
+    })
+
+    cy.getTestId('custom-status-badge').should('contain.text', 'Badge: Active')
+  })
```
filipgutica, line 878: Blocking: the slot-first direction is correct, but the current test only checks `rowValue`.

Please cover the complete public payload: `row`, `rowValue`, `column`, `rowIndex`, `refreshCell`, and `selected`.

Please also verify that `refreshCell` updates reused renderer content. The renderer lifecycle is part of the package contract.

packages/core/table-data-grid/src/components/TableDataGrid.cy.ts
```diff
@@ -824,4 +829,44 @@ describe('<TableDataGrid />', () => {
...
+    })
+
+    mountTestTableDataGrid({
+      fetcher,
+      onRowClick,
+    })
+
+    cy.contains('.ag-row', 'Gateway service').click()
+    cy.get('@rowClick').should('have.been.calledOnceWith', rows[0])
+  })
```
filipgutica, line 944: Blocking: this test covers a normal row click, but it does not cover the behavior behind `disableRowClick`.

Please add an action-column slot with `disableRowClick: true`. The test should click a button inside that slot.

The test should prove that:

- `row:click` is not emitted.
- `cell:click` still emits `{ row, columnKey, value }`.
- A normal `row:click` includes the second event argument.

This protects the action-cell behavior defined in the specification.

filipgutica (changes requested):
Thanks for reworking this. The direction now matches the contract in the spec and PR #3209 much more closely.

The earlier architectural blockers are resolved:

- Pinning and sorting no longer add unmanaged AG Grid state to the public API.
- Custom cell content now uses column-key slots instead of renderer callbacks.
- Row and cell interactions now support `disableRowClick`, `(row, event)`, and `cell:click`.

I have two remaining blockers before approval: public documentation and contract test coverage. I left focused comments for both areas.

filipgutica (approved):

PR #3715, opened 2026-08-20
fix(analytics): validate zoom range for time series chart [MA-5288]

jordojordo:
# Summary

Fix [MA-5288](https://konghq.atlassian.net/browse/MA-5288)

This will fix two issues when zooming on a timeseries chart:
1. Selecting a range where the start and end time are the same
2. Selecting a time on the chart without dragging (zooming) would cause the next drag time to jump to the first time that was clicked.

If the user ends up with the same start and end time when zooming, the dispatched action will simply not be emitted, which prevents an invalid time range from being selected.

<details closed>
<summary>Before</summary>

[before.webm](https://github.com/user-attachments/assets/e00dc6c8-8826-43e7-bd17-673c16d1eae3)

</details>

<details closed>
<summary>After</summary>

[after.webm](https://github.com/user-attachments/assets/2727286b-6bd0-416a-9a0b-36f506b202b2)

</details>

[MA-5288]: https://konghq.atlassian.net/browse/MA-5288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics/analytics-chart/src/components/AnalyticsChart.cy.ts
+++ b/packages/analytics/analytics-chart/src/components/AnalyticsChart.cy.ts
@@ -91,6 +91,32 @@ const selectChartArea = () => {
   })
 }
 
+const DRAG_WAIT = 300
+
+const pressAndHoldChartArea = ({ x, y }: { x: number, y: number }) => {
+  const selector = '.chart-container > canvas'
+
+  cy.get(selector).trigger('mousedown', x, y)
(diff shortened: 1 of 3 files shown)


filipgutica (changes requested):
One minor bug mentioned in slack - zoom tooltip not cleared/removed when generating a zoom where start === end

filipgutica (approved):

PR #3751, opened 2026-08-27
feat(analytics): dashboard context type refactors and prop updates [MA-5394]

Chalks:
Supports [MA-5394](https://konghq.atlassian.net/browse/MA-5394)

# Summary

These are non-breaking changes in order to get the dashboard context to be a single source of truth that exposes all possible configuration externally. Currently we have a decent amount of configuration that is inferred from various properties, and this refactor is just making it easier to see where those places are.

Everything that consumes DashboardRenderer externally shouldn't have anything change due to this PR.

[MA-5394]: https://konghq.atlassian.net/browse/MA-5394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analytics/dashboard-renderer/src/composables/useDashboardContext.ts
+++ b/packages/analytics/dashboard-renderer/src/composables/useDashboardContext.ts
@@ -12,7 +12,7 @@ import {
 } from '../constants'
 import { useAnalyticsConfigStore, useDatasourceConfigStore } from '@kong-ui-public/analytics-config-store'
 
-export default function useDashboardInternalContext({
+export default function useDashboardContext({
   context,
   globalFilters = ref([]),
   isFullscreen = ref(false),
@@ -23,12 +23,20 @@ export default function useDashboardInternalContext({
   isFullscreen?: Readonly<Ref<boolean>>
   preview?: Readonly<Ref<boolean>>
 }): {
-  internalContext: Readonly<Ref<DashboardRendererContext>>
+  editable: Readonly<Ref<boolean>>
+  enrichedContext: Readonly<Ref<DashboardRendererContext>>
+  filters: Readonly<Ref<AllFilters[]>>
   queryReady: Readonly<Ref<boolean>>
+  refreshInterval: Readonly<Ref<number>>
+  showTileActions: Readonly<Ref<boolean>>
+  showTileZoomActions: Readonly<Ref<boolean>>
+  timeSpec: Readonly<Ref<TimeRangeV4>>
+  tz: Readonly<Ref<string>>
+  zoomable: Readonly<Ref<boolean>>
 } {
   const configStore = useAnalyticsConfigStore()
   const datasourceStore = useDatasourceConfigStore()
-  const { loading: configloading } = storeToRefs(configStore)
+  const { loading: configLoading } = storeToRefs(configStore)
   const { loading: datasourceLoading } = storeToRefs(datasourceStore)
 
   const timeSpec = computed<TimeRangeV4>(() => {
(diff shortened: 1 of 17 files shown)


packages/analytics/dashboard-renderer/src/composables/useDashboardContext.ts
```diff
@@ -42,14 +50,56 @@ export default function useDashboardInternalContext({
...
+    }
+
+    if (showTileActions === undefined) {
+      return true
+    }
+
+    return showTileActions
+  })
+
+  const zoomable = computed<boolean>(() => {
```
filipgutica, line 89: Just want to clarify this:

This PR adds `zoomable` and `showTileZoomActions` to the public contract via `DashboardRendererContext`. That makes them look configurable even though this logic remains their only source. Could we keep them out of the caller context? 

Or was the intention to have their caller-provided values override the computed values here?

packages/analytics/dashboard-renderer/src/composables/useDashboardContext.ts
```diff
@@ -89,37 +139,47 @@ export default function useDashboardInternalContext({
...
+      editable: editable.value,
+      showTileActions: showTileActions.value,
+      showTileZoomActions: showTileZoomActions.value,
+      zoomable: zoomable.value,
     }
   })
 
   const queryReady = computed<boolean>(() => {
-    return !!context.value.timeSpec || (!configloading.value && !datasourceLoading.value)
+    return !configLoading.value && !datasourceLoading.value
```
filipgutica, line 163: Why did we get rid of `!!context.value.timeSpec`

filipgutica (dismissed):
Couple comments

packages/analytics/dashboard-renderer/src/types/renderer-types.ts
```diff
@@ -13,16 +13,8 @@ export interface DashboardRendererContext {
...
-// The DashboardRenderer component fills in optional values before passing them down to the tile renderers.
-export interface DashboardRendererContextInternal extends Required<DashboardRendererContext> {
-  zoomable: boolean
-  /**
-   * only used internally as controlling this externally is just a matter of
-   * including a zoom handler on the DashboardRenderer component.
-   */
-  showTileZoomActions: boolean
+  zoomable?: boolean
+  showTileZoomActions?: boolean
```
filipgutica, line 17: I don't think these should be public-facing, since they are computed internally? 

Re. https://github.com/Kong/public-ui-components/pull/3751#discussion_r3876205257

packages/analytics/dashboard-renderer/src/composables/useDashboardContext.ts
Chalks: I'm going to remove the computed values in the next PR and replace them with a `zoomConfiguration` object. I mainly did it just for keeping types happy without making a ton of changes.

packages/analytics/dashboard-renderer/src/composables/useDashboardContext.ts
Chalks: `context.value.timeSpec` is being replaced by anyone who uses this composable with `enrichedContext.value.timeSpec` (which is in turn just `timeSpec.value` from lines  42-51 of this composable). That value is guaranteed to be defined by this composable. So checking `!!timeSpec.value || ...` doesn't actually doing anything, as it will always have a value.

packages/analytics/dashboard-renderer/src/types/renderer-types.ts
Chalks: same answer https://github.com/Kong/public-ui-components/pull/3751#discussion_r3882365409

jordojordo (approved):

PR #3758, opened 2026-08-31
feat(monaco-editor): add zod-based YAML schema validation

2eha0:
## Summary
- `createZodValidator`/`collectValidators` (features) adapt Zod schemas into composable validators.
- `runYAMLValidation` (languages/yaml/validation.ts) runs those validators against a YAML model and maps issues back to source ranges (`ancestor` underlines a whole block, `document` anchors at the file start).
- Exports `getYAMLNodeRange`/`getYAMLNodeAtPath` from code-lenses.ts so validation.ts can reuse them.

### DIFF
--- a/packages/core/monaco-editor/src/languages/yaml/validation.ts
+++ b/packages/core/monaco-editor/src/languages/yaml/validation.ts
@@ -1,0 +1,102 @@
+import { getModelContext } from '../../singletons/model-contexts'
+import { getYAMLNodeAtPath, getYAMLNodeRange } from './code-lenses'
+
+import type { editor, IRange } from 'monaco-editor'
+import type { Document as YAMLDocument } from 'yaml'
+
+import type { ValidateFn, ValidationIssue } from '../../features/validation'
+import type { ModelContextGetter } from '../../types'
+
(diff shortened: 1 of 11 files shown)


packages/core/monaco-editor/src/languages/yaml/validation.ts
```diff
@@ -0,0 +1,102 @@
...
+  const context = await (options?.contextGetter ?? getModelContext)(model)
+  if (context.isDefault || context.language !== 'yaml' || !context.document) {
+    return []
+  }
+
+  const document = context.document
+
+  let value: unknown
+  try {
+    value = document.toJS()
```
sumimakito, line 97: QQ: Do we need to take care of `document.errors`?

2eha0: Good catch — added a check for `document.errors` that bails out before schema validation runs, plus tests covering a few error cases (649db754c). Thanks!

sumimakito (approved):
LGTM

arashsheyda (approved):

PR #3760, opened 2026-08-31
feat(entities-plugins): support expressible plugin config fields

TT1228:
# Summary

KM-3155

The Gateway now marks a config field `expressible`, meaning its value can alternatively be supplied as an expression evaluated on every request, which takes priority over the plain value. The expression lives in a twin field under a new root-level `expressions` record that mirrors `config`.

Renders this generically, from the schema convention rather than per plugin:

- `useExpressionField` maps a field path to its twin (first segment swapped for `expressions`, array indices preserved) and owns its read/write/clear.
- `ExpressionEditor` is the collapsible "+ Add expression" control.
- `ExpressionField` pairs a field's plain input with its editor, and is what `Field.vue` dispatches to for any field with a twin — so a plugin override replaces both halves rather than having the expression appended to it. The type mapping moved to `field-dispatch.ts` so both can share it.
- `StandardLayout` keeps `expressions` in the schema it hands to the form and in the model keys it controls; without either, every control silently disappears and saved expressions never load.
- `Form` never renders the `expressions` record as a field of its own, and `buildFormSchema` skips it so VFG does not flatten it into stray inputs.

rate-limiting gains the controls automatically. rate-limiting-advanced places them per limit row, keeping `expressions.limit` index-aligned with `config.limit` across add/remove/preset, and picks up its new `custom_key`. Its request-limits card is rebuilt to the new design: window type as radio cards, "Add limit" replacing the per-row buttons, "Custom" placeholders.

rate-limiting becomes a folder-based form so `limit_by` leads it instead of sitting in Advanced, using `ObjectField`'s existing `omit`/`fieldsOrder`.

Also adds filler handlers for expression fields, an example to the free-form mocking schema, and the two plugin schema fixtures.


packages/entities/entities-plugins/src/components/free-form/plugins/rate-limiting-advanced/RequestLimitsForm.vue
```diff
@@ -157,36 +182,76 @@ const requestLimits = computed<RequestLimit[]>(() => {
...
-const addRequestLimit = (index: number) => {
+/**
+ * The Gateway pairs `expressions.limit` with `config.limit` by position, so any
+ * insert or removal in one has to happen in the other.
+ *
+ * Only runs when the twin array already exists: without it no row has an
+ * expression, nothing can drift, and creating it here would submit a run of
+ * empty slots.
+ */
+const alignExpressionLimits = (mutate: (limits: Array<string | EmptyValue>) => void) => {
```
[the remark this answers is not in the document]
TT1228: We are not introducing the generic expression for `ArrayField` in this PR

packages/entities/entities-plugins/src/components/free-form/plugins/_shared/use-expression-mode.ts
```diff
@@ -0,0 +1,52 @@
...
+ * with nothing else changed.
+ *
+ * @param schema The plugin's raw schema, from `PluginFormLayoutProps`.
+ */
+export function useExpressionMode(schema: () => FormSchema) {
+  // Defaults to on: `KM-3034-features-316` is not registered in the consuming
+  // app yet, so an absent flag has to mean "show it" or the feature would be
+  // invisible everywhere. Once the flag exists, whatever it provides wins —
+  // including turning this off — and this default should become `false`.
+  const enabled = inject<boolean>(FEATURE_FLAGS.KM_3034_FEATURES_316, true)
```
TT1228: `FEATURE_FLAGS.KM_3034_FEATURES_316` was temporarily set to true to enable it on preview link

packages/entities/entities-plugins/src/components/free-form/plugins/rate-limiting-advanced/RateLimitingAdvancedForm.vue
```diff
@@ -12,12 +13,17 @@ import { AUTOFILL_SLOT, AUTOFILL_SLOT_NAME } from '@kong-ui-public/forms'
...
+import { useExpressionMode } from '../_shared/use-expression-mode'
 
 import type { PluginFormLayoutProps as Props } from '../../shared/layout/provider'
 import type { FreeFormPluginData } from '../../../../types/plugins/free-form'
 
 const props = defineProps<Props>()
 
+// `limit` and `custom_key` are expressible; gate their expression editors
+// with the rest of the 3.16 features, leaving the fields themselves.
+const { gatedSchema } = useExpressionMode(() => props.schema)
```
2eha0, line 36: Why do we need `useExpressionMode`?

packages/entities/entities-plugins/src/types/plugins/form-schema.ts
```diff
@@ -103,6 +125,24 @@ export interface StringFieldSchema extends FieldSchema {
...
+ * It always holds the expression source as a string. `expressible_kong_type` is
+ * the type the expression must evaluate to, and `source_field` mirrors the
+ * schema of the field it overrides.
+ *
+ * Not a member of `UnionFieldSchema`: it is structurally a `string` field and is
+ * rendered as one. The extra keys are read through this type where needed.
+ */
+export interface ExpressionFieldSchema extends StringFieldSchema {
+  expressible_kong_type: FieldSchemaType
+  source_field: UnionFieldSchema
```
2eha0: Where does the `source_field` come from?

packages/entities/entities-plugins/src/components/free-form/shared/ExpressionEditor.vue
```diff
@@ -0,0 +1,198 @@
+<template>
+  <div
+    v-if="expression.available.value"
```
2eha0, line 12: Duplicated available check

packages/entities/entities-plugins/src/components/free-form/shared/ExpressionField.vue
```diff
@@ -0,0 +1,100 @@
...
+    <!--
+      The plain value, rendered by whatever component its schema type maps to.
+      Deliberately not a `Field`: that would dispatch back here and recurse.
+    -->
+    <component
+      :is="valueComponent"
+      v-if="valueComponent"
+      :autofocus="autofocus"
+      :name="absoluteName"
+    />
```
2eha0, line 20: This doesn't seem to be the right way to integrate the editor into the existing fields.

2eha0: Can you verify whether the `filedRenderers` still work for the expressible field?

packages/entities/entities-plugins/src/components/free-form/plugins/rate-limiting/RateLimitingForm.vue
```diff
@@ -0,0 +1,45 @@
+<template>
```
2eha0, line 1: I don't think RL needs its own template.

packages/entities/entities-plugins/src/components/free-form/plugins/rate-limiting-advanced/RateLimitingAdvancedForm.vue
TT1228: It's used to override the schema `expressible: true` so the expression will be gated.

packages/entities/entities-plugins/src/types/plugins/form-schema.ts
TT1228: This comes from schema by KoKo, we can keep it or remove it. I will post in KoKo team thread to know whether this field will be meaningful.
```
{
    "second": {
        "expressible_kong_type": "number",
        "len_max": 1024,
        "len_min": 0,
        "source_field": {
            "description": "The number of HTTP requests that can be made per second.",
            "expressible": true,
            "gt": 0,
            "type": "number"
        },
        "type": "string"
    }
}
```

packages/entities/entities-plugins/src/components/free-form/plugins/rate-limiting/RateLimitingForm.vue
TT1228: RL have some fields reordered and `custom_key` customized, let me take a look to find out whether they can be expressed in a ts file.

packages/entities/entities-plugins/src/components/free-form/plugins/rate-limiting-advanced/RateLimitingAdvancedForm.vue
2eha0: Why can't the expression be gated by `useExpression`?

packages/entities/entities-plugins/src/types/plugins/form-schema.ts
2eha0: I don't see `source_field ` in Admin API, it seems to appear only in Koko, let's remove it.

packages/entities/entities-plugins/src/components/free-form/plugins/rate-limiting/RateLimitingForm.vue
2eha0: I think this should work
```
renderRules: {
   bundles: [['config.limited_by', 'config.custom_key']]
}
```

2eha0: If the promotion of `limited_by` is necessary, I recommend introducing a new feature in `renderRules`
```
renderRules: {
  bundles: [['config.limited_by', 'config.custom_key']],
  promoted: ['config.limited_by'], // <-
}
```

packages/entities/entities-plugins/src/components/free-form/plugins/rate-limiting-advanced/RateLimitingAdvancedForm.vue
TT1228: I was thinking we better separate it from `useExpressionField`. `useExpressionMode` is more about release and will be removed soon.

packages/entities/entities-plugins/src/components/free-form/plugins/rate-limiting/RateLimitingForm.vue
TT1228: Let me create a ticket for this `promoted` pattern, can we discuss how to implement that in a follow-up PR?

packages/entities/entities-plugins/src/components/free-form/shared/ExpressionField.vue
TT1228: I used `<FieldRenderers />` to wrap `CustomKey`(which wraps `ExpressionField`) and it's working as expected.

2eha0 (approved):

2eha0:
Due to release time pressure, let's merge it first and optimize the code later.

PR #3761, opened 2026-08-31
feat(table-data-grid): add single-column sorting [OLLY-1244]

afrzhussain:
### Summary

JIRA: [OLLY-1244](https://konghq.atlassian.net/browse/OLLY-1244)
Arch ref: [link](https://docs.google.com/document/d/1aKjabauYhCdWIwDLrz1eNjPT-CkzR366_pNy1xtkh28/edit?tab=t.0#heading=h.ti92zrd8x6co)
Github POC PR: [link](https://github.com/Kong/public-ui-components/pull/3209)

Adds single-column sorting to `TableDataGrid`.

- Mark a column sortable via `header.sortable` (optionally `header.showSortIcon`); AG Grid renders its built-in sort control and handles the click.
- New `tableConfig` prop holds the current sort + page size, controlled or uncontrolled. A grid-driven sort change emits `sort` (sort-only payload) then `update:tableConfig` (full config).
- The public fetcher contract gains a `sort` param. A sort change is treated as a request-context change lik the cursor chain resets from block 0, since a cursor produced under one sort order isn't valid underanother.
- Multi-column sort is disabled — `suppressMultiSort` plus a structural backstop that collapses to the most more than one ends up sorted through any other path.

### Demo

https://github.com/user-attachments/assets/a93aa189-73ee-441a-805f-ee2ff3afe537

[OLLY-1244]: https://konghq.atlassian.net/browse/OLLY-1244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/table-data-grid/src/types/index.ts
```diff
@@ -39,12 +39,32 @@ export type TableDataGridHeader<Row extends object = TableDataGridRow> = {
...
    */
   disableRowClick?: boolean
+  /** Enables sorting on this column via AG Grid's built-in header sort control. */
+  sortable?: boolean
+  /**
+   * Shows the unsorted sort icon on this column's header even when it isn't
+   * the active sort, instead of only on hover or once sorted. Only relevant
+   * when `sortable` is true.
+   */
+  showSortIcon?: boolean
```
Chalks, line 49: **opinion, not a blocker**

When we have multiple pieces of configuration that affect the same thing, I like nesting it and/or namespacing it. I would prefer something like:

```
sortEnabled?: boolean
sortShowUnsortedIcon?: boolean
```

or something like:

```
sort?: {
  enabled: boolean
  showUnsortedIcon: boolean
}
```

very much opinion territory. Feel free to dismiss this feedback.

packages/core/table-data-grid/src/composables/useTableDataGridSort.ts
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @param activeSort Current resolved sort, read from `useTableDataGridConfig`.
+ * @param emitSort Called with the new sort whenever a grid interaction changes it.
+ * @param patchTableConfig Writes the new sort into the current `tableConfig`.
+ */
+export const useTableDataGridSort = <Row extends object = TableDataGridRow>({
+  activeSort,
+  emitSort,
+  patchTableConfig,
+}: {
```
Chalks, line 17: **question**

I'm a little unsure of what's happening here. It looks like you're accomplishing sort by forcing a configuration change that then triggers a refetch of data? The consequence of this is that instead of getting the nice sort animation that happens in ag grid natively, we get the full refresh behavior. Compare the sort video in the body of this PR to this demo from ag grid:

https://github.com/user-attachments/assets/191f4e53-a69e-4943-a359-973f356d9363

Basically the question is... can I still do basic client side sorting like ag grid supports ([described here](https://www.ag-grid.com/javascript-data-grid/row-sorting/#custom-sorting))? For lots of cases, I don't need to refetch on a sort, I just need client side sort.

packages/core/table-data-grid/src/types/index.ts
afrzhussain: Hmm, I partly agree with this as well. Since the architecture for this was already defined 
[here](https://docs.google.com/document/d/1aKjabauYhCdWIwDLrz1eNjPT-CkzR366_pNy1xtkh28/edit?tab=t.0#heading=h.ti92zrd8x6co), I was mainly following the planned approach and keeping the changes consistent with it.

packages/core/table-data-grid/src/composables/useTableDataGridSort.ts
afrzhussain: The current implementation follows the [ERD](https://docs.google.com/document/d/1aKjabauYhCdWIwDLrz1eNjPT-CkzR366_pNy1xtkh28/edit?tab=t.0#heading=h.ti92zrd8x6co), whenever the request context changes (filters, search, page size, refresh, and now sort), we refetch, so sort is handled the same way as the others.

That said, good point about client-side sorting. The ERD didn’t really cover that use case, so it’s more of a gap than an intentional decision. We could add it as a separate mode using AG Grid’s `clientSide` row model for cases where all the data is already loaded. Happy to track that as a follow-up.

Chalks: yep, totally fine to leave it as a followup, thank you!

Chalks (approved):

PR #3771, opened 2026-09-03
feat(freeform): scaffold new blank package

2eha0:
# Summary

Add an empty `@kong-ui-public/freeform` package under `packages/core`, generated from the standard package template (same structure/conventions as other `core` packages).

This package will host the free-form plugin configuration framework currently living in `@kong-ui-public/entities-plugins`. It is intentionally blank for now — the actual migration of the free-form framework code will happen in a follow-up PR.


CODEOWNERS
```diff
@@ -38,6 +38,7 @@
 /packages/core/entities-config-editor/ @Kong/team-km @Kong/team-core-ui
 /packages/core/expressions/ @Kong/team-km @Kong/team-core-ui
 /packages/core/table-data-grid @Kong/team-data @Kong/team-core-ui
+/packages/core/freeform/ @Kong/team-km @Kong/team-core-ui
```
adamdehaven, line 41: If this is specifically for plugins, should the package name be more specific?

2eha0: Not exactly. It consists solely of form UI components and plugin bindings; the UI layer itself could be made more generic.

ValeryG (approved):

PR #3780, opened 2026-09-04
feat(analytics): cleaner zoom config and respecting explorer context [ma-5352]

Chalks:
[Satisfies MA-5352](https://konghq.atlassian.net/browse/MA-5352)

# Summary

* Removes `zoomable` and `showTileZoomActions` from `DashboardRendererContext` - they have never been used, and they were added by me a couple days ago as a stepping stone to get to this PR.
* Updates `useDashboardContext` to return `ZoomConfiguration` that addresses what `zoomable` and `showTileZoomActions` were trying to do
* Adds `vue-router` so that we can detect if we're in explorer, and disables the explore action in zoom if we are

### Screenshots

Screenshot 2026-09-04 at 12 32 42 PM

### Preview

https://github.com/kong-konnect/konnect-ui-apps/pull/13761


packages/analytics/dashboard-renderer/src/composables/useDashboardContext.ts
```diff
@@ -154,32 +144,79 @@ export default function useDashboardContext({
...
+  const showZoomExploreAction = computed<boolean>(() => {
+    const isExploreRoute = route?.path.endsWith('analytics/explorer') ?? false
+    return !preview.value && !isExploreRoute
+  })
+
+  const showZoomRequestsAction = computed<boolean>(() => {
+    return !preview.value
+  })
+
+  const zoomConfiguration = computed<ZoomConfiguration>(() => {
```
Chalks, line 170: This file contains the bulk of the important changes in this PR, and this `zoomConfiguration` in particular is the most important bit. Essentially the idea is that we have some properties that are inferred by other things, why are we including them in our config at all (privately or otherwise)? Now it's just built into this composable instead.

adorack (changes requested):
I'm really not comfortable using `vue-router` in this way.  Let's talk about it.

packages/analytics/dashboard-renderer/src/components/DashboardTile.spec.ts
```diff
@@ -257,9 +257,31 @@ const groupedMetricsResult = {
 
 describe('<DashboardTile /> zoom requests drilldown', () => {
   beforeEach(() => {
+    vi.restoreAllMocks()
     setupPiniaTestStore()
   })
 
+  it.skip('does not populate explore zoom actions if on the explorer route', async () => {
```
adorack: This test should probably be removed.

adorack (changes requested):

packages/analytics/dashboard-renderer/src/components/DashboardTile.spec.ts
Chalks: whoopsy, meant to. thanks

Chalks: done

adorack (approved):

PR #3781, opened 2026-09-04
feat(analytics): add api-requests to query bridge [MA-5440]

jordojordo:
# Summary

Close [MA-5440](https://konghq.atlassian.net/browse/MA-5440)

This will allow the `api-requests` endpoint to be queried through the query bridge, allowing for request records to be used in dashboards and elsewhere.

[MA-5440]: https://konghq.atlassian.net/browse/MA-5440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/analytics/analytics-utilities/src/types/explore/requests.ts
```diff
@@ -170,3 +170,80 @@ export interface RequestQuery {
   classified?: boolean
   cursor?: string
 }
+
+export const API_REQUESTS_MAX_PAGE_SIZE = 1000
+
+export interface RequestAiEntry {
```
adorack: For API requests in particular, we might not want to specify all of these types in `analytics-utilities`.  Config comes from the `/config` endpoint, and that should be valid for API requests, and that can drive which dimensions we try to extract from the response for use in charts and so forth.

I think it's fine to type it as `[field: string]: unknown`, or even just to type the `results` key as an object.  Happy to discuss further; I don't have full visibility into what you're doing with it.

adorack (changes requested):

packages/analytics/analytics-utilities/src/types/explore/requests.ts
jordojordo: That's fair, I can remove those types from here and provide it from the `/config` endpoint instead.

I'll keep the `RequestRecord` as a `[field: string]: unknown` to keep things at least _slightly_ typed.

adorack (approved):

PR #3810, opened 2026-09-09
feat(monaco-editor): monaco inline diff editor[DRWHO-655]

sharmaadityaHQ:
# Summary

This PR aims to add an inline monaco diff editor.

**Related jira** - https://konghq.atlassian.net/browse/DRWHO-655

**Preview**

https://github.com/user-attachments/assets/404e0e80-6e78-49d4-92bf-13e1dee24a11


adamdehaven (commented):
Why do we need a totally separate DiffEditor instance rather than mounting two of the existing component with the proper configuration? 

We have a much simpler implementation currently for the Portal Editor in Konnect-ui-apps; could this be simplified?

sharmaadityaHQ:
> We have a much simpler implementation currently for the Portal Editor in Konnect-ui-apps; could this be simplified?

@adamdehaven i feel the portal usecase (toggle diff) is different due to which both editor and diffEditor are handled in the same composable while using 2 different targets here - https://github.com/kong-konnect/konnect-ui-apps/blob/main/apps/portals/src/components/portal-studio/StudioCodeEditor.vue#L19.

here, 2 separate components/composables seem simpler, relevant and consumers can use one or compose both themselves basis usecase. let me know your thoughts or if i'm missing something here.

adamdehaven:
Might be worth having @Justineo or his team take a look as well before merging since their team helped create the shared package.

arashsheyda (changes requested):
I'd like to review as well. also the dark mode is broken when ran locally

packages/core/monaco-editor/src/composables/useMonacoDiffEditor.ts
```diff
@@ -0,0 +1,178 @@
...
+      }
+
+      if (!modifiedModel) {
+        const uri = monaco.Uri.parse(`inmemory://model/diff-modified/${options.language}-${crypto.randomUUID()}`)
+        modifiedModel = monaco.editor.createModel(toValue(options.modified), options.language, uri)
+      } else {
+        modifiedModel.setValue(toValue(options.modified))
+      }
+
+      const themeName = editorStates.theme === 'light' ? 'catppuccin-latte' : 'material-theme-darker'
```
arashsheyda, line 92: I think the logic here is pretty much duplicate of `useMonacoEditor`, we can have another  compsoable for shared logic e.g. `useMonacoLifecycle` and utils like `createOrUpdateModel`, `getMonacoTheme`

sharmaadityaHQ: i'll have a look on what all can be extracted into utils. some parts here look similar but have different logic in both composables.

2eha0 (commented):
My suggestion is that this function should be added to the basic editor as a composable capability, rather than creating a new xxxEditor every time a capability is added.

sharmaadityaHQ:
> rather than creating a new xxxEditor every time a capability is added.

@2eha0 i feel this is true for other capabilities that were added to the same MonacoEditor component but for diffing Monaco requires a separate `createDiffEditor`.

Also, if i try to extend the current MonacoEditor component by using the new composable (useMonacoDiffEditor) it would make a few things in the component complex. eg. v-if's around toolbar, empty state, diff target, defineExpose change basis mode, check on wrong props combination from a consumer (v-model/toolbar, original, modified).

2eha0:
@sharmaadityaHQ Thanks for the explanation.
> for diffing Monaco requires a separate createDiffEditor.

If so, that makes sense to me.

packages/core/monaco-editor/src/composables/useMonacoDiffEditor.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  /** Remeasure fonts in the diff editor with debouncing to optimize performance */
+  const remeasureFonts = useDebounceFn(() => monaco.editor.remeasureFonts(), 200)
+
+  const init = (): void => {
+    loadMonaco()
+
+    // `toValue()` safely unwraps refs, getters, or plain elements
+    watch([isMonacoLoaded, () => toValue(target)], ([_isLoaded, _target], [, previousTarget]) => {
```
2eha0, line 63: Can we extract this `watch` out of the `init`?

sharmaadityaHQ: I am following the same pattern as in `useMonacoEditor` here - https://github.com/Kong/public-ui-components/blob/main/packages/core/monaco-editor/src/composables/useMonacoEditor.ts#L202. The `init` is only called once.

packages/core/monaco-editor/src/composables/useMonacoDiffEditor.ts
sharmaadityaHQ: [26f4075](https://github.com/Kong/public-ui-components/pull/3810/changes/26f407562f903a95577c72aeb97e81d50ecd54b0)

packages/core/monaco-editor/src/composables/useMonacoDiffEditor.ts
```diff
@@ -0,0 +1,178 @@
...
+
+      _isSetup = true
+      editorStates.editorStatus = 'ready'
+      editorStates.hasContent = !!toValue(options.modified)
+      editorStates.currentLanguage = modifiedModel.getLanguageId()
+
+      // Track language changes on the modified model
+      trackDisposableForModel(modifiedModel, modifiedModel.onDidChangeLanguage((e) => {
+        editorStates.currentLanguage = e.newLanguage
+      }))
```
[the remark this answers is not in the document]
sharmaadityaHQ: [763f1a](https://github.com/Kong/public-ui-components/pull/3810/changes/763f1af5a82abca000d49aeeb1462a81337589c2)

2eha0 (approved):
LGTM

arashsheyda (approved):

PR #3812, opened 2026-09-09
feat(analytics): add scatter plot chart type [MA-5431]

jordojordo:
# Summary

Satisfy [MA-5431](https://konghq.atlassian.net/browse/MA-5431)

This will add a new Scatter chart type to analytics-chart. 

[Screencast_20260909_152931.webm](https://github.com/user-attachments/assets/e1c00c92-32b4-4d3c-ba3b-dec9a30012f6)

[MA-5431]: https://konghq.atlassian.net/browse/MA-5431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chalks (approved):
We're accepting that some of this will need changed in the future, but getting it to a place where we can start using it is worth getting this merged so we can iterate.

Very nice work getting this out so quick!

PR #3824, opened 2026-09-10
feat(analytics): support grouped multi-metric time series

filipgutica:
# Summary

Targets: [MA-5425](https://konghq.atlassian.net/browse/MA-5425)

Support time-series queries with multiple metrics and a grouping dimension. `AnalyticsChart` accepts a documented `activeMetric` prop, used only for grouped multi-metric time series. The dashboard tile owns the selection control in its header. Switching metrics preserves dimension labels and colors and updates the axis, tooltip, and thresholds without another query.

Filter thresholds separately because the threshold plugin draws all supplied thresholds, even for inactive metrics. This prevents inactive metrics’ threshold lines from remaining visible when switching selections.

Add grouped examples to the static and editable dashboard sandboxes, plus a host-owned control in the analytics-chart sandbox. Extend generated time-series data to support multiple metrics with dimensions.

Verified with 43 focused tests, package typechecks, lint, sandbox builds, and browser checks covering metric switching, query reuse, and narrow tile layouts.

https://github.com/user-attachments/assets/af028b81-ab85-4e29-aa9a-29d7f4458f9d

[MA-5425]: https://konghq.atlassian.net/browse/MA-5425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/analytics/analytics-chart/src/components/AnalyticsChart.vue
```diff
@@ -182,6 +185,47 @@ const exploreData = computed<ExploreResultV4 | undefined>(() => (
...
+  if (!result || !hasGroupedMetrics.value || !selectedMetric.value) {
+    return result
+  }
+
+  return {
+    ...result,
+    meta: { ...result.meta, metric_names: [selectedMetric.value] },
+  }
+})
+
```
filipgutica, line 213: Changing `metric_names` updates the datasets, but the threshold plugin still draws all the thresholds we pass in. This clears the ones for the other metrics so they don’t stick around when you switch.

packages/analytics/analytics-chart/src/utils/i18n.ts
```diff
@@ -0,0 +1,4 @@
+import english from '../locales/en.json'
+
+// Narrow dynamic names to known keys before passing them to the typed translation API.
+export const isChartLabel = (name: string): name is keyof typeof english.chartLabels => Object.hasOwn(english.chartLabels, name)
```
filipgutica, line 4: This was just a better way to check for the existence of a lang key + narrow it to prevent having to `@ts-ignore`

Chalks (approved):

PR #3832, opened 2026-09-11
feat(entities): add tags multi-select filter to Kong Manager entity lists

kingyue737:
## Summary

- Kong Manager's entity list filter (`EntityFilter`) supported single-value `select` fields only; `tags` was shown as a table column on every entity but was never a filter dimension.
- Extends the shared `FilterSchema`/`EntityFilter.vue` with a `multiple` mode rendered as `KMultiselect`; selected values are joined with `,` when applied (Admin API AND semantics) and split back out when restoring from the query string.
- Adds a new `useTagsFilter` composable that fetches the first page of `GET /tags` and surfaces the distinct tag names as select suggestions (`enableItemCreation` lets users type a tag not in that first page, since `/tags` returns paginated tag-to-entity associations rather than a deduplicated tag list).
- Wires a `tags` filter into every Kong Manager entity list still using `EntityFilter`: Routes, Gateway Services, Consumers, Consumer Groups, Plugins, Vaults, Upstreams, SNIs, Keys, Key Sets, Certificates, CA Certificates, Redis Configurations.

## Out of scope

- **Konnect** — gated out everywhere (`config.app === 'kongManager'` only).
- **Plugins' enhanced filter UI** (`PluginFilter.vue` / KFilterGroup, behind `pluginTableImprovements.filtering`) — untouched. That UI already has its own tags filter using `/` (OR) semantics; Plugins' Kong Manager view uses the plain `EntityFilter` path this PR extends, so it gets the new `,` (AND) tags filter without touching the other UI.
- Vaults in AI Gateway mode and Redis Configurations with Konnect-managed Redis enabled don't get the filter, since those already hide the `tags` column.

## Test plan

- [x] `EntityFilter.cy.ts` — new cases for multiselect rendering, comma-joined apply, clear, and restoring selections from an existing query string.
- [x] Existing Cypress component suites for every touched entity list re-run green (Routes, Gateway Services, Consumers, Consumer Groups, Vaults, Upstreams, SNIs, Keys, Key Sets, Certificates, CA Certificates, Redis Configurations, Plugins).
- [x] `vue-tsc` typecheck and `eslint` clean on all touched packages.
- [ ] Manual verification against a live Admin API `/tags` response was not done in this session (no ticket/manual QA pass requested for this draft).

### DIFF
--- a/packages/entities/entities-certificates/src/components/CACertificateList.vue
+++ b/packages/entities/entities-certificates/src/components/CACertificateList.vue
@@ -221,6 +221,7 @@ import {
   useTableState,
   useFetcher,
   useDeleteUrlBuilder,
+  useTagsFilter,
   TableTags,
 } from '@kong-ui-public/entities-shared'
 import type {
@@ -233,9 +234,11 @@ import type {
   BaseTableHeaders,
   EmptyStateOptions,
   ExactMatchFilterConfig,
+  FilterFields,
   FuzzyMatchFilterConfig,
   TableErrorMessage,
 } from '@kong-ui-public/entities-shared'
+import type { SelectItem } from '@kong/kongponents'
 import '@kong-ui-public/entities-shared/dist/style.css'
 
 const emit = defineEmits<{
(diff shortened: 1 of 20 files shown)


kingyue737:
Currently we have no plan to enhance on-prem filtering experience

PR #3833, opened 2026-09-11
feat(core): init package for e-charts [OLLY-1320]

vaibhavrajsingh2001:
# Summary

**Background:**
there is a new requirement in AI observability world ([Jira](https://konghq.atlassian.net/browse/OLLY-1320)), where we need to add some new chart types that aren't supported by our existing analytics chart library (i.e. chart-js)
Hence, we want to explore using the [Apache e-charts](https://echarts.apache.org/en/index.html) package for them (it's free, open-source, has well maintained [vue package](https://vue-echarts.dev/), and covers all our new chart type requirements)

**Why not add it in exisitng analytics-charts package?**
although for now we'll use it inside analytics dashboards, but as soon as next month onwards, we need to use it in other parts of Konnect e.g. AI gateway MFE, AI observability L1 items like conversations, etc. Also, it'll be a very thin wrapper around e-charts package itself, unlike the current analytics-charts components that are heavily custom built around analytics APIs. Moreover, analytics-charts already bundles chart-js so adding another charting package will bump up the bundle size to too big.

Hence, spinning up a separate new core package for this.

Resolves: https://konghq.atlassian.net/browse/OLLY-1320

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -39,6 +39,7 @@
 /packages/core/expressions/ @Kong/team-km @Kong/team-core-ui
 /packages/core/table-data-grid @Kong/team-data @Kong/team-core-ui
 /packages/core/freeform/ @Kong/team-km @Kong/team-core-ui
+/packages/core/e-charts/ @Kong/team-core-ui @Kong/team-data
 
 # Portal packages
 /packages/portal/document-viewer/ @Kong/team-devx @Kong/team-core-ui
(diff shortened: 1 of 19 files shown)


ValeryG (approved):

Justineo:
Hi! I’m the author of Vue ECharts, and @kingyue737 also helps maintain the package. We’d be happy to help if you run into any issues while using it.
By the way, could we change the package name to `echarts`? This naming convention is more consistent with what we typically use in the ECharts community.

PR #3835, opened 2026-09-11
feat(echarts): add support for heatmap [OLLY-1320]

vaibhavrajsingh2001:
# Summary

image

image


packages/core/e-charts/src/components/chart-types/HeatmapChart.vue
```diff
@@ -0,0 +1,106 @@
...
+import { useChartColors } from '../../composables/useChartColors.ts'
+import { deepMerge } from '../../utils/deepMerge.ts'
+import type { EChartsOption } from 'echarts'
+import type { HeatmapChartProps } from '../../types/index.ts'
+
+// The renderer, grid and tooltip are registered by the base `ECharts` component
+use([HeatmapChart, VisualMapComponent])
+
+const {
+  option,
```
afrzhussain: nit: should this `options` ?

vaibhavrajsingh2001: That's what the underlying prop is called in the library, hence singular instead of plural

https://github.com/ecomfe/vue-echarts/blob/621d1cdd0f7cd7d7b9d61b53ae24693154258935/src/ECharts.ts#L63

packages/core/e-charts/src/utils/deepMerge.ts
```diff
@@ -0,0 +1,34 @@
...
+const isPlainObject = (value: unknown): value is Mergeable => (
+  typeof value === 'object' && value !== null && !Array.isArray(value)
+)
+
+/**
+ * Deep-merge chart option objects. Nested plain objects are merged
+ * recursively; any other value type (including arrays) in `overrides`
+ * replaces the value in `base`. `null`/`undefined` overrides are ignored.
+ */
+export const deepMerge = <T extends Mergeable>(base: T, ...overrides: Array<Mergeable | null | undefined>): T => {
```
afrzhussain: question: is this not available anywhere else in our codebase ?

afrzhussain (approved):
LGTM

packages/core/e-charts/src/utils/deepMerge.ts
vaibhavrajsingh2001: in konnect-ui-apps we sometimes use the `deepmerge` package, but this usecase is pretty simple hence didn't use a package

vaibhavrajsingh2001: comparison:

| | Our `deepMerge` | `@fastify/deepmerge` (v3.2.1) | `lodash-es` `mergeWith` |
| --- | --- | --- | --- |
| Arrays (`series`, `inRange.color`, `dataZoom`) | Replaced | Concatenated by default; replacing needs `mergeArray` | Merged by index by default; replacing needs a customizer |
| `undefined` overrides | Ignored | Overwrite by default; ignoring needs `onlyDefinedProperties: true` | Ignored when the target has a value |
| Several overrides at once | Yes | Only with `all: true` | Yes |
| Class instances (e.g. ECharts gradients) | Spread as plain objects | Not supported by default; needs `cloneProtoObject` | Merged as objects |
| `__proto__` / prototype keys | Not guarded (small fix to add) | Guarded | Guarded |
| Mutates inputs | No | No | Yes, the first argument (pass `{}` first) |
| Config to match our behavior | None | 4 options | 1 customizer |
| New dependency for consumers | None | Yes | Yes (already used by other packages in this repo) |
| Size | ~34 lines, tested | ~21 kB unpacked | Shared lodash modules |

PR #3846, opened 2026-09-16
chore(deps): bump swrv to ^1.3.0, kongponents to 9.64.18, adopt swrv/testing

merrywhether:
Raises the `swrv` floor to `^1.3.0`, bumps `@kong/kongponents` to `9.64.18`, and wires swrv's per-mount cache isolation so component tests stop sharing caches.

kongponents 9.64.18 externalizes its runtime dependencies and declares `swrv ^1.3.0`, so it now resolves the same swrv instance these packages do instead of carrying an inlined copy. That is what makes isolation reachable at all — a vendored copy can never share a cache with the app's.

Isolation is wired per runner, because one mechanism does not cover both:

- **Vitest** — `swrv/testing` in the `setupFiles` of the three packages that actually use swrv, rather than repo-wide. `mergeConfig` concatenates, so the shared `vitest.setup.ts` entry survives.
- **Cypress** — `swrv/testing` cannot work there. Cypress bundles its own `@vue/test-utils`, so the config that module mutates is not the one its `mount()` reads. The mount command calls `provideSwrvCache(app)` instead, which Cypress does merge.

Also drops a `swrv/dist/types` deep import in `useRequest.ts`; those types are root exports as of swrv 1.3.0.

[TEST_COVERAGE_EXEMPT_REASON] The only prod code change is moving the location of type imports which is verified by type-checking.


package.json
```diff
@@ -60,6 +60,7 @@
     "stylelint-config-recommended-scss": "^17.0.1",
     "stylelint-config-recommended-vue": "^2.0.0",
     "stylelint-order": "^8.1.1",
+    "swrv": "^1.3.0",
```
merrywhether, line 63: Necessary for Cypress config, can be dropped when Cypress is removed

adamdehaven (approved):

Chalks (approved):

PR #3855, opened 2026-09-21
fix(analytics): fall back when the query bridge has no feature-flag evaluator

merrywhether:
# Summary

`useEvaluateFeatureFlag`'s guard only checked that the injected query bridge was truthy, not that it exposed `evaluateFeatureFlagFn`, so it can then throw a `TypeError` when an undefined callback is invoked, inside a `computed` during render. This went live in #3849, which added `useEvaluateFeatureFlag()` calls to `useScatterDatasets`, `useExploreResultToDatasets`, and `useExploreResultToTimeDatasets`.

`ScatterChartRenderer.cy.ts` is one of the specs that [provides a partial bridge](https://github.com/Kong/public-ui-components/blob/main/packages/analytics/dashboard-renderer/src/components/ScatterChartRenderer.cy.ts#L73) and thus triggers this.

Also renames `useEvauluateFeatureFlag.ts` to `useEvaluateFeatureFlag.ts` (fixing the typo) in both the analytics-chart and dashboard-renderer packages while I'm touching the file.

## Verification

- Reproduced the hang locally: `pnpm exec cypress run --component -b chrome --spec './src/components/ScatterChartRenderer.cy.ts'` hangs indefinitely every time. After fix, test passes every time.
- Added unit tests for `useEvaluateFeatureFlag` in both packages: no bridge, bridge with the function, and bridge without the function (latter of which replicates the current situation)

## Notes

- Another option is just removing the cast from the test config and forcing the full shape to be expressed, but that doesn't fix the underlying runtime issue so I went with the more robust change. Happy to swap if that is prefered.
- I'm looking into getting #3248 merged to prevent issues like this from merging in the first place


Chalks (approved):
thank you for fixing the spelling on that composable. :D

